PR #3, opened 2019-04-02
Add a `goldenTestCBOR` function.

dnadales:
This function is basically a re-implementation of 'goldenTestBi', but it can be
used with values of types that don't have a `Bi` instance.

I didn't dare to implement `goldenTestBi` in terms of `goldenTestCBOR`, since the encoding and decoding functions seem to differ slightly.

Part of https://github.com/input-output-hk/cardano-ledger/pull/394

### DIFF
--- a/binary/test/Test/Cardano/Binary/Helpers/GoldenRoundTrip.hs
+++ b/binary/test/Test/Cardano/Binary/Helpers/GoldenRoundTrip.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE RankNTypes          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
 
 -- | Golden and round-trip testing of 'Bi' instances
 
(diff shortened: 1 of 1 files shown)


binary/test/Test/Cardano/Binary/Helpers/GoldenRoundTrip.hs
```diff
@@ -65,6 +76,13 @@ failHexDumpDiff x y = case hexDumpDiff x y of
...
 
+-- | Check that the 'encode' and 'decode' function of the 'Bi' instances work as
+-- expected w.r.t. the give reference data, this is, given a value @x::a@, and
+-- a file path @fp@:
+--
+-- - The encoded data should coincide with the contents of the @fp@.
+-- - Decoding @fp@ should give as a result @x@
+--
 goldenTestBi :: (Bi a, Eq a, Show a, HasCallStack) => a -> FilePath -> Property
 goldenTestBi x path = withFrozenCallStack $ do
```
ruhatch: I think we should define this one in terms of the other one!

dnadales: Done!

PR #4, opened 2019-04-03
[cardano-ledger #145] Split Bi class into ToCBOR and FromCBOR

ruhatch:


binary/src/Cardano/Binary.hs
```diff
@@ -0,0 +1,12 @@
+module Cardano.Binary
```
dnadales: I'm just curious, what was the rationale for using re-exports?

binary/src/Cardano/Binary/EvilInstances.hs
```diff
@@ -1,25 +0,0 @@
-{-# LANGUAGE FlexibleInstances    #-}
```
dnadales: Did we really have a module with this name? :scream_cat:

binary/test/Test/Cardano/Binary/Bi.hs
```diff
@@ -16,69 +16,66 @@ import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Range as Range
```
dnadales: Is `Test.Cardano.Binary.Bi` still a good name for this module? Shouldn't be renamed to something like `Test.Cardano.Binary.RoundTrip`?

PR #12, opened 2019-05-07
Expose `reAnnotate`, which will recover an annotation via reserialisa…

nc6:
…tion


stack.yaml
```diff
@@ -1,15 +1,15 @@
...
-    commit: 71c0b1340e0b47d6f463c005406468b49d603dc6
+    commit: d8091767addc1a51e11cb2f04392e4eb40e39b17
     subdirs:
       - .
       - test
 
-nix:
-  shell-file: ./scripts/nix/stack-shell.nix
+# nix:
+#   shell-file: ./scripts/nix/stack-shell.nix
```
ruhatch: Is this meant to be commented?

PR #17, opened 2019-05-22
Add support for buildkite and coveralls

### LINKED ISSUE #1: Set up basic buildkite CI
Copy the set up from `cardano-chain`

DeepakKapiswe:
Closes #1


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,34 @@
...
+      # so copy the source into /build/cardano-base
+      - "rm -rf /build/cardano-base"
+      - "cp -R . /build/cardano-base"
+      - "cd /build/cardano-base"
+      - "nix-build scripts/buildkite -o stack-rebuild"
+      - "./stack-rebuild"
+    agents:
+      system: x86_64-linux
+
+  # - label: 'brittany'
```
ruhatch: You can remove this commented section

nix/default.nix
```diff
@@ -0,0 +1,45 @@
...
+      # cross compilation plumbing to make Template Haskell
+      # work when cross compiling.  For now we need to
+      # list the packages that require template haskell
+      # explicity here.
+      iohk-module
+      {
+        # katip has an version bound of Win32 < 2.6; this however
+        # implies that it's incompatible with ghc-8.6 (on windows).
+        # Let's force it to accept out packageset.
+        packages.katip.doExactConfig = true;
```
ruhatch: I don't think `katip` is an actual dependency here, so you can remove this line and comment above

scripts/buildkite/cache-s3.nix
```diff
@@ -0,0 +1,25 @@
+{ stdenv, fetchurl, zlib, gmp }:
```
ruhatch: I don't think we need this file anymore, so you can delete it

Issue #24, opened 2019-06-20
The module Cardano.Binary.Limit is unused


erikd:
This module was carried over from the old code base, but is not used anywhere in the code base.

Lets kill it.

PR #25, opened 2019-06-20
remove unused module Limit.hs

DeepakKapiswe:
addresses #24


ruhatch:
I think this is used in the `cardano-crypto-wrapper` package...could you confirm that please

DeepakKapiswe:
yes, Limit is used at several places in cardano-ledger.

DeepakKapiswe:
Cardano-ledger uses Limit Type at several places so we can't remove it, although it is not used in Cardano-binary.

erikd:
> yes, Limit is used at several places in cardano-ledger.

Its seems the only place `Limits` or any of the associated stuff is used in in the `Cardano.Crypto.Limits` modules and that in turn is only used in tests. Nowhere else in this new code base is `Cardano.Binary.Limits` or `Cardano.Crypto.Limits` used.

ruhatch:
Aha, so we should move it to `Test.Cardano.Binary.Limits`?

DeepakKapiswe:
Limit type is used only at `Test.Cardano.Crypto.Limits` so I think we can drop the Limit.hs module here.

binary/test/Test/Cardano/Binary/Helpers.hs
```diff
@@ -179,72 +168,6 @@ extensionProperty = forAll @a (arbitrary :: Gen a) $ \input ->
     (encoded :: a) = unsafeDeserialize (serialize u)   -- Step 3
   in encoded === input
 
---------------------------------------------------------------------------------
--- Message length
---------------------------------------------------------------------------------
-
-msgLenLimitedCheck :: ToCBOR a => Limit a -> a -> Property
```
ruhatch: Can you check whether this stuff has been translated into `hedgehog` tests in `cardano-crypto-tests`? We should have these tests, even if they're currently unused.

DeepakKapiswe: https://github.com/input-output-hk/cardano-ledger/blob/e2d4663f2251ef76bae94e48fef910b8f5f867bd/crypto/test/Test/Cardano/Crypto/Limits.hs#L59

PR #27, opened 2019-06-27
Add cardano-crypto-class package with crypto type classes

ruhatch:
This is all copied from `ouroboros-consensus` and then adapted to use `cardano-binary` instead of `serialise`.


binary/src/Cardano/Binary/Serialize.hs
```diff
@@ -70,6 +71,10 @@ serializeEncoding =
     -- sacred.
         strategy = Builder.safeStrategy 1024 4096
 
+serializeEncoding' :: Encoding -> ByteString
```
mgudemann: maybe add a comment / doc

binary/src/Cardano/Binary/ToCBOR.hs
```diff
@@ -500,6 +504,50 @@ instance (ToCBOR a, ToCBOR b, ToCBOR c, ToCBOR d) => ToCBOR (a,b,c,d) where
...
+  toCBOR (a, b, c, d, e) =
+    E.encodeListLen 5
+      <> toCBOR a
+      <> toCBOR b
+      <> toCBOR c
+      <> toCBOR d
+      <> toCBOR e
+
+  encodedSizeExpr size _ =
+    1
```
mgudemann: is this for the `Word8` tag?

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -0,0 +1,105 @@
+name:                cardano-crypto-class
+version:             2.0.0
+synopsis:            Type classes abstractign over cryptography primitives for Cardano
```
mgudemann:
```suggestion
synopsis:            Type classes abstracting over cryptography primitives for Cardano
```

cardano-crypto-class/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2016 IOHK
```
mgudemann: different date from entry in `.cabal` file below

scripts/buildkite/rebuild.hs
```diff
@@ -13,7 +13,7 @@ import Turtle
 -- | Run build and upload coverage information when successful
 main :: IO ()
 main = do
-  buildResult <- buildStep (Just ["--scenario=ContinuousIntegration"])
+  buildResult <- buildStep Nothing
```
mgudemann: what is the result of this change ?

ruhatch: Just removes a test flag that's not used in this package

binary/src/Cardano/Binary/ToCBOR.hs
ruhatch: For the list len I believe

binary/src/Cardano/Binary/Serialize.hs
ruhatch: Sure thing

PR #28, opened 2019-07-02
Bump versions of dependencies

erikd:


erikd:
`snapshot.yaml` only exists in `cardano-prelude` and obviously that file does not contain a hash for `cardano-prelude`.

rvl:
I mean the revision on line 1 of `stack.yaml`.

erikd:
Congratulations @rvl, you just found a bug in my `cardano-repo-tool` :).

Will fix that and then update this PR.

PR #41, opened 2019-09-05
Annotated decoding changes

redxaxder:


binary/src/Cardano/Binary/Annotated.hs
```diff
@@ -91,3 +103,54 @@ class Decoded t where
...
+  pure $ x $ sliceOffsets start end bytes
+  where
+  sliceOffsets :: ByteOffset -> ByteOffset -> LByteString -> LByteString
+  sliceOffsets start end = BSL.take (end - start) . BSL.drop start
+
+-- | Equivalent to withSlice for strict ByteStrings
+withSlice' :: AnnotatedDecoder s (ByteString -> a) -> AnnotatedDecoder s a
+withSlice' = withSlice . fmap (. BSL.toStrict)
+
+{-# WARNING fromCBORPlaceholder "placeholder for empty deserialization to that things compile for now" #-}
```
dnadales:
```suggestion
{-# WARNING fromCBORPlaceholder "placeholder for empty deserialization so that things compile for now" #-}
```

binary/src/Cardano/Binary/Annotated.hs
```diff
@@ -91,3 +102,53 @@ class Decoded t where
 instance Decoded (Annotated b ByteString) where
   type BaseType (Annotated b ByteString) = b
   recoverBytes = annotation
+
+-------------------------------------------------------------------------
+-- Annotated Decoder
+-------------------------------------------------------------------------
+
+
+-- | An AnnotatedDecoder produces a value which needs a reference to the original ByteString to be
```
nc6: I'm probably slow, but I still struggle to understand exactly what's meant here. Maybe it's possible to include a small example?

PR #54, opened 2019-10-30
Add optional context to DSIGN and KES

edsko:


edsko:
Used this in consensus to great effect.

PR #59, opened 2019-11-22
Move Slot related primitives into cardano-base

nc6:
There are a few primitive types that pertain to the full gamut of cardano packages - mostly concerning slots and epochs. This PR pulls these into a new base package, which can then be used over the place as a minimal dependency.

The STM-backed implementation of `EpochInfo` is not moved over, because it depends on more things from `ouroboros-consensus`. It may ultimately make sense to move those also.


edsko:
Moving some of these shared types to a base library seems to make sense to me. Just not entirely sure that `EpochInfo` should live there also. That doens't feel quite so "basic". It is moreover stateful and instances of it should probably be provided by the node rather than conjured up out of nothing. I'm nto sure that it should live in this new `slotting` library.

mrBliss:
Related https://github.com/input-output-hk/ouroboros-network/issues/936

PR #62, opened 2019-12-12
Add ToCBOR for BlockNo

nc6:


nc6:
Mostly to ensure (and make clear) that the `ToCBOR` instance does the same as the `Serialise` instance.

PR #63, opened 2019-12-16
Add `Serialise` instance for `WithOrigin`.

nc6:


slotting/src/Cardano/Slotting/Slot.hs
```diff
@@ -46,7 +46,17 @@ genesisSlotNo = SlotNo 0
...
+    ( Eq,
+      Ord,
+      Show,
+      Generic,
+      Functor,
+      Foldable,
+      Traversable,
+      Serialise,
+      NoUnexpectedThunks
+    )
```
mrBliss: I'd go for "comma first"

PR #74, opened 2020-02-04
Use quiet-0.2 for some data types

nfrisby:
E.G. `show (SlotNo 3) == "SlotNo 3"` now instead of `"SlotNo {unSlotNo = 3}"`.

### DIFF
--- a/slotting/src/Cardano/Slotting/Block.hs
+++ b/slotting/src/Cardano/Slotting/Block.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module Cardano.Slotting.Block
@@ -12,13 +13,15 @@ import Cardano.Prelude (NoUnexpectedThunks)
 import Codec.Serialise (Serialise (..))
 import Data.Word (Word64)
 import GHC.Generics (Generic)
+import Quiet (Quiet (..))
 
 -- | The 0-based index of the block in the blockchain.
 -- BlockNo is <= SlotNo and is only equal at slot N if there is a block
 -- for every slot where N <= SlotNo.
 newtype BlockNo = BlockNo {unBlockNo :: Word64}
-  deriving stock (Show, Eq, Ord, Generic)
+  deriving stock (Eq, Ord, Generic)
   deriving newtype (Enum, Bounded, Num, Serialise, NoUnexpectedThunks)
+  deriving (Show) via (Quiet BlockNo)
 
 instance ToCBOR BlockNo where
   toCBOR = encode
--- a/slotting/cardano-slotting.cabal
+++ b/slotting/cardano-slotting.cabal
@@ -45,5 +45,6 @@ library
                       , containers
                       , mmorph
                       , mtl
+                      , quiet
                       , serialise
                       , transformers
--- a/slotting/src/Cardano/Slotting/Slot.hs
+++ b/slotting/src/Cardano/Slotting/Slot.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module Cardano.Slotting.Slot
@@ -26,11 +27,13 @@ import Codec.Serialise (Serialise (..))
 import Data.Typeable (Typeable)
 import Data.Word (Word64)
 import GHC.Generics (Generic)
+import Quiet (Quiet (..))
 
 -- | The 0-based index for the Ourboros time slot.
 newtype SlotNo = SlotNo {unSlotNo :: Word64}
-  deriving stock (Show, Eq, Ord, Generic)
+  deriving stock ( Eq, Ord, Generic)
   deriving newtype (Enum, Bounded, Num, NFData, Serialise, NoUnexpectedThunks)
+  deriving (Show) via (Quiet SlotNo)
 
 instance ToCBOR SlotNo where
   toCBOR = encode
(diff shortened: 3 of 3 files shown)


erikd:
Does the `read . show == id` identity still hold?

slotting/src/Cardano/Slotting/Block.hs
```diff
@@ -9,7 +9,8 @@ module Cardano.Slotting.Block
 where
 
 import Cardano.Binary (FromCBOR (..), ToCBOR (..))
-import Cardano.Prelude (NoUnexpectedThunks)
+import Cardano.Prelude (NoUnexpectedThunks, GShowK1 (..),
```
mrBliss: Why is it necessary to import `GShowK1 (..)`? It would be nice if we could avoid that.

nfrisby: Yes, it's an implementation detail of the `GHC.Generics` function, which relies on `coerce`, which needs the constructor "in scope". :(

I _think_ importing the module qualified would satisfy that restriction without having to list `GShowK1` explicitly. Would this be acceptable?

```
import Cardano.Prelude (NoUnexpectedThunks)
import qualified Cardano.Prelude as CPrelude
```

nfrisby:
> Does the `read . show == id` identity still hold?

It does not! I am completely flabbergasted, blindsided. I did not expect this at all. Thank you for explicitly asking the obvious question.

```
Prelude> newtype SlotNo = SlotNo {unSlotNo :: Int} deriving (Read)
Prelude> read "SlotNo 3" :: SlotNo
SlotNo {unSlotNo = *** Exception: Prelude.read: no parse
```

I has assumed `Read` for a record type could also accept the plain arguments. (I.E. That `read . show = id` but also that `read . otherShow` might also be `id`.)

@erikd @mrBliss Should I abandon this path, or should I double-down and also provide a suitable `generic-deriving` `Read` implementation? My hunch is _abandon_: I probably wouldn't have introduced any of this if I had realized `Read` _required_ record syntax.

erikd:
The main reason I use `Read` (which I use less here at IOHK than I have in previous jobs or in my own code) is so that when I get a failure in a property test I can get the counterexample ans paste it into a test file or GHCi to begin debugging. Especially for larger data structures having to edit what I paste to get it accepted is a real pain in the neck.

In order to get around this I take one of two paths:
* Hand implement `Read` and `Show` and use a property test to validate them
* Derive `Read` and `Show` and have a separate `renderX` function that does pretty printing.

Like you said, here at IOHK we tend not to use `Read` much, so this may not be an issue.

nfrisby:
> The main reason I use `Read` (which I use less here at IOHK than I have in previous jobs or in my own code) is so that when I get a failure in a property test I can get the counterexample ans paste it into a test file or GHCi to begin debugging. Especially for larger data structures having to edit what I paste to get it accepted is a real pain in the neck.

I absolutely feel that same pain! I have been copying-and-pasting data structures from QuickCheck output that contain multiple `Map`/`Set`s multiples times per day, and qualifying each `fromList` is obnoxious. I didn't think to use `read` for that (it's so far off my radar); it's a great use in the short-term.

In fact, the main motivation for this whole `gshowsPrecWithoutRecordSyntax` is for the QuickCheck output etc to be more legible in the first place :)

erikd:
> n fact, the main motivation for this whole gshowsPrecWithoutRecordSyntax is for the QuickCheck output etc to be more legible in the first place :)

But standard derived `Read`/`Show` instances may be the easiest way to get that result.

erikd:
@nfrisby https://hackage.haskell.org/package/quiet-0.2/docs/Quiet.html

Is that of any use?

nfrisby:
> But standard derived Read/Show instances may be the easiest way to get that result.

Unless you mean something else by "standard", I find the record syntax on these simplistic constructors makes it quite hard to read the expression.

> @nfrisby https://hackage.haskell.org/package/quiet-0.2/docs/Quiet.html
> 
> Is that of any use?

Yes, that looks like the exact sort of thing I wanted. I'm weary that it doesn't rely on `generic-deriving`'s definitions of `read/show`, which I consider to be the most mature behind GHC's, but getting `Read/Show` right isn't Herculean after all and `Quiet`'s test suite looks non-trivial. Thanks for the pointer.

erikd:
@nfrisby How would `generic-sop` improve `quiet` ?

jacobstanley:
> I'm weary that it doesn't rely on `generic-deriving`'s definitions of `read/show`

@nfrisby sorry to barge in, I was talking with @erikd offline

I have been carrying around the `quiet` code for a number of years, only put it on Hackage this week, but if I remember right the `Show` code is a [derivative](https://github.com/jacobstanley/quiet/blob/master/LICENSE#L2) of `generic-deriving`. The `Read` code I wrote myself, but as you note is fairly comprehensively tested.

The code is simpler than `generic-deriving` as it defers to `Show` for everything but the generic structure (rather that implementing a new `GShow` typeclass for all types)

I use it for the same pains discussed here, QuickCheck/Hedgehog counterexamples!

edsko:
I think copy-and-paste-ability into ghci is indeed very useful. Just want to point out that the derived `Show` without the record fields _still_ satisfies that requirement.

erikd:
@edsko I do not think that is true when the constructor is not exported from the module it is defined in (eg for a smart constructor).

edsko:
> @edsko I do not think that is true when the constructor is not exported from the module it is defined in (eg in order to force use of an alternative smart constructor).

I didn't articulate myself very well. You are right of course, a derived `Show` instance versus a custom one might not work. I was commenting merely on @nfrisby 's changes here: _if_ the standard derived `Show` instance works, then _his_ derived `Show` instance (which omits the record names) would _also_ still work.

nfrisby:
> @nfrisby sorry to barge in, I was talking with @erikd offline

Not at all -- thanks for commenting! And thank you for packaging up and sharing your work on Hackage.

nfrisby:
OK. I've updated this to use `quiet`, but it will need input-output-hk/cardano-prelude#111, which adds `quiet` to the wrapped LTS snapshot. Once that merges, I can update the build config files here accordingly.

slotting/src/Cardano/Slotting/Block.hs
```diff
@@ -12,14 +12,18 @@ import Cardano.Prelude (NoUnexpectedThunks)
...
 -- | The 0-based index of the block in the blockchain.
 -- BlockNo is <= SlotNo and is only equal at slot N if there is a block
 -- for every slot where N <= SlotNo.
 newtype BlockNo = BlockNo {unBlockNo :: Word64}
-  deriving stock (Show, Eq, Ord, Generic)
+  deriving stock (Eq, Ord, Generic)
   deriving newtype (Enum, Bounded, Num, Serialise, NoUnexpectedThunks)
 
+instance Show BlockNo where
+  showsPrec = Quiet.qshowsPrec
```
mrBliss, line 25: I would have used
```haskell
import Quiet (Quiet (..))

  ..
  deriving (Show) via (Quiet BlockNo)
```
Not a fan?

nfrisby: Same response as https://github.com/input-output-hk/ouroboros-network/pull/1019#discussion_r427992547

newhoggy:
Looks like this is dealt with by the related PR https://github.com/input-output-hk/cardano-base/pull/151

Closing.

PR #76, opened 2020-02-05
SRE-33 haskell overhaul

craigem:
This is the overhaul of Haskell for `cardano-base` along the lines of `cardano-node` [PR #507](https://github.com/input-output-hk/cardano-node/pull/507/).


lib.nix
```diff
@@ -1,18 +1,42 @@
...
+  haskellPackages = import ./nix/pkgs.nix {
+    inherit pkgs;
+    src = ./.;
+  };
   lib = pkgs.lib;
-in lib // { inherit iohkNix pkgs; inherit (iohkNix) nix-tools; }
+  niv = (import sources.niv {}).niv;
+  isCardanoBase = with lib; package:
+    (package.isHaskell or false) &&
+      (hasPrefix "cardano-base" package.identifier.name);
```
disassembler: need to fix the prefixes here. none of these packages start with cardano-base

scripts/buildkite/cache-s3.nix
```diff
@@ -0,0 +1,25 @@
+{ stdenv, fetchurl, zlib, gmp }:
```
disassembler: why are you adding this file?

stack.yaml
```diff
@@ -1,17 +1,117 @@
-resolver: https://raw.githubusercontent.com/input-output-hk/cardano-prelude/cb54482c7e3d45f6c1aac490dc937ed835de0333/snapshot.yaml
+resolver: https://raw.githubusercontent.com/input-output-hk/cardano-prelude/00487726c4bc21b4744e59d913334ebfeac7d68e/snapshot.yaml
```
disassembler: this absolutely looks wrong

scripts/buildkite/rebuild.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE LambdaCase          #-}
```
disassembler: is this file being changed intentionally?

craigem: reverted this back to the version from master with #9025568

scripts/buildkite/cache-s3.nix
craigem: It's called from default.nix.

Removed the file and it's reference form `default.nix` in dba088f

stack.yaml
craigem: Reverted to the resolver from master with commit 860e263

lib.nix
craigem: Prefix removed in commit 729da10

disassembler:
bors ping

PR #84, opened 2020-03-04
Add comment for `iterationCountKES` function

mgudemann:
`iterationCountKES` should return the number of times a KES signature key has already been updated

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/KES/Class.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/KES/Class.hs
@@ -70,10 +70,15 @@ class ( Typeable v
 
   deriveVerKeyKES :: SignKeyKES v -> VerKeyKES v
 
+  -- | Update the KES signature key to the specified period. The intended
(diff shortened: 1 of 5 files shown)


mgudemann:
I'd be happy to rename this. Would `updateCountKES` be ok?

edsko:
> I'd be happy to rename this. Would updateCountKES be ok?

I'd prefer that, but I think better still would be for `update` to take an argument (which period to move forward to); if that change is made, it's less that we are _counting how often_ we updated but rather "what is the current period" I guess..?

mgudemann:
ok, fine with me. Then both passing a past period or a future period with too many evolutions would result in `Nothing`, right?

edsko:
Just thinking out loud: perhaps a past period should result in an exception instead? In other words, `update` would take a precondition in terms of (the renamed vesion of) `udpateCount`? After all, doing so is a bug, whereas exceeding it's period is just something that can happen  and should be taken into account?

PR #87, opened 2020-03-26
Add bangs to arguments of SigMockDSIGN

mrBliss:


mrBliss:
I ran into this thunk while working on https://github.com/input-output-hk/ouroboros-network/issues/1558

PR #103, opened 2020-04-17
Fail when decoding a negative Natural instead of throwing

mrBliss:
Decoding a negative `Natural` should not result in `Exception: arithmetic
underflow` but in a `cborError`.


cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -60,7 +60,6 @@ library
                      , cryptonite
                      , deepseq
                      , memory
-                     , reflection
```
edsko, line 63: :-o

PR #105, opened 2020-04-23
Add Eq instance for SignKeyKES

Jimbo4350:
- Needed for round trip testing in cardano-node
- Also added `Eq` constraints so in future the relevant `Eq` instances will need to be defined.


nc6:
Pretty sure these were deliberately _removed_. What is the rationale behind adding them?

PR #109, opened 2020-05-04
slotting: Add Read instance for SlotNo

erikd:
* Read instance is needed for the CLI.
* Use the Quiet versions of Read/Show.


erikd:
Seems like I am losing my mind. Please disregard.

PR #117, opened 2020-05-22
HasCallStack in EpochInfo

edsko:
This is important, because the EpochInfo required by the Shelley ledger is
`EpochInfo Identity`, but that is really not possible, the `EpochInfo` we will
feed it has a limited range (and must do). We should fix Shelley, but until
such time, having the callstack available makes debugging a lot easier.


edsko:
Fixed.

PR #121, opened 2020-06-04
Update index-state and deps

erikd:


erikd:
@dcoutts ok-ed this in Slack.

PR #125, opened 2020-06-09
Add FiniteBytes

phadej:
This can be used as a value secure memory foreign pointer points
to. Later, we might want to revisit this
as all sizes are static (hash digest, key length, ...)
and not use Storable at all. Additional benefit
would be from not having `peek` and `poke`.

But currently, being able to peek and show maybe useful.


phadej:
@dcoutts could this be merged (given CI is happy). I added few more tools.

Here `Show` instance is legit, as `FiniteBytes` may or may not be secret bytes. I'll add a `newtype` over `ForeignPtr` to differentiate pointers secret data from everything else (will do as part of #129 )

phadej:
part of #138

PR #129, opened 2020-06-22
Sodium hash

phadej:
Implementing `Hash` module using `libsodium`.

### DIFF
--- a/cardano-crypto-class/memory-example/Main.hs
+++ b/cardano-crypto-class/memory-example/Main.hs
@@ -17,7 +17,7 @@ import           System.Environment (getArgs)
 import           System.Posix.Process (getProcessID)
 #endif
 
-import qualified Data.ByteString as SB
+import qualified Data.ByteString as BS
 
 import           Cardano.Crypto.Libsodium
 import           Cardano.Crypto.Libsodium.MLockedBytes.Internal (MLockedFiniteBytes (..))
@@ -42,14 +42,14 @@ main = do
 
     -- example SHA256 hash
     do
-      let input = SB.pack [0..255]
+      let input = BS.pack [0..255]
       MLFB hash <- digestMLockedBS (Proxy @SHA256) input
       traceMLockedForeignPtr hash
       print (digest (Proxy @SHA256) input)
 
     -- example Blake2b_256 hash
     do
-      let input = SB.pack [0..255]
+      let input = BS.pack [0..255]
       MLFB hash <- digestMLockedBS (Proxy @Blake2b_256) input
       traceMLockedForeignPtr hash
       print (digest (Proxy @Blake2b_256) input)
--- a/cardano-crypto-class/memory-example/Main.hs
+++ b/cardano-crypto-class/memory-example/Main.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeApplications #-}
 
+-- deprecations disabled to allow using traceMLockedForeignPtr
 {-# OPTIONS_GHC -Wno-deprecations #-}
 module Main (main) where
 
--- a/cardano-crypto-class/memory-example/Main.hs
+++ b/cardano-crypto-class/memory-example/Main.hs
@@ -19,7 +19,7 @@ import           System.Posix.Process (getProcessID)
 import qualified Data.ByteString as SB
 
 import           Cardano.Crypto.Libsodium
-import           Cardano.Crypto.Libsodium.SecureBytes.Internal (SecureFiniteBytes (..))
+import           Cardano.Crypto.Libsodium.MLockedBytes.Internal (MLockedFiniteBytes (..))
 import           Cardano.Crypto.Hash (SHA256, Blake2b_256, digest)
 
 main :: IO ()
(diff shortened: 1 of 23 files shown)


phadej:
@dcoutts this is ready for review

cardano-crypto-class/src/Cardano/Crypto/Libsodium/Constants.hsc
```diff
@@ -0,0 +1,24 @@
...
+    CRYPTO_SHA512_STATE_SIZE,
+    CRYPTO_BLAKE2B_256_STATE_SIZE,
+    )  where
+
+#include <sodium.h>
+
+-- From https://libsodium.gitbook.io/doc/advanced/sha-2_hash_function
+-- and https://libsodium.gitbook.io/doc/hashing/generic_hashing
+
+type CRYPTO_SHA256_BYTES = #{const crypto_hash_sha256_BYTES}
```
nc6, line 16: Could use pattern synonyms for these, and then we can match on them

cardano-crypto-class/src/Cardano/Crypto/Libsodium/Hash.hs
```diff
@@ -0,0 +1,134 @@
...
+digestMLockedFB = digestMLockedStorable
+
+digestMLockedBS
+    :: forall h proxy. (SodiumHashAlgorithm h)
+    => proxy h -> SB.ByteString -> IO (MLockedFiniteBytes (SizeHash h))
+digestMLockedBS p bs = SB.useAsCStringLen bs $ \(ptr, len) -> do
+    digestMLocked p (castPtr ptr) len
+
+-------------------------------------------------------------------------------
+-- Hash expansion
```
nc6, line 73: Can you provide some more detail here? What is hash expansion, and why would I want to do it?

nc6: Oh, or this is specific to the SUM KES construction?

cardano-crypto-class/src/Cardano/Crypto/Libsodium/Memory/Internal.hs
```diff
@@ -0,0 +1,83 @@
...
+
+finalizeMLockedForeignPtr :: forall a. MLockedForeignPtr a -> IO ()
+finalizeMLockedForeignPtr = coerce (finalizeForeignPtr @a)
+
+traceMLockedForeignPtr :: (Storable a, Show a) => MLockedForeignPtr a -> IO ()
+traceMLockedForeignPtr fptr = withMLockedForeignPtr fptr $ \ptr -> do
+    a <- peek ptr
+    print a
+
+{-# DEPRECATED traceMLockedForeignPtr "Don't leave traceMLockedForeignPtr in production" #-}
```
nc6, line 44: Ah, deprecation is for this? Makes sense..

cardano-crypto-class/memory-example/Main.hs
```diff
@@ -12,7 +16,11 @@ import           System.Environment (getArgs)
 import           System.Posix.Process (getProcessID)
 #endif
 
-import           Cardano.Crypto.Libsodium (allocSecureForeignPtr, sodiumInit)
+import qualified Data.ByteString as SB
```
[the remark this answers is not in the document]
phadej: This was copied from `Cardano.Crypto.Hash.Class`

```
cardano-crypto-class/memory-example/Main.hs:import qualified Data.ByteString as SB
cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs:import qualified Data.ByteString as SB
cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs:import qualified Data.ByteString.Char8 as SB8
cardano-crypto-class/src/Cardano/Crypto/Libsodium/Hash.hs:import qualified Data.ByteString as SB
cardano-crypto-tests/src/Test/Crypto/Hash.hs:import qualified Data.ByteString as SB
```

should I change the rest too?

---

FWIW: HLint has an option to say how qualified modules should be named. E.g. https://github.com/phadej/phadej-universe/blob/2f1cb7169b89802fd9180e09d2f4397a06917d5b/z-scripts/.hlint.yaml#L69-L90

cardano-crypto-class/src/Cardano/Crypto/FiniteBytes.hs
```diff
@@ -0,0 +1,144 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE MagicHash           #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UnboxedTuples       #-}
+module Cardano.Crypto.FiniteBytes
```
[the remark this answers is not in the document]
phadej: I figured out that I should use `PinnedFiniteBytes`, and maybe rename it to `PinnedSizedBytes` which would be a bit more obvious name, but will add the documentation there.

cardano-crypto-class/src/Cardano/Crypto/Libsodium/Constants.hsc
phadej: Note: `type`, not value.

cardano-crypto-class/src/Cardano/Crypto/Libsodium/Hash.hs
phadej: Yes this is specific to  SUM KES. The current name for this is `splitSeed`, but that one isn't good either as it uses *some* hashing (and produces seed of size specific to the hash function).

Better name would help.

FWIW, sodium's blake2b implementation allows to pass *key* to the hash function, thus AFAICS salting the result, which we do here manually. See https://libsodium.gitbook.io/doc/hashing/generic_hashing#usage

cardano-crypto-class/src/Cardano/Crypto/Libsodium/Memory/Internal.hs
phadej: Yes, makes `-Werror` useful and not only annoying ;)

cardano-crypto-class/memory-example/Main.hs
```diff
@@ -1,9 +1,13 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+
+{-# OPTIONS_GHC -Wno-deprecations #-}
```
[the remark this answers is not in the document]
phadej: use of traceMLockedFiniteBytes, which shouldn't be present in non-testing/debug mode. Pushed a commit adding a comment.

cardano-crypto-class/src/Cardano/Crypto/Libsodium/Constants.hsc
nc6: Yep, sorry, saw how you were using it below.

cardano-crypto-class/memory-example/Main.hs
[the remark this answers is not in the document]
phadej: Renamed

cardano-crypto-class/src/Cardano/Crypto/FiniteBytes.hs
```diff
@@ -0,0 +1,144 @@
...
+        res <- c_sodium_compare_unsafe (byteArrayContents x) (byteArrayContents y) size
+        return (compare res 0)
+      where
+        size :: CSize
+        size = fromInteger (natVal (Proxy :: Proxy n))
+
+-- |
+--
+-- If given 'String' is too long, it is truncated,
+-- If it is too short, it is padded with zeros.
```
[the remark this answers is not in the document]
phadej: Done

phadej:
part of #138

PR #133, opened 2020-06-23
Crypto benchmark

tdammers:
Benchmarking suite for the `cardano-crypto-praos` and `cardano-crypto-class` packages. Currently benchmarks only the Simple and Praos VRF implementations.


newhoggy:
Merged via https://github.com/input-output-hk/cardano-base/pull/180

PR #142, opened 2020-07-07
increase the size of ShortHash from 4 to 8 bytes

JaredCorduan:
Having `ShortHash` only have 4 bytes leads to the occasional property test failure resulting from a hash conflict.  I would like to increase it to 8 bytes, hopefully striking a middle ground between conveniently short but not too prone to conflicts.

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Mock.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Mock.hs
@@ -118,7 +118,7 @@ instance DSIGNAlgorithm MockDSIGN where
       = Nothing
 
     rawDeserialiseSigDSIGN bs
-      | [hb, kb] <- splitsAt [sizeHash (Proxy :: Proxy ShortHash), 8] bs
+      | [hb, kb] <- splitsAt [fromIntegral $ sizeHash (Proxy :: Proxy ShortHash), 8] bs
       , Just h   <- hashFromBytes hb
       , let k = readBinaryWord64 kb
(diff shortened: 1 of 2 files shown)


cardano-crypto-class/src/Cardano/Crypto/DSIGN/Mock.hs
```diff
@@ -118,7 +118,7 @@ instance DSIGNAlgorithm MockDSIGN where
       = Nothing
 
     rawDeserialiseSigDSIGN bs
-      | [hb, kb] <- splitsAt [4, 8] bs
+      | [hb, kb] <- splitsAt [8] bs
```
redxaxder, line 121:
```suggestion
      | [hb, kb] <- splitsAt [sizeHash (Proxy :: Proxy ShortHash), 8] bs
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Mock.hs
```diff
@@ -118,7 +118,7 @@ instance DSIGNAlgorithm MockDSIGN where
       = Nothing
 
     rawDeserialiseSigDSIGN bs
-      | [hb, kb] <- splitsAt [4, 8] bs
+      | [hb, kb] <- splitsAt [sizeHash (Proxy :: Proxy ShortHash), 8] bs
```
redxaxder, line 121:
```suggestion
      | [hb, kb] <- splitsAt [fromIntegral $ sizeHash (Proxy :: Proxy ShortHash), 8] bs
```

PR #150, opened 2020-07-20
Add hlint tests

newhoggy:


nc6:
Can you please provide more context as to what these things are and why we want them?

PR #157, opened 2020-08-18
Compile with GHC8.10.2

nc6:
This PR does not yet update CI to build with GHC 8.10, but makes various changes that bring the code into a place where it compiles with GHC 8.10. As such, downstream libraries may choose to build it with later GHC versions, and it should be easier to move to 8.10 when we decide to make the switch.


mrBliss:
It seems that the build failed while compiling GHC itself on `x86_64-w64-mingw32` :slightly_frowning_face:

nc6:
And a strange error about no `HasResolution` in the cardano-binary tests.

binary/cardano-binary.cabal
```diff
@@ -60,7 +60,12 @@ library
                        -Wredundant-constraints
                        -Wpartial-fields
                        -Wcompat
-
+  if impl(ghc >=8.8)
+    ghc-options:       -fno-warn-missing-deriving-strategies
+  if impl(ghc >=8.10)
+    ghc-options:       -fno-warn-missing-safe-haskell-mode
+                       -fno-warn-prepositive-qualified-module
+                       -fno-warn-unused-packages
```
mrBliss, line 68: I prefer `-Wno-X` over `-fno-warn-X`, then it's consistent with the warning flags we use above.

nc6: I have no opinion on this; I'm just taking the same approach as was used in `cardano-prelude`https://github.com/input-output-hk/cardano-prelude/blob/master/cardano-prelude.cabal#L79

mrBliss: Yeah, I saw that too. Would you mind if I pushed a commit to this PR that changed it to `-Wno-X`?

nc6: Not at all!

mrBliss: Done

binary/cardano-binary.cabal
```diff
@@ -60,7 +60,12 @@ library
                        -Wredundant-constraints
                        -Wpartial-fields
                        -Wcompat
-
+  if impl(ghc >=8.8)
+    ghc-options:       -Wno-missing-deriving-strategies
+  if impl(ghc >=8.10)
+    ghc-options:       -Wno-missing-safe-haskell-mode
+                       -Wno-prepositive-qualified-module
+                       -Wno-unused-packages
```
mrBliss, line 68: Actually, are all these flags that disable warnings even needed (in the other cabal files as well)? I assume you have copied them from `cardano-prelude`, but there `-Weverything` is used. Here `-Wall` is used, so these extra warnings don't have to be disabled. (I didn't have to disable them in https://github.com/input-output-hk/ouroboros-network/pull/2542)

I prefer `-Wall` + a manual selection of flags over `-Weverything`, which forces you to disable some flags based on the GHC version.

mrBliss (approved):
LGTM, I think you can remove these conditionals from the cabal files.

binary/cardano-binary.cabal
nc6: Hmm, certainly some of them were needed in some places. But it's true that after adding them to a couple of files, I pre-emptively added them in other places...

mrBliss: BTW, I wouldn't disable `-Wno-unused-packages`, I'd just fix the warnings :slightly_smiling_face:

mrBliss: I just tried removing all those flags myself and it compiled without warnings (except for those from the C code).

mrBliss: > I prefer `-Wall` + a manual selection of flags over `-Weverything`, which forces you to disable some flags based on the GHC version.

In other words, no changes are needed to the cabal files at all.

mrBliss:
I'm assuming you're planning on rebasing/squashing? It would be nice to get those two merge commits out of the history.

PR #161, opened 2020-09-15
Cardano.Crypto.KES using libsodium bindings

tdammers:
@phadej 's #138, rebased onto `master`, and fixing divergent development in the `Hash.Short` module.

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/KES/Class.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/KES/Class.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
@@ -5,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeApplications #-}
 
 -- | Abstract key evolving signatures.
 module Cardano.Crypto.KES.Class
@@ -43,14 +45,15 @@ import Data.Typeable (Typeable)
 import GHC.Exts (Constraint)
 import GHC.Generics (Generic)
 import GHC.Stack
+import GHC.TypeLits (Nat, KnownNat, natVal)
 
 import Cardano.Prelude (NoUnexpectedThunks)
 import Cardano.Binary (Decoder, decodeBytes, Encoding, encodeBytes, Size, withWordSize)
 
-import Cardano.Crypto.Seed
 import Cardano.Crypto.Util (Empty)
 import Cardano.Crypto.Hash.Class (HashAlgorithm, Hash, hashWith)
 
+import qualified Cardano.Crypto.Libsodium as NaCl
 
 class ( Typeable v
       , Show (VerKeyKES v)
@@ -61,9 +64,11 @@ class ( Typeable v
       , NoUnexpectedThunks (SigKES     v)
       , NoUnexpectedThunks (SignKeyKES v)
       , NoUnexpectedThunks (VerKeyKES  v)
+      , KnownNat (SeedSizeKES v)
       )
       => KESAlgorithm v where
 
+  type SeedSizeKES v :: Nat
 
   --
   -- Key and signature types
@@ -155,10 +160,11 @@ class ( Typeable v
   -- Key generation
   --
 
-  genKeyKES :: Seed -> SignKeyKES v
+  genKeyKES :: NaCl.MLockedSizedBytes (SeedSizeKES v) -> SignKeyKES v
 
   -- | The upper bound on the 'Seed' size needed by 'genKeyKES'
   seedSizeKES :: proxy v -> Word
+  seedSizeKES _ = fromInteger (natVal (Proxy @(SeedSizeKES v)))
 
   --
   -- Secure forgetting
--- a/cardano-crypto-class/src/Cardano/Crypto/KES/Simple.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/KES/Simple.hs
@@ -10,6 +10,7 @@
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE NoStarIsType #-}
 
 -- | Mock key evolving signatures.
 module Cardano.Crypto.KES.Simple
@@ -26,7 +27,7 @@ import qualified Data.ByteString as BS
 import           Data.Vector ((!?), Vector)
 import qualified Data.Vector as Vec
 import           GHC.Generics (Generic)
-import           GHC.TypeNats (Nat, KnownNat, natVal)
+import           GHC.TypeNats (Nat, KnownNat, natVal, type (*))
 
 import           Cardano.Prelude (NoUnexpectedThunks, forceElemsToWHNF)
 import           Cardano.Binary (FromCBOR (..), ToCBOR (..))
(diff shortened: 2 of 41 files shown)


cardano-crypto-class/src/Cardano/Crypto/Libsodium/UnsafeC.hs
```diff
@@ -0,0 +1,10 @@
+{-# LANGUAGE CApiFFI             #-}
+module Cardano.Crypto.Libsodium.UnsafeC (
+    c_sodium_compare_unsafe,
+    ) where
+
+import Foreign.C.Types (CSize (..))
+import Foreign.Ptr (Ptr)
+
+-- | Unsafe variant of 'c_sodium_compare'.
```
nc6, line 9: Would be nice to be clear on how exactly this is unsafe!

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -0,0 +1,229 @@
...
+    psbUseAsSizedPtr,
+    psbCreate,
+    psbCreateSized,
+    ptrPsbToSizedPtr,
+  ) where
+
+import Control.Monad.ST (runST)
+import Control.Monad.Primitive  (primitive_)
+import Data.Char (ord)
+import Data.Primitive.ByteArray (ByteArray (..), MutableByteArray (..), copyByteArrayToAddr, newPinnedByteArray, unsafeFreezeByteArray, foldrByteArray, byteArrayContents, writeByteArray, mutableByteArrayContents)
```
nc6, line 33: [style] long lines!

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -0,0 +1,229 @@
...
+
+-------------------------------------------------------------------------------
+-- derivative from primitive
+-------------------------------------------------------------------------------
+
+-- | Create a 'ByteArray' from a list of a known length. If the length
+--   of the list does not match the given length, this throws an exception.
+pinnedByteArrayFromListN :: forall a. Prim.Prim a => Int -> [a] -> ByteArray
+pinnedByteArrayFromListN n ys = runST $ do
+    marr <- newPinnedByteArray (n * Prim.sizeOf (head ys))
```
nc6, line 216: We should at least document that this will error for empty lists. `pinnedByteArrayFromListN 0 []` will fail :-)

nc6 (approved):
Approved, but left some comments that it might be nice to address.

Presumably as a result of this PR, any downstream user of KES _must_ now call `c_sodium_init` somewhere? Do we intend to leave this to the resulting application rather than wrapping it here?

cardano-crypto-class/src/Cardano/Crypto/KES/Class.hs
```diff
@@ -155,10 +160,11 @@ class ( Typeable v
   -- Key generation
   --
 
-  genKeyKES :: Seed -> SignKeyKES v
+  genKeyKES :: NaCl.MLockedSizedBytes (SeedSizeKES v) -> SignKeyKES v
```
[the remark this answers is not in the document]
tdammers: Agree in principle, but IIRC we've already had this discussion, and the conclusion was that any KES application that we would possibly want to use for real will require secure forgetting, and that, at least for the time being, the only implementation that can do this and that we will realistically implement in the foreseeable future, is the libsodium one; hence we decided to keep it like this, with the option to refactor later, if and when we need to.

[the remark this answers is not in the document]
phadej: `MLockedSizedBytes` is not really `libsodium` specific, here's just no other means (and maybe it is good thing) to create them.

cardano-crypto-class/src/Cardano/Crypto/Libsodium/UnsafeC.hs
phadej: foreign import capi *unsafe*

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
phadej: I asked about code style when working on the original PR, I got reply "there isn't".

cardano-crypto-class/src/Cardano/Crypto/KES/Simple.hs
```diff
@@ -193,31 +199,31 @@ instance DSIGNAlgorithm d => NoUnexpectedThunks (SigKES     (SimpleKES d t))
 instance DSIGNAlgorithm d => NoUnexpectedThunks (SignKeyKES (SimpleKES d t))
 instance DSIGNAlgorithm d => NoUnexpectedThunks (VerKeyKES  (SimpleKES d t))
 
-instance (DSIGNAlgorithm d, Typeable d, KnownNat t)
+instance (DSIGNAlgorithm d, Typeable d, KnownNat t, KnownNat (SeedSizeDSIGN d GHC.TypeLits.* t))
```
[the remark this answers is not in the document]
tdammers: Yes, absolutely.

cardano-crypto-class/src/Cardano/Crypto/KES/Sum.hs
```diff
@@ -90,9 +92,12 @@ type Sum7KES d h = SumKES h (Sum6KES d h)
 --
 data SumKES h d
 
-instance (KESAlgorithm d, HashAlgorithm h, Typeable d)
+instance ( KESAlgorithm d, NaCl.SodiumHashAlgorithm h, Typeable d
```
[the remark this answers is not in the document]
tdammers: Check.

cardano-crypto-class/src/Cardano/Crypto/Libsodium/UnsafeC.hs
tdammers: I think @Oleg would be the person to ask, but I would wager that the `import unsafe` part is what this is about. In a nutshell, the "safe" version (found in cardano-crypto-class/src/Cardano/Crypto/Libsodium/C.hs) is a "safe" import, so that suggests that the "safe" version may call back into Haskell, while the "unsafe" version may not (but will produce better performance, because it doesn't have to do the associated bookkeeping). Why this is necessary however, I don't know.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
tdammers: Fixed.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
tdammers: I'd actually prefer a more robust solution. I'll make it `die` explicitly for now, but it might be prudent to lift this constraint to the type level, using non-empty list types.

tdammers: Or we could just return a zero-length byte array in that case? Thoughts?

cardano-crypto-class/src/Cardano/Crypto/Libsodium/C.hs
```diff
@@ -0,0 +1,164 @@
...
+foreign import capi "sodium.h &sodium_free" c_sodium_free_funptr :: FunPtr (Ptr a -> IO ())
+
+-------------------------------------------------------------------------------
+-- Hashing: SHA256
+-------------------------------------------------------------------------------
+
+-- | @int crypto_hash_sha256(unsigned char *out, const unsigned char *in, unsigned long long inlen);@
+--
+-- <https://libsodium.gitbook.io/doc/advanced/sha-2_hash_function>
+foreign import capi "sodium.h crypto_hash_sha256" c_crypto_hash_sha256 :: SizedPtr CRYPTO_SHA256_BYTES -> Ptr CUChar -> CULLong -> IO Int
```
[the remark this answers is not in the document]
tdammers: Fixed.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
phadej: This is private function used in `psbFromBytes`. Don't overthink it.

nc6:
It looks like only the stack build is failing. Honestly, I suggest just removing the stack.yaml file and the stack build from the CI config, unless you particularly want to maintain it.

PR #164, opened 2020-09-19
Implement DSIGNAlgorithm instance for BIP32-Ed25519

intricate:


cardano-crypto-class/src/Cardano/Crypto/DSIGN/Bip32Ed25519.hs
```diff
@@ -0,0 +1,202 @@
...
+xPrvFromBytes :: ByteString -> Maybe CC.XPrv
+xPrvFromBytes bytes
+    | BS.length bytes /= 96 = Nothing
+    | otherwise = do
+        let (prv, cc) = BS.splitAt 64 bytes
+        pub <- ed25519ScalarMult (BS.take 32 prv)
+        eitherToMaybe $ CC.xprv $ prv <> pub <> cc
+  where
+    eitherToMaybe :: Either a b -> Maybe b
+    eitherToMaybe = either (const Nothing) Just
```
erikd, line 197: Is this not in `Cardano.Prelude`?

erikd (approved):
LGTM.

If `eitherToMaybe` is in `Cardano.Prelude` it probably best to use it rather than re-implementing it (even if it is trivial).

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Bip32Ed25519.hs
intricate: Nope, it's not. Maybe it should be, though.

nc6:
@intricate Can you clarify what the requirement for this is? We don't use this in Shelley, and I'm somewhat loathe to introduce the dependency on `cardano-crypto`, which was until this point relegated to Byron

intricate:
@nc6: The `cardano-crypto` BIP32-Ed25519 extended key implementation (`XPub`, `XPrv`, etc.) is actually used by a few projects, even in the Shelley era, e.g. `cardano-api`, `cardano-wallet`, and `cardano-addresses` all support these kinds of extended keys for Shelley.

The reason for wanting a `DSIGNAlgorithm` implementation in `cardano-base` would be for providing a common, familiar interface for other Cardano-related project to utilize. For example, at the moment, when working with extended keys in `cardano-api`, we have to explicitly make calls to `cardano-crypto` rather than utilizing the same `DSIGNAlgorithm` interface that we already use for regular Ed25519 crypto operations.

This implementation would also be useful in helping us standardize the binary serialization formats for BIP32-Ed25519 keys. For example, in the `cardano-crypto` implementation, BIP32-Ed25519 extended signing keys (`XPrv`) are serialized to a 128-byte binary representation whereas other projects utilize a more compact 96-byte representation (Jörmungandr, `cardano-addresses`, etc.). 

Going forward, we actually want to use the 96-byte representation throughout the Cardano ecosystem. In light of that, it seemed that it'd make sense to have it supported in `cardano-base` alongside the other crypto implementations rather than requiring the other projects to directly use the `cardano-crypto` types and provide their own potentially error-prone serialization and conversion functions.

intricate:
@nc6: After re-reading your comment, I think I agree that we shouldn't introduce a dependency on `cardano-crypto` here. I'm going to probably end up moving this instance into `cardano-api`.

dcoutts (commented):
This looks pretty good.

I've discussed this with @intricate and we agree with @nc6 that this is a useful instance but it should live elsewhere than in the cardano-crypto-classes.

It's true that the node itself never uses extended keys. This is one of the neat features of extended keys used for HD derivation: that the verification side (i.e. in the node) never needs to know anything about them, and they are a purely client/wallet concern.

The appropriate place for this instance is probably the cardano-api, where it can then be used by the cli, the wallet and other node clients.

@intricate note that when we do move it, we need to add the appropriate tests. See the tests for the other instances in this package to see what needs to be covered.

PR #165, opened 2020-09-23
Proposed forget-logging mockup KES wrapper (RFC)

tdammers:
In order to improve testability of the forgetfulness feature of the `KESAlgorithm` typeclass, we provide a wrapper type, `ForgetMockKES`, which wraps around any other KES implementation, and adds two logging actions, one in the `genKeyKES` method, and one in the `forgetSignKeyKES` method. Both emit a log message to stdout, consisting of a rendering of the key in question (as per `Show`), prefixed with `GEN:` or `DEL:` to indicate key generation and deletion, respectively.

The intended use case for this is to run KES-consuming code against this wrapped KES instance, capturing stdout over the lifecycle of the code; and then, by analyzing the output, pairing up `GEN` and `DEL` lines per unique key, we can check whether the code behaves as intended. Specifically, we can verify that for each generated key, the `forgetSignKeyKES` method is invoked exactly once.

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -43,6 +43,7 @@ library
 
                        Cardano.Crypto.KES.Class
                        Cardano.Crypto.KES.Mock
+                       Cardano.Crypto.KES.ForgetMock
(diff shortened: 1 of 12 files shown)


iquerejeta:
This is now part of #255. Closing.

PR #166, opened 2020-09-29
Remove stack build.

nc6:
This was not being maintained by anybody and causing breakages in PR. In
common with other repositories, remove the stack configuration and CI
build.

Note that this will need somebody with admin access to this repo (I do not have it) to remove the webhook for buildkite, which is no longer used.


mrBliss:
There is some Stack-related stuff in https://github.com/input-output-hk/cardano-base/tree/master/scripts/buildkite

PR #172, opened 2020-10-15
Add configuration for haskell-language-server.

nc6:


hie.yaml
```diff
@@ -0,0 +1,28 @@
+cradle:
+  cabal:
+    - path: "binary/src"
+      component: "lib:cardano-binary"
+
+    - path: "binary/test"
+      component: "cardano-binary:test:test"
+
+    - path: "binary/test/./"
+      component: "lib:cardano-binary-test"
```
mrBliss, line 10: Two components for the same path.

hie.yaml
```diff
@@ -0,0 +1,28 @@
...
+      component: "lib:cardano-crypto-praos"
+
+    - path: "cardano-crypto-tests/src"
+      component: "lib:cardano-crypto-tests"
+
+    - path: "cardano-crypto-tests/test"
+      component: "cardano-crypto-tests:test:test-crypto"
+
+    - path: "slotting/src"
+      component: "lib:cardano-slotting"
```
mrBliss, line 28: This is forcing users to use a multi-component cradle. You can locally modify this file, but that is annoying when switching branches/rebasing unless you tell Git to always ignore that file.

I find multi-component cradles have some downsides (see https://input-output-rnd.slack.com/archives/C48HA39RS/p1602514989082600?thread_ts=1602493286.060600&cid=C48HA39RS). Especially the fact that calling each call to `cabal build/run` on the CLI will require reconfiguring all packages in question :facepalm:.

For those reasons, I use a direct cradle. I'm not saying that everybody should use it, but maybe we should not make it harder for people to use their own custom `hie.yaml`? Maybe some weirdos would like to use Stack? :slightly_smiling_face: 

What about having a `hie-cabal.yaml` file that users can opt-in? They can use `ln -s hie-cabal.yaml hie.yaml` to opt-in.

nc6: Yeah, good suggestion!

mrBliss (approved):

PR #174, opened 2020-10-16
Avoid using broken api function base16.decode

newhoggy:
Broken as in broken API compatibility.  This change allows the code to compile against any version of `base16-bytestring`.  It is currently set to `0.1.1.7`.  In future this will be bumped to `1.0.0.0` requiring no code change.

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
@@ -44,6 +44,7 @@ module Cardano.Crypto.Hash.Class
   )
 where
 
+import Control.Monad (join)
 import Data.List (foldl')
 import Data.Maybe (maybeToList)
 import Data.Proxy (Proxy (..))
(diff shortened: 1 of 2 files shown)


cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
```diff
@@ -44,6 +44,7 @@ module Cardano.Crypto.Hash.Class
   )
 where
 
+import Control.Monad
```
erikd, line 47: You should be specific about what you are importing.

erikd (dismissed):

cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
newhoggy: Fixed!

intricate (approved):

edsko:
@newhoggy I am not working on this project anymore so please don't request reviews from me :)

PR #180, opened 2020-10-29
Benchmark VRF and KES. Enable benchmarks in Github Actions CI

newhoggy:
This is a rebase of an earlier PR https://github.com/input-output-hk/cardano-base/pull/133 with updates to build with latest code.

Benchmark VRF and KES.  Enable benchmarks in Github Actions CI.


newhoggy:
@tdammers I will need your help because after the rebase the benchmarks fail:

```
benchmarking KES/Sum6KES/genKey
Benchmark bench-crypto: ERROR
cabal: Benchmarks failed for bench:bench-crypto from
cardano-crypto-tests-2.0.0.
```

newhoggy:
The failure was caused by the benchmarks not calling `sodiumInit` at start up.  This has been fixed.

cardano-crypto-praos/src/Cardano/Crypto/VRF/Praos.hs
```diff
@@ -132,16 +133,25 @@ newtype SignKey = SignKey { unSignKey :: ForeignPtr SignKeyValue }
   deriving (Generic)
   deriving NoThunks via OnlyCheckWhnf SignKey
 
+instance NFData SignKey where
+  rnf a = seq a ()
+
```
erikd, line 138: Is it not possible to do?:
```
deriving instance NFData SignKey
```

erikd (approved):
Possibly worth removing the explicit `NFData` definitions.

cardano-crypto-praos/src/Cardano/Crypto/VRF/Praos.hs
newhoggy: This is the error:

```
 No instance for (NFData (ForeignPtr SignKeyValue))
```

PR #207, opened 2021-03-18
Drop support for ghc-8.6.5

newhoggy:
* Add lower-bounds to `base`.
* Remove redundant `if` conditionals from cabal files.


erikd (changes requested):
Needs the `base` lower bound bumped as well.

newhoggy:
> Needs the `base` lower bound bumped as well.

Fixed! Thanks!

erikd (approved):

PR #210, opened 2021-03-31
Add serialization tests for UTCTime

DavidEichmann:
I've split into 2 commits for clarity.

### DIFF
--- a/binary/test/Test/Cardano/Binary/Serialization.hs
+++ b/binary/test/Test/Cardano/Binary/Serialization.hs
@@ -16,7 +16,7 @@ import qualified Data.ByteString.Short as BS.Short
 
 import Cardano.Prelude
 
-import Hedgehog 
+import Hedgehog
 import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Range as Range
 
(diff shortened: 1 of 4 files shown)


DavidEichmann:
Does anyone know why we seem to have 2 `.cabal` files that declare the same testsuite?

DeepakKapiswe (approved):
LGTM

binary/test/Test/Cardano/Binary/Serialization.hs
```diff
@@ -147,6 +152,26 @@ instance FromCBOR TestStruct where
...
+    Gen.integral (Range.constantFrom 0 0 ((86401e12) - 1))
+
+genDay :: Gen Time.Day
+genDay = Time.fromOrdinalDate <$> genYear <*> genDayOfYear
+
+genYear :: Gen Integer
+genYear = Gen.integral (Range.linear (-10000) 10000)
+
+genDayOfYear :: Gen Int
+genDayOfYear = Gen.int (Range.constantFrom 1 1 366)
```
newhoggy, line 174: What happens in leap years?

newhoggy (approved):

newhoggy (approved):

binary/test/Test/Cardano/Binary/Serialization.hs
DavidEichmann: Yes, this is perhaps a bit subtle, but `fromOrdinalDate` should take care of clipping this in the case of non-leap years. I figured the random distribution having that small quirk was not worth fixing with extra complexity in the generator.

DavidEichmann:
bors merge

PR #212, opened 2021-04-07
Add time to EpochInfo

### LINKED ISSUE #205: Add Time to EpochInfo
The `Cardano.Slotting.EpochInfo.API.EpochInfo` record only includes conversions between slots and epochs.

For Alonzo/extended UTxO, the ledger will need conversions from a slot to a UTC time.

The HardFork Combinator currently only provides `EpochInfo` to the ledger. This ticket proposes providing an enriched variant of `EpochInfo` that also includes slot-to-time conversions (`EpochTimeInfo`?), and then for CAD-2713 the HFC can provide that to the ledger.

nfrisby:
Fixes #205.

### DIFF
--- a/slotting/cardano-slotting.cabal
+++ b/slotting/cardano-slotting.cabal
@@ -38,11 +38,13 @@ common project-config
 library
   import:               base, project-config
 
-  exposed-modules:      Cardano.Slotting.Block
-                        Cardano.Slotting.Slot
+  exposed-modules:
+                        Cardano.Slotting.Block
                         Cardano.Slotting.EpochInfo
                         Cardano.Slotting.EpochInfo.API
                         Cardano.Slotting.EpochInfo.Impl
+                        Cardano.Slotting.Slot
+                        Cardano.Slotting.Time
 
   build-depends:        aeson
                       , base
@@ -51,5 +53,6 @@ library
                       , deepseq
                       , mmorph
                       , nothunks
-                      , serialise
                       , quiet
+                      , serialise
+                      , time             >=1.9.1 && <1.11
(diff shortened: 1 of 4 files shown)


nfrisby:
Everything in the new `Cardano.Slotting.Time` has been migrated verbatim from `Ouroboros.Consensus.Blockchain.WallClock.Types`; this PR merely upstreams these newtypes/conversions/etc.

The `EpochInfo` cannot (easily) return `UTCTime` directly because the underlying machinery in the HFC, which ultimately builds the `EpochInfo`, does not depend on the `SystemStart`.

Some mocks and/or tests in downstream Ledger and Consensus have been using `EpochInfo` in contexts that have not explicitly required a `SlotLength`. The simplest integration option there would be to simply invent a dummy slot length for each such invocation, since existing `EpochInfo` use cases cannot depend on the `SlotLength`. However, I instead introduced the "shortcuts" in `.Impl` so that the existing use-cases in mocks & tests can remain oblivious of `SlotLength`.

An alternative would be to leave `EpochInfo` alone and instead create a separate `TimeInfo` record. The HFC and Ledger interface would have to change to exchange the pair `(EpochInfo m, TimeInfo m)`. All the mock&tests could remain untouched. However, I think it's reasonable for `EpochInfo` to be monolothic and representative of the actual use-case's context (which will always have both an epoch size and a slot length) and to patch-up the tests instead.

nfrisby:
I converted this to Draft PR because I want to hear back from the Ledger Team that the content here is sufficient for their https://jira.iohk.io/browse/CAD-2684 needs. That's the only thing I'm waiting for though. I'm therefore adding @polinavino as a reviewer here.

nc6 (approved):
LGTM, and I think this will satisfy our requirements.

polinavino:
LGTM also, thank you!

PR #220, opened 2021-06-25
Defunctionalize EpochInfo and add To/FromCBOR instances

DavidEichmann:
I'm trying (still :-/ ) to add a query on the cardano-node for the ledger configuration. In the case of a Shelley node, that requires sending the `Globals` and hence a `EpochInfo` which currently contains functions and is not serializable! With this PR I'm defunctionalizing `EpochInfo` so it is just data and is now serializable.

Is this safe to do, or do we in fact need the more general functional representation? I've searched through all occurrences of the string "EpochInfo" in the https://github.com/input-output-hk/cardano-haskell repo. I've only found 3 places where we construct an `EpochInfo` and I think they are compatible with this refactoring:
* See https://github.com/input-output-hk/cardano-base/pull/220#pullrequestreview-695609516
* `dummyEpochInfo` ouroboros-network/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/EpochInfo.hs
  * all functions throw errors
* `hardForkEpochInfo` ouroboros-network/ouroboros-consensus-test/test-consensus/Test/Consensus/HardFork/History.hs
  * all functions throw errors
* `fixedEpochInfo` cardano-base/slotting/src/Cardano/Slotting/EpochInfo/Impl.hs
  * Super easy to defunctionalize! The arguments to this function have now become the fields of the `EpochInfo` type.

### DIFF
--- a/slotting/cardano-slotting.cabal
+++ b/slotting/cardano-slotting.cabal
@@ -51,7 +51,6 @@ library
                       , cardano-binary
                       , cborg
                       , deepseq
-                      , mmorph
                       , nothunks
                       , quiet
                       , serialise
--- a/slotting/src/Cardano/Slotting/EpochInfo/API.hs
+++ b/slotting/src/Cardano/Slotting/EpochInfo/API.hs
@@ -1,121 +1,90 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 
 module Cardano.Slotting.EpochInfo.API
   ( EpochInfo (..),
-    epochInfoSize,
     epochInfoFirst,
     epochInfoEpoch,
     epochInfoRange,
     epochInfoSlotToRelativeTime,
     epochInfoSlotToUTCTime,
-
-    -- * Utility
-    hoistEpochInfo,
-    generalizeEpochInfo,
   )
 where
 
+import Cardano.Binary
+import Cardano.Slotting.EpochInfo.Impl
 import Cardano.Slotting.Slot (EpochNo (..), EpochSize (..), SlotNo (..))
-import Cardano.Slotting.Time (RelativeTime, SystemStart, fromRelativeTime)
-import Control.Monad.Morph (generalize)
-import Data.Functor.Identity
+import Cardano.Slotting.Time
 import Data.Time.Clock (UTCTime)
-import GHC.Stack (HasCallStack)
-import NoThunks.Class (NoThunks, OnlyCheckWhnfNamed (..))
+import GHC.Generics
+import NoThunks.Class (NoThunks)
 
--- | Information about epochs
---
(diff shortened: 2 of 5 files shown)


nfrisby (changes requested):
`EpochInfo` will definitely need more information than just one `EpochSize` and one `SlotLength`.

I agree that it doesn't need to be a monadic function, but it _will_ have to answer questions about a time span over which the epoch size and/or slot length has varied. (EG It was 20 seconds per slot in Byron and it's 1 second per slot in Shelley -- and a single `EpochInfo` needs to do calculations that are influenced by that historical change.)

Here is where Consensus currently constructs such a rich `EpochInfo`.

https://github.com/input-output-hk/ouroboros-network/blob/b749bc878bb05478f924d6e2d4970d1bdaef94ef/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/Combinator/State.hs#L147-L157

Please schedule a call with me if you'd like. It may be that we want the serializable `cardano-base` definition to take inspiration from the HFC data that underlies the above function. You can find that here https://github.com/input-output-hk/ouroboros-network/blob/c82309f403e99d916a76bb4d96d6812fb0a9db81/ouroboros-consensus/src/Ouroboros/Consensus/HardFork/History/Summary.hs#L198 I think the ultimately necessary data is simpler the above but also more complicated than just `[(EpochSize, SlotLength, NumEpochs)]`.

DavidEichmann:
Indeed you're right. I've missed that very important case! Ironically it's right next to one of the other cases I did find. If I flatten out a `Summary`, it is isomorphic to (ignoring the empty list case):

```
[( -- Era Start (inclusive)
  RelativeTime  -- The NominalDiffTime ?since system start?
, SlotNo
, EpochNo
  -- EraEnd (exclusive, Nothing means unbounded)
, Maybe
  ( RelativeTime  -- The NominalDiffTime ?since system start?
  , SlotNo
  , EpochNo
  )
  -- EraParams
, EpochSize
, SlotLength
, SafeZone
)]
```

The invariants documented above `data EraSummary` mean that the era end's `RelativeTime` and `SlotNo` can be derived from the era start and the era end's epoch number (or equivalently the `NumEpochs` as you hinted in your comment @nfrisby). I also suspect that we don't need the `SafeZone` information. Let's simplify to:

```
[( -- Era Start (inclusive)
  RelativeTime  -- The NominalDiffTime ?since system start?
, SlotNo
, EpochNo
  -- EraEnd (exclusive, Nothing means unbounded)
, Maybe EpochNo
  -- EraParams
, EpochSize
, SlotLength
)]
```
I wonder if `Summary` has these 2 invariants:

* Each era-end equals the the next era-start
* Eras are in chronological order.

For repesenting `EpochInfo` I think we can weaken these 2 invariants to simply:

* The eras must not overlap

This implies that:

* At most 1 era has an open era-end bound

I might suggest this representation:


```
data EpochInfo = EpochInfo
  -- Eras with closed bounds
  [( -- Era Start (inclusive)
    RelativeTime  -- The NominalDiffTime ?since system start?
  , SlotNo
  , EpochNo
    -- EraEnd (exclusive, Nothing means unbounded)
  , EpochNo
    -- EraParams
  , EpochSize
  , SlotLength
  )]

  -- Optional final era with open end bound
```

DavidEichmann:
This is no longer necessary. I proposed this as a way to serialize `EpochInfo`, but serialization has been resolve in https://github.com/input-output-hk/ouroboros-network/pull/3214.

PR #225, opened 2021-07-16
Add a type class for measures

nfrisby:
See the individual commits and their messages.

### DIFF
--- a/base-deriving-via/src/Data/DerivingVia/GHC/Generics/Semigroup.hs
+++ b/base-deriving-via/src/Data/DerivingVia/GHC/Generics/Semigroup.hs
@@ -1,0 +1,44 @@
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE EmptyCase            #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE LambdaCase           #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# OPTIONS -Wno-unticked-promoted-constructors #-}
+
+-- | "GHC.Generics" definition of '<>'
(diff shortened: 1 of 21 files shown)


measures/src/Data/Measure.hs
```diff
@@ -0,0 +1,38 @@
+{-# LANGUAGE BangPatterns      #-}
+{-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Data.Measure (
```
nfrisby, line 5: Add comment here

base-deriving-via/src/Data/DerivingVia/GHC/Generics/Semigroup.hs
```diff
@@ -0,0 +1,42 @@
...
+  gsappend :: rep x -> rep x -> rep x
+
+instance Monoid c => GSemigroup (K1 i c) where
+  gsappend (K1 l) (K1 r) = K1 (l <> r)
+
+instance GSemigroup f => GSemigroup (M1 i c f) where
+  gsappend (M1 l) (M1 r) = M1 (gsappend l r)
+
+instance GSemigroup V1 where
+  gsappend = error "GSemigroup V1"
```
coot, line 28: Should we use a `TypeError` here as well?

coot: Actually it shouldn't be an error, `Void` has a valid `Semigroup` instance.

nfrisby: Thanks @coot; that was a copy-paste error. Nice catch.

dcoutts (approved):

nc6 (approved):
LGTM

nfrisby:
My latest updates were because I didn't realize how many times I had replicated the mindless `V1` mistake that @coot caught. I think I got them all now.

nc6:
bors merge

PR #227, opened 2021-08-05
Fix memory liveness with touch.

lehins:
It is unsafe to operate on pointers without ensuring RTS keeps the `ByteArray#` alive. This PR fixes that problem.

There is still another small problem in `psbUseAsCPtr` and `psbUseAsSizedPtr`, but `keepAlive#` is only available in GHC-9.0 and marking those functions with `NOINLINE` will incur unnecessary performance penalty. 

Here is a recent blog post on that topic: https://well-typed.com/blog/2021/06/keepAlive-story/


dcoutts (approved):
Good catch.

I would suggest using one of the "with" style wrappers rather than explicit `byteArrayContents` and `touch`.

dcoutts (approved):

PR #228, opened 2021-08-16
Bump nix infra

nc6:


nix/sources.json
```diff
@@ -22,17 +36,5 @@
         "type": "tarball",
         "url": "https://github.com/input-output-hk/iohk-nix/archive/65d5b5eabf5252c7ece10efdc507f0a5c9b2aa8a.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
-    },
-    "nixpkgs": {
```
redxaxder, line 26: Should we be removing this key?
I saw the tooling giving advice to get rid of it, but I'm not sure what the consequences are.

nc6: Consequences are that we use the `nixpkgs` specified in the `iohk-nix` revision. It allows for more sharing.

redxaxder (approved):

PR #229, opened 2021-08-24
Specialize the hash representation of Blake2b

redxaxder:


cardano-crypto-class/src/Cardano/Crypto/Hash/Blake2b.hs
```diff
@@ -14,24 +17,123 @@ where
...
+   x3 = Prim.indexByteArray# byteArray 3#
+   x4 = Prim.indexByteArray# byteArray 4#
+   x5 = Prim.indexByteArray# byteArray 5#
+   x6 = Prim.indexByteArray# byteArray 6#
+
+-- | Converts a hash rep into a ShortByteString
+blake2b224HashToShort :: Blake2b_224Hash -> ShortByteString
+blake2b224HashToShort (Blake2b_224Hash x0 x1 x2 x3 x4 x5 x6) = SBSI.SBS byteArray
+  where
+  !(BA.ByteArray byteArray) = BA.byteArrayFromListN 7 [x0,x1,x2,x3,x4,x5,x6]
```
lehins, line 79: Relying on `byteArrayFromListN` here will be a lot more expensive then it has to be because of extra list allocation and an unnecessary loop. Mutable approach is a much better alternative:
```suggestion
  !(BA.ByteArray byteArray) = runST $ do
      mba <- newByteArray 28
      writeByteArray mba 0 x0
      writeByteArray mba 1 x1
      writeByteArray mba 2 x2
      writeByteArray mba 3 x3
      writeByteArray mba 4 x4
      writeByteArray mba 5 x5
      writeByteArray mba 6 x6
      unsafeFreezeByteArray mba
```

cardano-crypto-class/src/Cardano/Crypto/Hash/Blake2b.hs
```diff
@@ -14,24 +17,123 @@ where
...
+   x4 = Prim.indexByteArray# byteArray 4#
+   x5 = Prim.indexByteArray# byteArray 5#
+   x6 = Prim.indexByteArray# byteArray 6#
+   x7 = Prim.indexByteArray# byteArray 7#
+
+-- | Converts a hash rep into a ShortByteString
+blake2b256HashToShort :: Blake2b_256Hash -> ShortByteString
+blake2b256HashToShort (Blake2b_256Hash x0 x1 x2 x3 x4 x5 x6 x7) = SBSI.SBS byteArray
+  where
+  !(BA.ByteArray byteArray) = BA.byteArrayFromListN 8 [x0,x1,x2,x3,x4,x5,x6,x7]
```
lehins, line 122: Same suggestion as for `blake2b224HashToShort`

cardano-crypto-class/src/Cardano/Crypto/Hash/Blake2b.hs
```diff
@@ -14,24 +17,123 @@ where
...
 import GHC.IO.Exception (ioException)
+import qualified Data.ByteString.Short.Internal as SBSI
+import qualified Data.ByteString.Short as SBS
+import Data.ByteString.Short (ShortByteString)
+import Data.Word (Word64)
 
 import qualified Data.ByteString as B
 import qualified Data.ByteString.Internal as BI
+import           Control.DeepSeq (NFData)
+import           NoThunks.Class (NoThunks)
```
lehins, line 32: Inconsistent import formatting

lehins (changes requested):
Except for relying on list, LGTM.

cardano-crypto-class/src/Cardano/Crypto/Hash/Blake2b.hs
```diff
@@ -14,24 +17,128 @@ where
...
 
 data Blake2b_224
 data Blake2b_256
 
+data PackedBytes32 =
+  PackedBytes32
+    {-# UNPACK #-} !Word64
+    {-# UNPACK #-} !Word64
+    {-# UNPACK #-} !Word64
+    {-# UNPACK #-} !Word64
```
lehins, line 45: You can save here an extra 4 bytes if you use: [`Word32#`](https://downloads.haskell.org/~ghc/8.10.4/docs/html/libraries/base-4.14.1.0/GHC-Exts.html#t:Word32-35-) for `Blake2b_224`

You'd have to use indexing and writing from `GHC.Exts` though, since those aren't default types until ghc-9.2

* [`indexWord8ArrayAsWord32#`](https://downloads.haskell.org/~ghc/8.10.4/docs/html/libraries/base-4.14.1.0/GHC-Exts.html#v:indexWord8ArrayAsWord32-35-)
* [`writeWord8ArrayAsWord32#`](https://downloads.haskell.org/~ghc/8.10.4/docs/html/libraries/base-4.14.1.0/GHC-Exts.html#v:writeWord8ArrayAsWord32-35-)

PR #230, opened 2021-08-27
Alternative hash packing approach

redxaxder:


cardano-crypto-class/src/Cardano/Crypto/Hash/Packed.hs
```diff
@@ -0,0 +1,66 @@
...
+    {-# UNPACK #-} !Word64
+    {-# UNPACK #-} !Word64
+    {-# UNPACK #-} !Word64
+  deriving (Eq, Generic)
+
+
+instance Ord Bytes32 where
+  -- Note: the derived ord instance would produce different orderings
+  -- on machines with different byte orders.
+  compare x y = compare (unpack x) (unpack y)
```
lehins, line 26: This is a huge problem. It will affect performance tremendously, therefore this approach ins not acceptable!!!
Consider that on each insert into a `Map` or a `Set` there will be 2*log(n) conversions to SBS in order to perform a comparison.

cardano-crypto-class/src/Cardano/Crypto/Hash/Packed.hs
```diff
@@ -0,0 +1,66 @@
...
+import           Data.ByteString.Short (ShortByteString)
+import qualified Data.ByteString.Short.Internal as SBSI
+import           NoThunks.Class (NoThunks)
+import           GHC.Generics (Generic)
+import           Data.Word (Word64, Word8)
+import qualified Control.Monad.ST as ST
+import           Control.DeepSeq (NFData)
+
+data Bytes32 = Bytes32
+    {-# UNPACK #-} !Word8 -- length
```
lehins, line 15: This will add not 1 byte extra. but 8 bytes. because `data Word8 = W8# Word#` and because it will be aligned to the bigger `Word64` type that follows anyways.

cardano-crypto-class/src/Cardano/Crypto/Hash/Packed.hs
redxaxder: What do you suggest?

Machine-dependent ordering isn't acceptable either.

cardano-crypto-class/src/Cardano/Crypto/Hash/Packed.hs
redxaxder: True. I'm not sure why you're telling me this.

cardano-crypto-class/src/Cardano/Crypto/Hash/Packed.hs
lehins: Implementing a fast machine independent ordering without conversion to SBS. This will be a major pain, but it is doable

cardano-crypto-class/src/Cardano/Crypto/Hash/Packed.hs
lehins: I am telling you this because that is the overhead that we can avoid,

redxaxder: I mean that I thought we already both know the amount of the overhead this approach would involve.

I'm not sure zero overhead is possible without a drastic impact on upstream uses. Your alternative has the same drastic impact as the original, and for the same reasons.

lehins: Define "drastic impact on downstream uses". I don't think going through the compiler errors and adding a constraint is such a big of a deal. Especially since we want to squeeze as much savings from this as possible, don't we?

PR #231, opened 2021-08-27
Implement PackedBytes for more compact representation of hashes

lehins:
Alternative solution to #229 and #230

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
@@ -1,18 +1,22 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DefaultSignatures #-}
-{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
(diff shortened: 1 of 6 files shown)


cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
```diff
@@ -122,20 +105,25 @@ class
...
+deriving via OnlyCheckWhnfNamed "Hash" (Hash h a) instance NoThunks (Hash h a)
 
-pattern UnsafeHash :: forall h a. HashAlgorithm h => ShortByteString -> Hash h a
-pattern UnsafeHash bytes <- UnsafeHashRep (fromHashRep (Proxy :: Proxy h) -> bytes)
-  where
-  UnsafeHash bytes = UnsafeHashRep (unsafeToHashRep (Proxy :: Proxy h) bytes)
+instance NFData (Hash h a) where
+  rnf = rwhnf
+
+instance KnownNat (SizeHash h) => Eq (Hash h a) where
```
redxaxder, line 116: I thought the purpose of the change was to no longer have a constraint bubbling up from the Eq instance.

lehins:
I'll close this PR. In case we aren't happy with savings in #232 we can revive this one to get that extra machine word from unpacking and proper 28byte support.

PR #234, opened 2021-09-08
remove HashAlgorithm constraint from Show Hash

JaredCorduan:
And also add a `HeapWords` instance for `Hash`.


redxaxder (approved):

lehins (changes requested):
This is a very dangerous thing to do! There is a much safer way.

JaredCorduan:
> This is a very dangerous thing to do! There is a much safer way.

What do you suggest? I suspect you'll need to know what we are doing with the constructor?

JaredCorduan:
@lehins the export is removed, instead I've added a `HeapWords` instance for `Hash` and we will deal with the other problem another way.

lehins (approved):
:+1:

PR #235, opened 2021-09-12
Serialise packed bytes

kk-hainq:
`ouroboros-network` has an orphan `Serialise` instance on `Hash`:
https://github.com/input-output-hk/ouroboros-network/blob/22d9b995e2c302fb14532aa459bc6b81627267bc/ouroboros-consensus/src/Ouroboros/Consensus/Util/Orphans.hs#L64

That gets propagated to places like this:
https://github.com/input-output-hk/ouroboros-network/blob/22d9b995e2c302fb14532aa459bc6b81627267bc/ouroboros-consensus-mock/src/Ouroboros/Consensus/Mock/Ledger/UTxO.hs#L70-L72

Since `PackedBytes` isn't exported (intentionally?), I wonder if adding the `Serialise` instance here makes sense?
If this is the way to go we'd love to add some tests as well.

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -103,6 +103,7 @@ library
                       , memory
                       , nothunks
                       , primitive
+                      , serialise
                       , text
                       , transformers
                       , vector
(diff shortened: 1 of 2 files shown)


lehins (approved):
> Since PackedBytes isn't exported (intentionally?)

Yep, intentionally. Thank you for getting rid of an orphan instance

nc6:
bors merge

nc6:
Merged through #237

PR #239, opened 2021-09-27
Fix for a potential memory leak

lehins:
This PR can help avoid potential memory leaks in places where Hash is not forced after conversion from `ByteString`/`ShortByteString`


nc6 (approved):

dcoutts (commented):
LGTM

PR #240, opened 2021-10-01
Remove overlapping rewrite rules.

nc6:
As per
https://downloads.haskell.org/~ghc/7.0.3/docs/html/users_guide/rewrite-rules.html,
if GHC encounters more than one matching rewrite rule, it will choose
arbitrarily which one fires. As such, the rewrite rules converting
`pack*Bytes` to `pack*BytesN` are problematic - they are always
applicable, and hence may fire in place of the specific ones.

Consequently, we remove them. Since they are optimisations anyway, this
should not change the semantics of the system.


lehins (approved):
Nice catch @nc6!

PR #241, opened 2021-10-05
Make sure UnsafeHash is not as dangerous

lehins:
It is currently possible to construct an utterly dangerous hash value that can cause segfaults and buffer overflows. It is better to blow up with an error rather than construct a dangerous hash that could lead to a vulnerability. Cost of checking the length is justified in this case IMHO.


coot (approved):
LGTM

PR #244, opened 2021-10-28
Change KES signature format to be more compact

tdammers:
Change KES signature representation and serialization format.

This is so that we can reduce the size of KES signatures.

Previously, we stored in the signature:

- At the leaf node level (SingleKES):
    - The DSIGN signature of the leaf
- At the branch node level (SumKES):
    - The signature of the 'active' child node (the subtree containing the current period)
    - The DSIGN VK's of both child nodes

This means that for a Merkle tree N levels deep, we store 1 DSIGN signature + 2N-1 DSIGN VK's.

In the new implementation, we store:

- At the leaf node level (SingleKES):
    - The DSIGN signature of the leaf
    - The DSIGN VK for that signature
- At the branch node level (SumKES):
    - The signature of the 'active' child node
    - The DSIGN VK for the 'inactive' child node

This representation reduces storage size to 1 DSIGN signature + N DSIGN VK's.

For performance reasons, we now split the verification process into two steps:

1. verifyDSIGN on the 'sigma' part of the signature. We include the VerKey for the leaf node in the signature itself, and just verify against that. This is not sufficient as validation on its own, hence...
2. ...recursively calculate the root VerKey along the Merkle path down to the leaf node, and compare it against the VerKey to verify against.

The advantage of this is that we only ever calculate the hashes we
absolutely need, and avoid duplicate calculation efforts.

In order to facilitate migrating from the old representation to the new one,
the old SumKES has been left (mostly) unchanged, and the new implementation has
been added under the name CompactSumKES (with CompactSingleKES as the leaf
node type).

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/KES/CompactSingle.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/KES/CompactSingle.hs
@@ -1,0 +1,221 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | A standard signature scheme is a forward-secure signature scheme with a
+-- single time period.
+--
+-- This is the base case in the naive recursive implementation of the sum
+-- composition from section 3 of the \"MMM\" paper:
+--
+-- /Composition and Efficiency Tradeoffs for Forward-Secure Digital Signatures/
+-- By Tal Malkin, Daniele Micciancio and Sara Miner
+-- <https://eprint.iacr.org/2001/034>
+--
+-- Specfically it states:
+--
+-- > In order to unify the presentation, we regard standard signature schemes
+-- > as forward-seure signature schemes with one time period, namely T = 1.
+--
+-- So this module simply provides a wrapper 'CompactSingleKES' that turns any
+-- 'DSIGNAlgorithm' into an instance of 'KESAlgorithm' with a single period.
+--
+-- See "Cardano.Crypto.KES.CompactSum" for the composition case.
+--
+-- Compared to the implementation in 'Cardano.Crypto.KES.Single', this flavor
+-- stores the VerKey used for signing along with the signature. The purpose of
+-- this is so that we can avoid storing a pair of VerKeys at every branch node,
+-- like 'Cardano.Crypto.KES.Sum' does. See 'Cardano.Crypto.KES.CompactSum' for
+-- more details.
+module Cardano.Crypto.KES.CompactSingle (
+    CompactSingleKES
+  , VerKeyKES (..)
+  , SignKeyKES (..)
+  , SigKES (..)
+  ) where
+
+import Data.Proxy (Proxy(..))
+import Data.Typeable (Typeable)
+import GHC.Generics (Generic)
+import NoThunks.Class (NoThunks)
+import qualified Data.ByteString as BS
+import           Control.Monad (guard)
+
+import Control.Exception (assert)
+
+import Cardano.Binary (FromCBOR (..), ToCBOR (..))
+
+import Cardano.Crypto.Hash.Class
+import Cardano.Crypto.DSIGN.Class
+import qualified Cardano.Crypto.DSIGN as DSIGN
+import Cardano.Crypto.KES.Class
+import Control.DeepSeq (NFData)
+
+
+-- | A standard signature scheme is a forward-secure signature scheme with a
+-- single time period.
+--
+data CompactSingleKES d
+
+deriving newtype instance NFData (VerKeyDSIGN d) => NFData (VerKeyKES (CompactSingleKES d))
+deriving newtype instance NFData (SignKeyDSIGN d) => NFData (SignKeyKES (CompactSingleKES d))
+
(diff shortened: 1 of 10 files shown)


cardano-crypto-class/src/Cardano/Crypto/KES/CompactSum.hs
```diff
@@ -0,0 +1,400 @@
...
+    -- | From Figure 3: @(sk_0, r_1, vk_0, vk_1)@
+    --
+    data SignKeyKES (CompactSumKES h d) =
+           SignKeyCompactSumKES !(SignKeyKES d)
+                         !Seed
+                         !(VerKeyKES d)
+                         !(VerKeyKES d)
+        deriving Generic
+
+    -- | From Figure 3: @(sigma, vk_0, vk_1)@
```
iquerejeta, line 162: We should maybe update this comment.

cardano-crypto-class/src/Cardano/Crypto/KES/CompactSum.hs
```diff
@@ -0,0 +1,400 @@
...
+    --
+    data SigKES (CompactSumKES h d) =
+           SigCompactSumKES !(SigKES d) -- contains VerKey for this branch!
```
iquerejeta, line 165: Does it? The `SigKES` does not contain the `VerKey` of the whole tree. It simply contains the DSIGN verkey, right? It does contain the data to be able to compute it though.

iquerejeta: It is only the `CompactSingle` that contains the `VerKey` of its own branch/leaf.

iquerejeta (commented):
Can't comment for the haskell code, but changes are sound. We check that the `VerKey` of the DSIGN instance of a KES signature is part of some expected MerkleTree (defined by a public key). Then we verify that the DSIGN signature is valid against that particular key.

iquerejeta:
@dcoutts , if you don't have cycles to review this, who should look into it? Thanks

nc6:
I'll review

cardano-crypto-tests/src/Test/Crypto/KES.hs
```diff
@@ -43,17 +43,25 @@ tests =
   , testKESAlgorithm (Proxy :: Proxy (Sum1KES Ed25519DSIGN Blake2b_256)) "Sum1KES"
   , testKESAlgorithm (Proxy :: Proxy (Sum2KES Ed25519DSIGN Blake2b_256)) "Sum2KES"
   , testKESAlgorithm (Proxy :: Proxy (Sum5KES Ed25519DSIGN Blake2b_256)) "Sum5KES"
+  , testKESAlgorithm (Proxy :: Proxy (CompactSingleKES Ed25519DSIGN))  "SingleKES"
+  , testKESAlgorithm (Proxy :: Proxy (CompactSum1KES Ed25519DSIGN Blake2b_256)) "Sum1KES"
+  , testKESAlgorithm (Proxy :: Proxy (CompactSum2KES Ed25519DSIGN Blake2b_256)) "Sum2KES"
+  , testKESAlgorithm (Proxy :: Proxy (CompactSum5KES Ed25519DSIGN Blake2b_256)) "Sum5KES"
```
lehins, line 49:
```suggestion
  , testKESAlgorithm (Proxy :: Proxy (CompactSingleKES Ed25519DSIGN))  "CompactSingleKES"
  , testKESAlgorithm (Proxy :: Proxy (CompactSum1KES Ed25519DSIGN Blake2b_256)) "CompactSum1KES"
  , testKESAlgorithm (Proxy :: Proxy (CompactSum2KES Ed25519DSIGN Blake2b_256)) "CompactSum2KES"
  , testKESAlgorithm (Proxy :: Proxy (CompactSum5KES Ed25519DSIGN Blake2b_256)) "CompactSum5KES"
```

cardano-crypto-class/src/Cardano/Crypto/KES/CompactSum.hs
```diff
@@ -0,0 +1,400 @@
...
+                     | otherwise = (vk_other, verKeyFromSigKES ctxt t' sigma)
+
+hashPairOfVKeys :: (KESAlgorithm d, HashAlgorithm h)
+                => (VerKeyKES d, VerKeyKES d)
+                -> Hash h (VerKeyKES d, VerKeyKES d)
+hashPairOfVKeys =
+    hashWith $ \(a,b) ->
+      rawSerialiseVerKeyKES a <> rawSerialiseVerKeyKES b
+
+slice :: Word -> Word -> ByteString -> ByteString
```
nc6, line 327: I feel like we should move these into a shared module, rather than repeating them all over the place.

nc6 (approved):
Made a few suggestions, but otherwise looks good.

cardano-crypto-class/src/Cardano/Crypto/KES/CompactSingle.hs
```diff
@@ -0,0 +1,221 @@
...
+  , SignKeyKES (..)
+  , SigKES (..)
+  ) where
+
+import Data.Proxy (Proxy(..))
+import Data.Typeable (Typeable)
+import GHC.Generics (Generic)
+import NoThunks.Class (NoThunks)
+import qualified Data.ByteString as BS
+import           Control.Monad (guard)
```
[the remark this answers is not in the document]
tdammers: Copy-pasted from Single.hs, and there it is probably the result of a merge, or further copy-pasting in an earlier commit. Agree that it's odd, but if we change it, then we should do it in both places.

cardano-crypto-tests/src/Test/Crypto/KES.hs
tdammers: Good catch!

cardano-crypto-class/src/Cardano/Crypto/KES/CompactSum.hs
tdammers: Correct. Fixing.

cardano-crypto-class/src/Cardano/Crypto/KES/CompactSum.hs
tdammers: Done.

cardano-crypto-class/src/Cardano/Crypto/KES/CompactSum.hs
tdammers: Agree. Will do.

nc6 (approved):

PR #246, opened 2021-11-19
Fix compatibility with aeson-2

michaelpj:
This change works both pre- and post-2.0.


michaelpj:
I haven't bumped the local index-state, so the CI will still build with the old version of aeson, but I checked locally by adding `constraints: aeson >= 2.0.0.0` to `cabal.project` that it worked. I'm not too worried about someone accidentally undoing this, since it's a single fairly obscure change.

lehins (approved):
Yeah, that should work just fine.

However, later on I'd recommend setting `constraints: aeson >= 2.0.1.0` in `cabal.project`, since `2.0.0.0` still uses `HashMap` underneath by default, while `2.0.1.0` onward uses `Map`

michaelpj:
I think we want people to move to aeson 2.0 eventually, but the point of this fix is that it won't _force_ downstream projects to do that.

PR #247, opened 2021-11-23
Upgrade VRF to IETF Draft 09

tdammers:


nc6:
I've rebased and updated this. I think CI should now pass, let's see.

PR #250, opened 2021-12-13
ghcjs support

angerman:


angerman:
I'll close this for now due to staleness.

PR #252, opened 2022-01-17
SECP256k1 support

kozross:
This is the first stage of [solving this issue](https://github.com/input-output-hk/plutus/issues/4233), as per [this request](https://github.com/input-output-hk/plutus/issues/4233#issuecomment-976718634). 

A few questions remain outstanding:

* <s>The size of the verification key (that is, public key) is not an exact multiple of 8 bits, as it's compressed and DER-encoded, for a total of 257 bits. I've thus set it to be 33 bytes in the implementation, but I am unsure if this is correct.</s>
* Secret key generation has to be implemented partially, as the original implementation returns in a `Maybe`. According to the implementation, we ensure that the only invariant is met, but this seems suspicious.

Tagging @michaelpj for review.

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -98,6 +98,7 @@ library
                       , bytestring
                       , cardano-binary
                       , cardano-prelude
+                      , cereal
                       , cryptonite
                       , deepseq
                       , integer-gmp
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -31,7 +31,6 @@ common project-config
                         -Wincomplete-uni-patterns
                         -Wpartial-fields
                         -Wredundant-constraints
-                        -Wunused-packages
 
   if (!flag(development))
     ghc-options:        -Werror
@@ -106,6 +105,7 @@ library
                       , nothunks
                       , primitive
                       , serialise
+                      , secp256k1-haskell
                       , text
                       , transformers
                       , vector
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -31,6 +31,7 @@ common project-config
                         -Wincomplete-uni-patterns
                         -Wpartial-fields
                         -Wredundant-constraints
+                        -Wunused-packages
 
   if (!flag(development))
     ghc-options:        -Werror
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -49,6 +49,7 @@ library
                         Cardano.Crypto.DSIGN.Ed448
                         Cardano.Crypto.DSIGN.Mock
                         Cardano.Crypto.DSIGN.NeverUsed
+                        Cardano.Crypto.DSIGN.SECP256k1
 
                         Cardano.Crypto.Hash.Blake2b
                         Cardano.Crypto.Hash.Class
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/SECP256k1.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/SECP256k1.hs
@@ -20,14 +20,36 @@ import GHC.Generics (Generic)
 import Control.DeepSeq (NFData)
 import qualified Crypto.Secp256k1 as SECP
 import NoThunks.Class (NoThunks)
-import Cardano.Crypto.DSIGN.Class (DSIGNAlgorithm (VerKeyDSIGN, SignKeyDSIGN, SigDSIGN,
-  SeedSizeDSIGN, SizeSigDSIGN, SizeSignKeyDSIGN, SizeVerKeyDSIGN, algorithmNameDSIGN,
-  deriveVerKeyDSIGN, signDSIGN, verifyDSIGN, genKeyDSIGN, rawSerialiseSigDSIGN,
-  Signable, rawSerialiseVerKeyDSIGN, rawSerialiseSignKeyDSIGN, rawDeserialiseVerKeyDSIGN,
-  rawDeserialiseSignKeyDSIGN, rawDeserialiseSigDSIGN), encodeVerKeyDSIGN, 
-  encodedVerKeyDSIGNSizeExpr, decodeVerKeyDSIGN, encodeSignKeyDSIGN, 
-  encodedSignKeyDESIGNSizeExpr, decodeSignKeyDSIGN, encodeSigDSIGN, 
-  encodedSigDSIGNSizeExpr, decodeSigDSIGN)
+import Cardano.Crypto.DSIGN.Class (
+  DSIGNAlgorithm (VerKeyDSIGN, 
+                  SignKeyDSIGN, 
+                  SigDSIGN,
+                  SeedSizeDSIGN, 
+                  SizeSigDSIGN, 
+                  SizeSignKeyDSIGN, 
+                  SizeVerKeyDSIGN, 
+                  algorithmNameDSIGN,
+                  deriveVerKeyDSIGN, 
+                  signDSIGN, 
+                  verifyDSIGN, 
+                  genKeyDSIGN, 
+                  rawSerialiseSigDSIGN,
+                  Signable, 
+                  rawSerialiseVerKeyDSIGN, 
+                  rawSerialiseSignKeyDSIGN, 
+                  rawDeserialiseVerKeyDSIGN,
+                  rawDeserialiseSignKeyDSIGN, 
+                  rawDeserialiseSigDSIGN), 
+  encodeVerKeyDSIGN, 
+  encodedVerKeyDSIGNSizeExpr, 
+  decodeVerKeyDSIGN, 
+  encodeSignKeyDSIGN, 
+  encodedSignKeyDESIGNSizeExpr, 
+  decodeSignKeyDSIGN, 
+  encodeSigDSIGN, 
+  encodedSigDSIGNSizeExpr, 
+  decodeSigDSIGN
+  )
 
 data SECP256k1DSIGN
 
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/SECP256k1.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/SECP256k1.hs
@@ -1,0 +1,58 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# OPTIONS_GHC -Wno-orphans #-} -- need NoThunks for secp256k1-haskell types
+
+module Cardano.Crypto.DSIGN.SECP256k1 where
+
+import Data.Kind (Type)
+import GHC.Generics (Generic)
+import Control.DeepSeq (NFData)
+import qualified Crypto.Secp256k1 as SECP
+import NoThunks.Class (NoThunks)
+import Cardano.Crypto.DSIGN.Class (DSIGNAlgorithm (VerKeyDSIGN, SignKeyDSIGN, SigDSIGN,
+  SeedSizeDSIGN, SizeSigDSIGN, SizeSignKeyDSIGN, SizeVerKeyDSIGN, algorithmNameDSIGN,
+  deriveVerKeyDSIGN, signDSIGN, verifyDSIGN, genKeyDSIGN, rawSerialiseSigDSIGN,
+  Signable))
+
+data SECP256k1DSIGN
+
+instance NoThunks (VerKeyDSIGN SECP256k1DSIGN)
+
+instance NoThunks (SignKeyDSIGN SECP256k1DSIGN)
+
+instance NoThunks (SigDSIGN SECP256k1DSIGN)
+
+instance DSIGNAlgorithm SECP256k1DSIGN where
+  type SeedSizeDSIGN SECP256k1DSIGN = 32
+  type SizeSigDSIGN SECP256k1DSIGN = 72
+  type SizeSignKeyDSIGN SECP256k1DSIGN = 32
+  type SizeVerKeyDSIGN SECP256k1DSIGN = 33 -- approximate, as it's 257 bits
+  type Signable SECP256k1DSIGN = ((~) SECP.Msg)
+  newtype VerKeyDSIGN SECP256k1DSIGN = VerKeySECP256k1 SECP.PubKey
+    deriving newtype (Eq, NFData)
+    deriving stock (Show, Generic)
+  newtype SignKeyDSIGN SECP256k1DSIGN = SignKeySECP256k1 SECP.SecKey
+    deriving newtype (Eq, NFData)
+    deriving stock (Show, Generic)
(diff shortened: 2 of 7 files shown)


cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -31,7 +31,6 @@ common project-config
                         -Wincomplete-uni-patterns
                         -Wpartial-fields
                         -Wredundant-constraints
```
tdammers, line 33: I would suggest leaving these things in place; these warning flags are there for a purpose, and I don't think there is a good reason to accept unused packages in a release build here.

You can use cabal.project.local to override things for local development.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SECP256k1.hs
```diff
@@ -0,0 +1,113 @@
...
+import Data.ByteString (ByteString)
+import Crypto.Random (getRandomBytes)
+import Cardano.Crypto.Seed (runMonadRandomWithSeed)
+import Data.Serialize (Serialize (get, put), runPut, runGet)
+import Data.Kind (Type)
+import GHC.Generics (Generic)
+import Control.DeepSeq (NFData)
+import qualified Crypto.Secp256k1 as SECP
+import NoThunks.Class (NoThunks)
+import Cardano.Crypto.DSIGN.Class (DSIGNAlgorithm (VerKeyDSIGN, SignKeyDSIGN, SigDSIGN,
```
tdammers, line 23: Consider formatting these as one per line - like this, they're harder to read, and small changes will lead to more diff noise.

tdammers (changes requested):
Could you also add this new DSIGN algorithm to the test suite? It may not be obvious, but if you look in the `cardano-crypto-tests` subproject, there's a file `src/Test/Crypto/DSIGN.hs`, and adding the new algorithm is as simple as adding a line to the `tests` data structure. You can pretty much just copy one of the existing lines and change the algorithm name to SECP256k1.

cardano-crypto-class/cardano-crypto-class.cabal
kozross: Sorry, what happened with these? Did I accidentally remove them?

kozross:
@tdammers I actually can't add the tests, due to this error:

```
src/Test/Crypto/DSIGN.hs:37:7: error:
    • Couldn't match type ‘(~)
                             secp256k1-haskell-0.5.0:Crypto.Secp256k1.Msg’
                     with ‘SignableRepresentation’
        arising from a use of ‘testDSIGNAlgorithm’
    • In the expression:
        testDSIGNAlgorithm (Proxy :: Proxy SECP256k1DSIGN) "SECP-256k1"
      In the second argument of ‘testGroup’, namely
        ‘[testDSIGNAlgorithm (Proxy :: Proxy MockDSIGN) "MockDSIGN",
          testDSIGNAlgorithm (Proxy :: Proxy Ed25519DSIGN) "Ed25519DSIGN",
          testDSIGNAlgorithm (Proxy :: Proxy Ed448DSIGN) "Ed448DSIGN",
          testDSIGNAlgorithm (Proxy :: Proxy SECP256k1DSIGN) "SECP-256k1"]’
      In the expression:
        testGroup
          "Crypto.DSIGN"
          [testDSIGNAlgorithm (Proxy :: Proxy MockDSIGN) "MockDSIGN",
           testDSIGNAlgorithm (Proxy :: Proxy Ed25519DSIGN) "Ed25519DSIGN",
           testDSIGNAlgorithm (Proxy :: Proxy Ed448DSIGN) "Ed448DSIGN",
           testDSIGNAlgorithm (Proxy :: Proxy SECP256k1DSIGN) "SECP-256k1"]
   |
37 |     , testDSIGNAlgorithm (Proxy :: Proxy SECP256k1DSIGN) "SECP-256k1"
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This is caused by the fact that `SignableRepresentation` is 'too big' a type (class) for what the implementation in `secp256k1-haskell` can sign: we can only sign an opaque `Msg` type, which is what that constraint is designed to enforce. I'm happy to repair the tests to deal with this, but it's possible I'm misunderstanding something.

cardano-crypto-class/cardano-crypto-class.cabal
michaelpj: I think he meant to comment on the line below. You removed `-Wunused-packages`.

michaelpj:
It does seem odd to have the `Signable` type family in the class and then force it to all be the same constraint in the tests. If we want to do that we should maybe just get rid of `Signable`...

Presumably @kozross there's a way to make a `Msg` from a `ByteString`, and `SignableRepresentation` lets you get a `ByteString`, so you could use that as your `Signable` constraint...

kozross:
@michaelpj Yes, you can make a `Msg` from a `ByteString`, but it's subject to constraints on size. Making SECP256k1's `Signable` be `SignableRepresentation` is inappropriate: you can't just jam any old `ByteString` in there, but the _very first_ instance of `SignableRepresentation` is `ByteString`! 

The primary issue here isn't _what_ the `Signable` constraint is: the issue is that it _must_ be `SignableRepresentation`, or the tests can't handle it. Basically, you've got two things at cross purposes here; only one of these assumptions can work presently without something being changed to accommodate the other:

* `Signable` is meant to be choosable per-algorithm, to work within any specific restrictions of the algorithm or its implementation. This is indeed the case here: we can only sign `Msg`. If so, the tests are written with inappropriate assumptions and have to be modified not to assume `SignableRepresentation` everywhere.
* `Signable` is meant to _always_ be `SignableRepresentation`, in which case you must either be able to reject inputs as not signable by the algorithm, or forbid anything that restricts what can be signed. In this case, why make the `Signable` constraint definable in the type class?

I'm happy to fix either issue, but I'd rather know which would be acceptable.

michaelpj:
I agree with your analysis, and personally I think it would be preferable for the tests to not assume that `Signable ~ SignableRepresentation`. What do you think @tdammers ?

kozross:
@michaelpj I've already worked out a fix to the tests which makes this assumption unnecessary. I still have some minor fixes to make that all the SECP256k1 tests pass, but once that's done, it _should_ be ready.

michaelpj:
I noticed you made some changes to the sizes, was that because they were wrong?

michaelpj (approved):
LGTM, so long as @tdammers is happy with the test changes.

kozross:
> I noticed you made some changes to the sizes, was that because they were wrong?

Yes: the tests also caught me, so it's good that @tdammers told me about them.

kozross:
@michaelpj I've addressed the conflicts - is this OK to merge?

michaelpj:
Tobias said he wanted to have a look at this today, so I'm waiting for that.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
```diff
@@ -11,17 +11,122 @@ module Test.Crypto.DSIGN
...
+  (===), 
+  Arbitrary(..), 
+  Gen, 
+  Property,
+  forAllShow
+  )
 import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.QuickCheck (testProperty)
 
+-- Captures ways of generating each of the associated types of DSIGN
```
tdammers, line 77: Why is it necessary to also provide a generalization of genVK and genSK, when we never use it? Can we not simply pass a `Gen (SigDSIGN v)`?

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
```diff
@@ -31,166 +136,159 @@ import Test.Tasty.QuickCheck (testProperty)
...
+  (DSIGNAlgorithm v, Signable v a, ContextDSIGN v ~ ()) 
+  => (a, SignKeyDSIGN v, SignKeyDSIGN v) 
   -> Property
-prop_dsign_verify_neg_key a sk sk' =
-  sk /= sk' ==>
-    let sig = signDSIGN () a sk
-        vk' = deriveVerKeyDSIGN sk'
-    in verifyDSIGN () vk' a sig =/= Right ()
-
+prop_dsign_verify_wrong_key (msg, sk, sk') =
```
tdammers, line 279: Removing the `sk /= sk'` precondition can lead to false negatives - when the "wrong" key happens to be identical to the correct key, the verification will of course succeed, and the test fails.

Without shrinking, this is very unlikely (by design), but I still think it's prudent to keep the precondition.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
```diff
@@ -31,166 +136,159 @@ import Test.Tasty.QuickCheck (testProperty)
...
-
-instance (DSIGNAlgorithm v,
-          ContextDSIGN v ~ (), Signable v ~ SignableRepresentation)
-      => Arbitrary (SigDSIGN v) where
-  arbitrary = do
-    a <- arbitrary :: Gen Message
-    sk <- arbitrary
-    return $ signDSIGN () a sk
-  shrink = const []
+prop_dsign_verify_wrong_msg (msg, msg', sk) =
```
tdammers, line 291: As with `prop_dsign_verify_wrong_key`, the `/=` precondition matters; it probably matters even more here, because messages are subject to more aggressive shrinking (or at least, they can be). E.g., if messages are `String`s, then the shrinker will come up with the test case where both messages are the empty string.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
tdammers: Oh, and this design also prevents shrinking, doesn't it?

tdammers (commented):
This isn't going to break anything, but I'm not 100% convinced that this is the best possible design here.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
kozross: We do use them. Consider line ranges 167-173, 180-187, 195-201, 208-209, 213-214, 218-223 at least. We also have several derived generators using one or both.

As far as preventing shrinking goes - we can't shrink anyway for almost anything, and indeed, the original design _also_ prevented shrinking, since all shrinkers were defined as `const []`. My changes have not removed any functionality or capability that existed before, but _added_ plenty.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
kozross: This precondition is not required, as the generator is now _guaranteed_ not to give equal keys. See `genWrongKey` for how this is done.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
kozross: Same as before; see `genWrongMsg`.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
tdammers: Ah, sorry about the confusion - yes, we are using the (quasi-)methods, but we are not using the ability to generalize - all the methodologies we define here use the same implementations for `genVerKey` and `genSignKey`, so we might as well just use `default{Ver,Sign}KeyGen`.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
```diff
@@ -11,17 +11,122 @@ module Test.Crypto.DSIGN
...
+  Property,
+  forAllShow
+  )
 import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.QuickCheck (testProperty)
 
+-- Captures ways of generating each of the associated types of DSIGN
+data TypesMethodology (a :: Type) = 
+  TypesMethodology (Gen (VerKeyDSIGN a))
+                   (Gen (SignKeyDSIGN a))
```
michaelpj, line 80: If we're going to do this, arguably we could get rid of the key generation method from the `DSIGN` class, since it's presumably only used for testing, and this lets us give the generators per-type as we need.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
michaelpj: If we did what I suggest below then we would need `genSignKey`, but perhaps still not `genVerKey`. Not sure it costs us much, though.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
kozross: I'm happy to make that change, `DSIGN` is already huge.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
kozross: @tdammers I'm happy to specialize in the way you describe. I don't think it's a particularly bad idea to have the _ability_ to generalize in the way the verification and sign keys are generated, even if we don't use it currently, but I'm not strongly wedded to this.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
kozross: This is potentially quite a breaking change, so I'm reluctant to do this unless absolutely necessary. I've fixed Tobias' original over-generality concern without this, but if you feel a change to `DSIGN` is needed to get this merged, I can do it.

michaelpj: Nope, not necessary, it just seems ugly to me to have the generator in there.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
tdammers: OK, makes sense.

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
tdammers: I totally understand the thinking here, it just tickles my YAGNI nerve. I would personally start with only generalizing what needs to be generalized, which, at this point, is only `genSig` bit.

And, upon further thought, what we're *really* generalizing here isn't even generating a *signature*, but generating a *message* to sign. If you compare the `defSigGen` function with the `go` function inside the `secp256k1Methodology`, the only difference is how the message is generated.

So what I would propose is that we ditch the `TypesMethodology` type entirely, and instead just pass a generator for the message type (`genSECPMsg` for secp256k1, `arbitrary` for everything else), and then implement `genSig` in terms of that.

tdammers (approved):
Remaining complaints from my side are largely cosmetic; as far as I'm concerned, we can merge now, and maybe bikeshed later.

michaelpj:
:+1:

iquerejeta:
Naming in this PR is again incorrect. We should not use `secp256k1SigGen`, or more generally, we should not use `secp256k1` as the name of the signature algorithm, as it is **not** a signature algorithm. This has been referred to incorrectly since the [first issue](https://github.com/input-output-hk/plutus/issues/4233), and should be resolved. In my opinion, the naming should be corrected not only in the code but also in the issues and PR descriptions.

kozross:
@iquerejeta What do you suggest instead?

iquerejeta:
I would suggest using the signature algorithm's name. I'm not sure which one you have implemented here, but if it is ECDSA, I would go with something like `EcdsaSecp256k1SigGen`, and if its Schnorr, I would use `SchnorrSecp256k1SigGen`. I think keeping the curve in the name is good, so that the distinction between the same algorithm over different curves is explicit in the naming.

kozross:
@iquerejeta Thank you - that's a good solution. @michaelpj - can I combine the name-change with the Schnorr signature PR I'm working on, or do you want that done separately.

michaelpj:
Happy for them both to go together.

PR #257, opened 2022-02-18
Add nixpkgs pin, bump haskell.nix

michaelpj:


michaelpj:
Hydra is unhappy, but I built this locally and it's got most of the way, so I'm pretty confident.

michaelpj:
And it builds.

michaelpj:
@nc6 would you mind approving?

michaelpj:
Oh dear, Haskell CI has been failing this way on master already...

nc6 (approved):
Approving for now to unblock, but we should aim to drop the separate nixpkgs pin and come back to using `nixpkgs` from `haskell.nix` before we update the base ref upstream.

PR #258, opened 2022-02-21
Schnorr signature support

kozross:
This forms the first step in addressing [this requirement](https://github.com/input-output-hk/plutus/pull/4368#issuecomment-1031782137), as well as addressing [this naming concern](https://github.com/input-output-hk/cardano-base/pull/252#issuecomment-1042724920). The test suite has been extended to cover this case as well.

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
@@ -8,7 +8,12 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# OPTIONS_GHC -Wno-orphans #-} -- need NoThunks for secp256k1-haskell types
 
-module Cardano.Crypto.DSIGN.EcdsaSecp256k1 where
+module Cardano.Crypto.DSIGN.EcdsaSecp256k1 (
+  EcdsaSecp256k1DSIGN,
+  VerKeyDSIGN (..),
+  SignKeyDSIGN (..),
+  SigDSIGN (..)
+  ) where
 
 import Cardano.Binary (FromCBOR (fromCBOR), ToCBOR (toCBOR, encodedSizeExpr))
 import Data.ByteString (ByteString)
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
@@ -23,7 +23,7 @@ import Data.Serialize (Serialize (get, put), runPut, runGet)
 import Data.Kind (Type)
 import GHC.Generics (Generic)
 import Control.DeepSeq (NFData)
-import qualified Crypto.Secp256k1 as SECP
+import qualified Crypto.Secp256k1 as ECDSA
 import NoThunks.Class (NoThunks)
 import Cardano.Crypto.DSIGN.Class (
   DSIGNAlgorithm (VerKeyDSIGN, 
@@ -69,31 +69,31 @@ instance DSIGNAlgorithm EcdsaSecp256k1DSIGN where
   type SizeSigDSIGN EcdsaSecp256k1DSIGN = 64
   type SizeSignKeyDSIGN EcdsaSecp256k1DSIGN = 32
   type SizeVerKeyDSIGN EcdsaSecp256k1DSIGN = 64
-  type Signable EcdsaSecp256k1DSIGN = ((~) SECP.Msg)
+  type Signable EcdsaSecp256k1DSIGN = ((~) ECDSA.Msg)
   newtype VerKeyDSIGN EcdsaSecp256k1DSIGN = 
-    VerKeyEcdsaSecp256k1 SECP.PubKey
+    VerKeyEcdsaSecp256k1 ECDSA.PubKey
     deriving newtype (Eq, NFData)
     deriving stock (Show, Generic)
   newtype SignKeyDSIGN EcdsaSecp256k1DSIGN = 
-    SignKeyEcdsaSecp256k1 SECP.SecKey
+    SignKeyEcdsaSecp256k1 ECDSA.SecKey
     deriving newtype (Eq, NFData)
     deriving stock (Show, Generic)
   newtype SigDSIGN EcdsaSecp256k1DSIGN = 
-    SigEcdsaSecp256k1 SECP.Sig
+    SigEcdsaSecp256k1 ECDSA.Sig
     deriving newtype (Eq, NFData)
     deriving stock (Show, Generic)
   algorithmNameDSIGN _ = "ecdsa-secp256k1"
   deriveVerKeyDSIGN (SignKeyEcdsaSecp256k1 sk) = 
-    VerKeyEcdsaSecp256k1 . SECP.derivePubKey $ sk
+    VerKeyEcdsaSecp256k1 . ECDSA.derivePubKey $ sk
   signDSIGN () msg (SignKeyEcdsaSecp256k1 k) = 
-    SigEcdsaSecp256k1 . SECP.signMsg k $ msg
+    SigEcdsaSecp256k1 . ECDSA.signMsg k $ msg
   verifyDSIGN () (VerKeyEcdsaSecp256k1 pk) msg (SigEcdsaSecp256k1 sig) = 
-    if SECP.verifySig pk sig msg
+    if ECDSA.verifySig pk sig msg
     then pure ()
     else Left "ECDSA-SECP256k1 signature not verified"
   genKeyDSIGN seed = runMonadRandomWithSeed seed $ do
     bs <- getRandomBytes 32
-    case SECP.secKey bs of 
+    case ECDSA.secKey bs of 
       Nothing -> error "Failed to construct a ECDSA-SECP256k1 secret key unexpectedly"
       Just sk -> pure . SignKeyEcdsaSecp256k1 $ sk
   rawSerialiseSigDSIGN (SigEcdsaSecp256k1 sig) = putting sig
@@ -129,11 +129,11 @@ instance FromCBOR (SigDSIGN EcdsaSecp256k1DSIGN) where
 
 -- Required orphans
 
-instance NoThunks SECP.PubKey
+instance NoThunks ECDSA.PubKey
 
-instance NoThunks SECP.SecKey
+instance NoThunks ECDSA.SecKey
 
-instance NoThunks SECP.Sig
+instance NoThunks ECDSA.Sig
 
 -- Helpers
 
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
@@ -8,7 +8,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# OPTIONS_GHC -Wno-orphans #-} -- need NoThunks for secp256k1-haskell types
 
-module Cardano.Crypto.DSIGN.SECP256k1 where
+module Cardano.Crypto.DSIGN.EcdsaSecp256k1 where
 
 import Cardano.Binary (FromCBOR (fromCBOR), ToCBOR (toCBOR, encodedSizeExpr))
 import Data.ByteString (ByteString)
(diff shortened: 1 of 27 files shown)


kozross:
@michaelpj and @iquerejeta - does this seem good?

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -0,0 +1,296 @@
...
+  SECP256k1XOnlyPubKey, 
+  secpKeyPairCreate, 
+  SECP256k1Context, 
+  secpKeyPairXOnlyPub,
+  SECP256k1SecKey, 
+  secpSchnorrSigVerify, 
+  secpContextSignVerify,
+  SECP256k1SchnorrSig, 
+  secpSchnorrSigSignCustom, 
+  secpContextCreate
```
iquerejeta, line 54: Some reason why some of these functions contain `Schnorr`, and other don't? or why some contain `secp` and others `SECP256k1`? And a nit, have we changed the order of the prefixes in purpose? In here we go `secp256k1schnorr` while above we go with `schnorrsecp256k1`.

cardano-crypto-class/src/Cardano/Crypto/Schnorr.hs
```diff
@@ -0,0 +1,88 @@
...
+secpContextSignVerify = secpContextSign .|. secpContextVerify
+
+data SECP256k1SecKey
+
+data SECP256k1SchnorrSig
+
+data SECP256k1KeyPair
+
+foreign import capi "secp256k1_extrakeys.h secp256k1_keypair_create"
+  secpKeyPairCreate ::
```
iquerejeta, line 55: are we excluding the prefix because the function is the same for both algorithms?

iquerejeta (commented):
Naming looks good. I've left some comments motivated by consistency. Note that I'm not a haskell expert. But if we decide to have some generic names (such as `secpKeyPairCreate` or `SECP256k1SecKey`), would it make sense to do qualified imports to make these types or functions explicitly linked to each algorithm?

cardano-crypto-class/src/Cardano/Crypto/Schnorr.hs
```diff
@@ -0,0 +1,88 @@
+{-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Cardano.Crypto.Schnorr (
```
michaelpj, line 5: Hmmm, do you think we could contribute this upstream to `secp256k1-haskell`?

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -0,0 +1,296 @@
...
+        )
+  newtype SigDSIGN SchnorrSecp256k1DSIGN = 
+    SigSchnorr256k1 (ForeignPtr SECP256k1SchnorrSig)
+      deriving NoThunks via (
+        OnlyCheckWhnfNamed "SigSchnorr256k1" (ForeignPtr SECP256k1SchnorrSig)
+        )
+  algorithmNameDSIGN _ = "schnorr-secp256k1"
+  {-# NOINLINE deriveVerKeyDSIGN #-}
+  deriveVerKeyDSIGN (SignKeySchnorr256k1 fp) = 
+    unsafeDupablePerformIO . withForeignPtr fp $ \skp -> do
```
michaelpj, line 115: Do we want to isolate this stuff in a module that provides a nicer Haskell interface, more like `secp256k1-haskell` provides?

michaelpj (commented):
I am extremely unqualified to review the FFI stuff, we should find someone to do that.

cardano-crypto-class/src/Cardano/Crypto/Schnorr.hs
kozross: We could, but I chose not to, in the interests of getting this out the door faster. Possibly useful for the future though.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
kozross: I don't feel this would help much. When it comes to FFI stuff, my policy is to bind as closely as possible to the C API, then wrap it directly. Adding another level of indirection in between doesn't strike me as useful, and has actually backfired quite badly in the past.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -0,0 +1,296 @@
...
+data SchnorrSecp256k1DSIGN
+
+instance DSIGNAlgorithm SchnorrSecp256k1DSIGN where
+  type SeedSizeDSIGN SchnorrSecp256k1DSIGN = 32
+  type SizeSigDSIGN SchnorrSecp256k1DSIGN = 64
+  type SizeSignKeyDSIGN SchnorrSecp256k1DSIGN = 32
+  type SizeVerKeyDSIGN SchnorrSecp256k1DSIGN = 64
+  type Signable SchnorrSecp256k1DSIGN = SignableRepresentation
+  newtype VerKeyDSIGN SchnorrSecp256k1DSIGN = 
+    VerKeySchnorr256k1 (ForeignPtr SECP256k1XOnlyPubKey)
```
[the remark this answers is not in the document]
kozross: Fair enough, I can do that.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
```diff
@@ -51,67 +56,75 @@ import Cardano.Crypto.DSIGN.Class (
...
-  type Signable SECP256k1DSIGN = ((~) SECP.Msg)
-  newtype VerKeyDSIGN SECP256k1DSIGN = VerKeySECP256k1 SECP.PubKey
+instance DSIGNAlgorithm EcdsaSecp256k1DSIGN where
+  type SeedSizeDSIGN EcdsaSecp256k1DSIGN = 32
+  type SizeSigDSIGN EcdsaSecp256k1DSIGN = 64
+  type SizeSignKeyDSIGN EcdsaSecp256k1DSIGN = 32
+  type SizeVerKeyDSIGN EcdsaSecp256k1DSIGN = 64
+  type Signable EcdsaSecp256k1DSIGN = ((~) SECP.Msg)
+  newtype VerKeyDSIGN EcdsaSecp256k1DSIGN = 
+    VerKeyEcdsaSecp256k1 SECP.PubKey
```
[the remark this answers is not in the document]
kozross: Yes, they are. I could rename the import if that would be clearer, but I figured it wouldn't matter much.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
kozross: This is because they mirror the naming of the C functions they wrap (without `256k1` everywhere, admittedly).

cardano-crypto-class/src/Cardano/Crypto/Schnorr.hs
kozross: This function is not algorithmically-tied. The `secp256k1_extrakeys.h` header provides general keypair-related manipulation: it is not specific to Schnorr or ECDSA signatures.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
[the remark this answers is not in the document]
iquerejeta: Oh, the advantage is simply the consistency of the code. It is a nit, but after all we are making this PR to use correct naming. IMO we should be consistent, and maintain correctness of naming throughout the code.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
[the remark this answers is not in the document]
iquerejeta: The problem I see is that in the import
```haskell
import qualified Crypto.Secp256k1 as SECP
```
there is no mention whatsoever to the actual signing algorithm. I would change the import to 
```haskell
import qualified Crypto.EcdsaSecp256k1 as SECP
```
or, if it is not possible, to
```haskell
import qualified Crypto.Secp256k1 as ECDSA
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -0,0 +1,296 @@
...
+    SigSchnorr256k1 (ForeignPtr SECP256k1SchnorrSig)
+      deriving NoThunks via (
+        OnlyCheckWhnfNamed "SigSchnorr256k1" (ForeignPtr SECP256k1SchnorrSig)
+        )
+  algorithmNameDSIGN _ = "schnorr-secp256k1"
+  {-# NOINLINE deriveVerKeyDSIGN #-}
+  deriveVerKeyDSIGN (SignKeySchnorr256k1 fp) = 
+    unsafeDupablePerformIO . withForeignPtr fp $ \skp -> do
+      let skp' :: Ptr CUChar = castPtr skp
+      allocaBytes 96 $ \kpp -> do
```
michaelpj, line 117: I note that this package has some utilities for slightly safer allocation, in [here](https://github.com/input-output-hk/cardano-base/blob/master/cardano-crypto-class/src/Cardano/Foreign.hs). I think you can use this in your foreign declarations (like [here](https://github.com/input-output-hk/cardano-base/blob/39f9cf0eaf15a6888a47ac9d8de0b37dd1131d0b/cardano-crypto-class/src/Cardano/Crypto/Libsodium/C.hs#L90)) to get some more safety and fewer magic numbers.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -0,0 +1,296 @@
...
+  {-# NOINLINE signDSIGN #-}
+  signDSIGN () msg (SignKeySchnorr256k1 skfp) = 
+    unsafeDupablePerformIO . withForeignPtr skfp $ \skp -> do
+      let bs = getSignableRepresentation msg
+      let skp' :: Ptr CUChar = castPtr skp
+      allocaBytes 96 $ \kpp -> do
+        res <- secpKeyPairCreate ctxPtr kpp skp'
+        when (res /= 1) (error "signDSIGN: Failed to create keypair")
+        sigFP <- mallocForeignPtrBytes 64
+        let (msgFP, msgOff, msgLen) = toForeignPtr bs
```
michaelpj, line 134: I think you could replace many (all?) the uses of `toForeignPointer` with `useAsCString(Len)`, which doesn't require the bytestring `Internal` module.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -0,0 +1,296 @@
...
+        OnlyCheckWhnfNamed "SigSchnorr256k1" (ForeignPtr SECP256k1SchnorrSig)
+        )
+  algorithmNameDSIGN _ = "schnorr-secp256k1"
+  {-# NOINLINE deriveVerKeyDSIGN #-}
+  deriveVerKeyDSIGN (SignKeySchnorr256k1 fp) = 
+    unsafeDupablePerformIO . withForeignPtr fp $ \skp -> do
+      let skp' :: Ptr CUChar = castPtr skp
+      allocaBytes 96 $ \kpp -> do
+        res <- secpKeyPairCreate ctxPtr kpp skp'
+        when (res /= 1) (error "deriveVerKeyDSIGN: Failed to create keypair")
```
michaelpj, line 119: The error handling around here seems a bit strange, but I note that some of the other modules at least throw IO exceptions: https://github.com/input-output-hk/cardano-base/blob/39f9cf0eaf15a6888a47ac9d8de0b37dd1131d0b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Ed25519.hs#L53

Perhaps that function could be moved to `Cardano.Crypto.Util` and used here too.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -0,0 +1,296 @@
...
+  type SizeSignKeyDSIGN SchnorrSecp256k1DSIGN = 32
+  type SizeVerKeyDSIGN SchnorrSecp256k1DSIGN = 64
+  type Signable SchnorrSecp256k1DSIGN = SignableRepresentation
+  newtype VerKeyDSIGN SchnorrSecp256k1DSIGN = 
+    VerKeySchnorr256k1 (ForeignPtr SECP256k1XOnlyPubKey)
+      deriving NoThunks via (
+        OnlyCheckWhnfNamed "VerKeySchnorr256k1" (ForeignPtr SECP256k1XOnlyPubKey)
+        )
+  newtype SignKeyDSIGN SchnorrSecp256k1DSIGN = 
+    SignKeySchnorr256k1 (ForeignPtr SECP256k1SecKey)
```
michaelpj, line 103: Why not use `PinnedSizedBytes` for these?

michaelpj:
All of my recent comments were just caused by taking a look at some of the other code in this package, and suggesting ways of making it more consistent and using the utilities that the package provides. It would be great if you could do such a pass yourself, you'll probably find more stuff that I haven't spotted!

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
kozross: I saw that, but I actually don't think it helps all that much. I prefer keeping close to C, magic numbers and all, and avoiding the cost of 'calling down' type-level numbers to the value level.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
kozross: I'm not sure whether `error` or an IO exception is more appropriate here. We're using `unsafe*PerformIO` everywhere anyway, so I don't think it makes that much difference.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
kozross: For the same reason as mentioned previously. My experience with FFI tells me that it's much better to be closer to C, and 'calling down' type level numbers to the value level is _not_ a cheap operation.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
michaelpj: Nonetheless, if it's going to be in this repository, I think it should follow the practices of this repository :shrug:

kozross: I respectfully disagree. You're injecting quite expensive code into (arguably) a hot code path for no real benefit, since nobody outside of the maintainers of this repo will see it. If magic numbers are a concern, I am happy to replace those with named constants.

michaelpj: > You're injecting quite expensive code into (arguably) a hot code path for no real benefit

I think both of these are wrong:
- The code is not expensive. GHC happily reifies the statically known information at compile time: I looked at the Core for the `Ed25519` module and it has numeric literals as the size arguments.
- This is not a hot path: we are calling expensive primitives, we shouldn't need to worry too much about the FFI overhead.

In any case, please remove the magic numbers. You already had to define named constants in the class: use them. I also don't think there's a good reason not to use the library helpers here, and it makes things more consistent.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
michaelpj: Again, this is false: look at the Core if you don't believe me. These are statically known, so there is no overhead.

kozross:
@michaelpj I have now rewritten the Schnorr signature work to use the wrappers in `cardano-base`, as well as `useAsCStringLen`. I believe this should resolve all outstanding concerns.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -160,107 +165,19 @@ instance DSIGNAlgorithm SchnorrSecp256k1DSIGN where
...
-  {-# NOINLINE show #-}
-  show (SigSchnorr256k1 fp) = 
-    ("SigSchnorr256k1 " <>) . 
-    show . 
-    unsafeDupablePerformIO . 
-    withForeignPtr fp $ \p -> 
-      unsafePackCStringLen (castPtr p, 64)
+      psb <- withForeignPtr (plusForeignPtr bsFP bsOff) $ \bsp -> do
+        psbCreate $ \skp -> do
+          copyPtr skp bsp 32
```
michaelpj: why did this change? _Please_ replace the magic numbers with constants.

michaelpj:
I think the only outstanding comment is the one about magic numbers. I would still prefer you used `SizedPtr`, but I can do that afterwards so long as it's actually clear what the numbers mean.

@iquerejeta are you happy with the naming now?

iquerejeta:
There are two unresolved comments above regarding naming. I agree with @kozross that the functions will not be exposed so we can be more flexible with the naming, but I still believe we need to be consistent and correct, for ease of maintenance of the code. The two particular cases are with the qualified import of ECDSA
```
import qualified Crypto.Secp256k1 as SECP
```
where we cannot see that it actually relates to ECDSA. I would prefer something along the lines of `import qualified Crypto.EcdsaSecp256k1`, but I would be ok with the second suggestion. 

And the second point is that `VerKeySchnorr256k1` should be changed to `VerKeySchnorrSecp256k1`.

kozross:
@michaelpj How do you get a `SizedPtr` to the contents of an arbitrary `ByteString`?

@iquerejeta What you are asking for makes very little sense. `Crypto.Secp256k1` refers _specifically_ to [this module here](https://hackage.haskell.org/package/secp256k1-haskell-0.6.0/docs/Crypto-Secp256k1.html). I can import it as `ECDSA`, but to be honest, if the naming confusion is _anywhere_, it's upstream. I can rename `VerkeySchnorr256k1` as you suggest, but then, for consistency, I would need to rename all the other constructors too.

michaelpj:
Okay, LGTM. Please merge master and I'll merge this.

michaelpj:
Normally I feel like you get the option to merge master into other people's branches, but not here :thinking:

iquerejeta:
> I can import it as ECDSA, but to be honest, if the naming confusion is anywhere, it's upstream.

Sounds good. I think that even if the naming confusion is upstream, we should make it as clear as we can in our code base. So that when future developers come to change the code, they know that they are handling the ECDSA algorithm, and not something different. My impression is that the upstream library was initially build only for ECDSA, and afterwards schnorr was added. That might justify some of the confusing naming. However, we are in a better position. We are including both ECDSA and Schnorr at the same time, so I don't see the need of making the same mistakes as upstream. 

> I can rename VerkeySchnorr256k1 as you suggest, but then, for consistency, I would need to rename all the other constructors too.

Sounds reasonable to me. I think it is important to be consistent.

kozross:
@iquerejeta `secp256k1-haskell` doesn't support Schnorr signatures at all, which is why I had to FFI into `secp256k1` directly.

michaelpj (approved):

tdammers (commented):
Looks good to me.

PR #260, opened 2022-02-25
Add `NFData` instance for `StrictSeq`

lehins:
Also moves some orphan instances from ledger: 
* `NFData` for: `WithOrigin`, `EpochNo`, `EpochSize`
* `ToCBOR` and `FromCBOR` for `EpochSize`
* `ToJSON`, `FromJSON` for `StrictSeq`


lehins:
Looks like there is a missing dependency which is causing CI to fail. Definitely unrelated to this PR

nc6 (approved):

PR #263, opened 2022-03-03
Install libsecp256k1 in CI

michaelpj:


michaelpj:
Oh god, Windows. I have no idea what to do about that. Thoughts @newhoggy ?

michaelpj:
And it looks like the ubuntu we're building on doesn't have new enough `libsecp256k1` :facepalm:

PR #264, opened 2022-03-07
Bump iohk-nix and drop nixpkgs pin

michaelpj:
It looks like the nixpkgs pin from `haskell.nix` *is* actually new
enough.

I didn't get cache hits until I switched to 8.10.7, which seems like a
good idea anyway.

### DIFF
--- a/nix/haskell.nix
+++ b/nix/haskell.nix
@@ -7,7 +7,7 @@
 , buildPackages
 , config ? {}
 # GHC attribute name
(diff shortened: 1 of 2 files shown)


michaelpj:
FYI @nc6 , I think this may also mean that you only need to bump `haskell.nix` in the ledger too.

michaelpj:
This is definitely wrong, I just checked it badly locally.

PR #266, opened 2022-03-10
BLS12-381 bindings

tdammers:
Implements #302


iquerejeta (commented):
That's great! 👍  

However, as you'll see in the comments below, we should change a few things. The main points are the following: 
* We should remove the pairing check (and by consequence the `two_miller_one_exp`)
* We should change the definition of "equal elements" for GT elements
* We should check points are in the group before computing the miller loop

In a nutshell we will be doing the optimization we did in `pairing_check` for all operations between GT elements. i.e. we are generalizing it so that, instead of only applying the optimization when checking equality of two pairing evaluations, we can apply the optimization to an arbitrary number of pairing evaluations. Therefore, we only compute the miller loop when "computing the pairing", and when we check for equality, we do the `finalcheck`, which handles the final exponentiation. 

To this end we should include the following test with three pairings to the library: 
```haskell
        (BLS.pairing (BLS.mult p a, q) * BLS.pairing (p, BLS.mult q b))
        ==
        BLS.pairing(BLS.mult p (a + b), q)
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,896 @@
...
+  (==) = ptEq
+
+---- Pairings
+
+pairing :: P1 -> P2 -> PT
+pairing p1 p2 = unsafePerformIO $
+  withP p1 $ \pp1 ->
+    withP p2 $ \pp2 ->
+      withNewPT' $ \ppt ->
+        c_blst_miller_loop ppt pp1 pp2
```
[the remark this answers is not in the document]
tdammers: No idea, I just followed the C code here, which takes `p1*` / `p2*`, not affines.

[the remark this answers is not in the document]
tdammers: Oh wait, it actually does take affines. Never mind, I'll change.

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
```diff
@@ -0,0 +1,224 @@
...
+testAssoc f a b c =
+  f a (f b c) === f (f a b) c
+
+testCommut :: (Show a, Eq a) => (a -> a -> a) -> a -> a -> Property
+testCommut f a b =
+  f a b === f b a
+
+testNeutral :: (Show a, Eq a) => a -> (a -> a -> a) -> a -> Property
+testNeutral z f x =
+  f x z === x
```
jmhrpr, line 178: This property does not seem to be included in any test tree

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
```diff
@@ -0,0 +1,224 @@
...
+    , testProperty "to/from Integer round-trip" $
+        \s -> s === ((unsafePerformIO . BLS.scalarFromInteger) . (unsafePerformIO . BLS.scalarToInteger) $ s)
+    , testCase "integer from scalar" $ do
+        s <- case BLS.scalarFromBS (BS.pack [0x12, 0x34]) of
+              Left err -> error (show err)
+              Right x -> return x
+        let expected = 0x1234
+        actual <- BLS.scalarToInteger s
+        assertEqual "0x1234" expected actual
+    ]
```
jmhrpr, line 82: I think we need to have a similar test tree for the Natural conversion functions

- Natural / C-String round-trip (`natAsCStr` / `cstrToNat`)
    - There's no `natFromBS` function, but the Natural / C-String round-trip uses `natToBS` so I guess that `natToBS` is tested here.
- Natural / C-String 32 round-trip (`natAsCStrL` / `cstrToNat`)
- Scalar / Natural round-trip (`scalarToNat` / `scalarFromNat`) (note `scalarFromNat` truncates 32 bytes)

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/C.hs
```diff
@@ -0,0 +1,618 @@
...
+  b <- mallocForeignPtrBytes sizeFr
+  withForeignPtr a $ \ap ->
+    withForeignPtr b $ \bp ->
+      copyBytes bp ap sizeFr
+  return (Fr b)
+
+---- Safe FP types / marshalling
+
+sizeFP :: forall i. Num i => i
+sizeFP = fromIntegral c_size_blst_scalar
```
jmhrpr, line 422: `c_size_blst_scalar` is 32, but the size of FP is 48. We need a `c_size_blst_fp` in `cbits/blst.h` which is then used here.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/C.hs
```diff
@@ -0,0 +1,618 @@
...
+  withForeignPtr a $ \ap ->
+    withForeignPtr b $ \bp ->
+      copyBytes bp ap sizeFP
+  return (FP b)
+
+fpToNat :: FP -> IO Natural
+fpToNat fp = withFP fp $ \fpPtr -> do
+  allocaBytes 48 $ \rawPtr -> do
+    c_blst_bendian_from_fp rawPtr fpPtr
+    cstrToNat rawPtr 48
```
jmhrpr, line 454: The hardcoded lengths of 48 could be replaced with `sizeFP` for consistency

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/C.hs
```diff
@@ -0,0 +1,618 @@
...
+
+natAsCStrL :: Int -> Natural -> (Ptr CChar -> Int -> IO a) -> IO a
+natAsCStrL i n f = do
+  let bs = padBS i $ natToBS n
+  BS.useAsCStringLen bs $ uncurry f
+
+scalarFromNat :: Natural -> IO Scalar
+scalarFromNat n = do
+  withNewScalar' $ \scalarPtr -> do
+    natAsCStrL 32 n $ \str _length -> do
```
jmhrpr, line 500: The hardcoded length of 32 can be replaced by `sizeScalar`

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+scalarFromBS bs =
+  if success then
+    Right scalar
+  else
+    Left BLST_BAD_SCALAR
+  where
+    (success, scalar) = unsafePerformIO $
+      withNewScalar $ \scalarPtr ->
+        BS.useAsCStringLen bs $ \(cstr, l) ->
+          c_blst_scalar_from_be_bytes scalarPtr cstr (fromIntegral l)
```
jmhrpr, line 823: I was looking at the Go bindings and they check that the length of the byte string is not less than `sizeScalar` before passing to the conversion function. That being said, `blst_scalar_from_be_bytes` does seem to be able to handle longer byte strings, so maybe it's ok - I think it takes the last 32 bytes of the provided bytestring. Again perhaps a question of whether we want to protect user from having their input silently truncated.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+padBS i b
+  | i > BS.length b
+  = BS.replicate (i - BS.length b) 0 <> b
+  | otherwise
+  = b
+
+integerAsCStr :: Integer -> (Ptr CChar -> Int -> IO a) -> IO a
+integerAsCStr n f = do
+  let bs = integerToBS n
+  BS.useAsCStringLen bs $ uncurry f
```
jmhrpr, line 508: This doesn't seem to be used anywhere (other than the tests, so maybe its just to check `cstrToInteger`), instead the padded version `integerAsCStrL` is used.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/C.hs
```diff
@@ -0,0 +1,618 @@
...
+cstrToNat :: Ptr CChar -> Int -> IO Natural
+cstrToNat p l = do
+  go l (castPtr p)
+  where
+    go :: Int -> Ptr CUChar -> IO Natural
+    go 0 _ = return 0
+    go n ptr = do
+      val <- peek ptr
+      res <- go (pred n) (plusPtr ptr 1)
+      return $ shiftL res 8 .|. fromIntegral val
```
jmhrpr, line 471: This is passed a big-endian bytestring in `fpToNat` but the bit-shifting logic for crafting the Natural seems to treat it as a little-endian

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+foreign import ccall "size_blst_affine1" c_size_blst_affine1 :: CSize
+foreign import ccall "size_blst_affine2" c_size_blst_affine2 :: CSize
+
+foreign import ccall "blst_p1_to_affine" c_blst_p1_to_affine :: AffinePtr Curve1 -> PPtr Curve1 -> IO ()
+foreign import ccall "blst_p2_to_affine" c_blst_p2_to_affine :: AffinePtr Curve2 -> PPtr Curve2 -> IO ()
+foreign import ccall "blst_p1_from_affine" c_blst_p1_from_affine :: PPtr Curve1 -> AffinePtr Curve1 -> IO ()
+foreign import ccall "blst_p2_from_affine" c_blst_p2_from_affine :: PPtr Curve2 -> AffinePtr Curve2 -> IO ()
+
+foreign import ccall "blst_p1_affine_in_g1" c_blst_p1_affine_in_g1 :: AffinePtr Curve1 -> IO Bool
+foreign import ccall "blst_p2_affine_in_g2" c_blst_p2_affine_in_g2 :: AffinePtr Curve2 -> IO Bool
```
jmhrpr, line 593: We can use the type synonyms here for consistency

```
  type Affine1Ptr = AffinePtr Curve1
  type Affine2Ptr = AffinePtr Curve2
  type P1Ptr = PPtr Curve1
  type P2Ptr = PPtr Curve2
  ```

File: cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs (hunk 2 above)
Comment (tdammers) [reply cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs:178]: Good catch. The purpose of this function is to test that a given value `z` is a neutral element (the `mzero` of a `Monoid`) over a given binary operation `f`. This may be a remnant of an earlier iteration, where we did test this property for some operations. It could possibly also be used to replace the existing "mult by p+1 is identity" test as `testNeutral (p+1) BLS.mult`.

File: cardano-crypto-class/cbits/blst_util.c [hunk 11]
```diff
@@ -0,0 +1,48 @@
...
+void blst_y_from_p1(blst_fp *dst, const blst_p1 *p) { memcpy(dst, &p->y, sizeof(blst_fp)); }
+void blst_z_from_p1(blst_fp *dst, const blst_p1 *p) { memcpy(dst, &p->z, sizeof(blst_fp)); }
+void blst_x_from_p2(blst_fp *dst, const blst_p2 *p) { memcpy(dst, &p->x, sizeof(blst_fp)); }
+void blst_y_from_p2(blst_fp *dst, const blst_p2 *p) { memcpy(dst, &p->y, sizeof(blst_fp)); }
+void blst_z_from_p2(blst_fp *dst, const blst_p2 *p) { memcpy(dst, &p->z, sizeof(blst_fp)); }
+
+void blst_x_from_affine1(blst_fp *dst, const blst_p1_affine *affine) { memcpy(dst, &affine->x, sizeof(blst_fp)); }
+void blst_y_from_affine1(blst_fp *dst, const blst_p1_affine *affine) { memcpy(dst, &affine->y, sizeof(blst_fp)); }
+void blst_x_from_affine2(blst_fp *dst, const blst_p2_affine *affine) { memcpy(dst, &affine->x, sizeof(blst_fp)); }
+void blst_y_from_affine2(blst_fp *dst, const blst_p2_affine *affine) { memcpy(dst, &affine->y, sizeof(blst_fp)); }
```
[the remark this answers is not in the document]
jmhrpr: If not, these also need to be removed from C.hs

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+foreign import ccall "blst_scalar_from_be_bytes" c_blst_scalar_from_be_bytes :: ScalarPtr -> Ptr CChar -> CSize -> IO Bool
+foreign import ccall "blst_scalar_from_bendian" c_blst_scalar_from_bendian :: ScalarPtr -> Ptr CChar -> IO ()
+
+---- Raw P1 functions
+
+foreign import ccall "size_blst_p1" c_size_blst_p1 :: CSize
+foreign import ccall "blst_p1_on_curve" c_blst_p1_on_curve :: P1Ptr -> IO Bool
+
+foreign import ccall "blst_p1_add_or_double" c_blst_p1_add :: P1Ptr -> P1Ptr -> P1Ptr -> IO ()
+foreign import ccall "blst_p1_mult" c_blst_p1_mult :: P1Ptr -> P1Ptr -> ScalarPtr -> CInt -> IO ()
```
jmhrpr, line 544: The C function signature has the 4th argument as a `size_t`, so I would of expected CSize instead of CInt but perhaps this was intentional? Hackage says that CSize uses Word64 and CInt uses Int32. There were one or two more cases like this, so I can find them again if this sounds wrong.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+  c_blst_affine_in_g :: AffinePtr curve -> IO Bool
+  c_blst_generator :: PPtr curve
+
+  c_blst_p_is_equal :: PPtr curve -> PPtr curve -> IO Bool
+  c_blst_p_is_inf :: PPtr curve -> IO Bool
+
+instance BLS_Curve Curve1 where
+  c_blst_on_curve = c_blst_p1_on_curve
+
+  c_blst_add = c_blst_p1_add
```
jmhrpr, line 363: I think because we eventually use this function to FFI call the `add_or_double` function from the C library it makes sense to change this to.
`c_blst_add = c_blst_p1_add_or_double` or `c_blst_add_or_double = c_blst_p1_add_or_double`

And then change all instances of `c_blst_add` to `c_blst_add_or_double`

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+
+  c_blst_generator = c_blst_p1_generator
+
+  c_blst_p_is_equal = c_blst_p1_is_equal
+  c_blst_p_is_inf = c_blst_p1_is_inf
+
+instance BLS_Curve Curve2 where
+  c_blst_on_curve = c_blst_p2_on_curve
+
+  c_blst_add = c_blst_p2_add
```
jmhrpr, line 386: Make clear that this uses `add_or_double` similar to with Curve1

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+foreign import ccall "blst_fr_from_scalar" c_blst_fr_from_scalar :: FrPtr -> ScalarPtr -> IO ()
+foreign import ccall "blst_scalar_from_be_bytes" c_blst_scalar_from_be_bytes :: ScalarPtr -> Ptr CChar -> CSize -> IO Bool
+foreign import ccall "blst_scalar_from_bendian" c_blst_scalar_from_bendian :: ScalarPtr -> Ptr CChar -> IO ()
+
+---- Raw P1 functions
+
+foreign import ccall "size_blst_p1" c_size_blst_p1 :: CSize
+foreign import ccall "blst_p1_on_curve" c_blst_p1_on_curve :: P1Ptr -> IO Bool
+
+foreign import ccall "blst_p1_add_or_double" c_blst_p1_add :: P1Ptr -> P1Ptr -> P1Ptr -> IO ()
```
jmhrpr, line 543: Also rename `c_blst_p1_add` to `c_blst_p1_add_or_double`

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+
+foreign import ccall "blst_p1_is_equal" c_blst_p1_is_equal :: P1Ptr -> P1Ptr -> IO Bool
+foreign import ccall "blst_p1_is_inf" c_blst_p1_is_inf :: P1Ptr -> IO Bool
+
+---- Raw P2 functions
+
+foreign import ccall "size_blst_p2" c_size_blst_p2 :: CSize
+foreign import ccall "blst_p2_on_curve" c_blst_p2_on_curve :: P2Ptr -> IO Bool
+
+foreign import ccall "blst_p2_add_or_double" c_blst_p2_add :: P2Ptr -> P2Ptr -> P2Ptr -> IO ()
```
jmhrpr, line 565: Change `c_blst_p2_add` to `c_blst_p2_add_or_double`

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,883 @@
...
+-- | Check whether a point is on its elliptic curve.
+onCurve :: BLS_Curve curve => P curve -> Bool
+onCurve p = unsafePerformIO $ withP p c_blst_on_curve
+
+-- | Check whether a point is in the group corresponding to its elliptic curve
+inGroup :: BLS_Curve curve => P curve -> Bool
+inGroup p = unsafePerformIO $ withP p c_blst_in_g
+
+-- | Curve point addition.
+add :: (BLS curve) => P curve -> P curve -> P curve
```
jmhrpr, line 686: Make clear that this uses `add_or_double` not `add`

jmhrpr (commented):
Comments on function naming for `add` vs `add_or_double`

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
```diff
@@ -0,0 +1,200 @@
...
+import Test.Crypto.Instances ()
+import Test.QuickCheck (
+--   (=/=), 
+    (===), 
+    (==>), 
+    Arbitrary(..), 
+    Property,
+    suchThatMap,
+--   Gen, 
+--   forAllShow
```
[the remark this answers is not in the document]
tdammers: Yep.

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
tdammers: Removed.

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
tdammers: Actually, the entire C.hs module, and with it the conversions to and from `Nat`, are obsolete; it was already excluded from the .cabal file, but I have now also deleted it from the repo. In practice, we will only need the conversions to and from `Integer`.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/C.hs
tdammers: Module has been removed, we will not be needing this.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
tdammers: Removed.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
tdammers: Done.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
tdammers: No, this was most definitely not intentional. `size_t` should pretty much always be marshalled to `CSize`, that's what it's for.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
tdammers: Done.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
tdammers: I have now changed the name to `addOrDouble`. Better to make it explicit than to rely on documentation.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
tdammers: Done.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
tdammers: Done.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
tdammers: Done.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/C.hs
tdammers: Module removed.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
jmhrpr: This is still using `CInt`, similarly on L340 and L559

jmhrpr (changes requested):

arthurgreef:
Hi - I get the following error when I execute nix-shell on this branch. Any ideas on how to fix my problem? Thanks.
```
trace: Not using IOHK default nixpkgs (use 'niv drop nixpkgs' to use default for better sharing)
building '/nix/store/qxgj384y8xwjhl082rg7ash8sx8d7dl9-git-ls-files.drv'...
error: The Nixpkgs package set does not contain the package: blst (system dependency).
       You may need to augment the system package mapping in haskell.nix so that it can be found.```

**Comment by arthurgreef** [comment]
Hi @tdammers, can we get an update on the PR? Also, is there a branch that exposes these bindings to dapps via Plutus?

**Comment by iquerejeta** [comment]
Hi @arthurgreef . This PR is still in draft, as we are waiting for benchmarks to determine whether these primitives will be usable in practice (due to the execution budget). The use-case we are currently considering is that of ATMS signatures. What are you  planning on using them for? 
And regarding your question, there is no branch that exposes these bindings via Plutus yet, and that most probably won't happen until we have not concluded that they can be used in practice.

btw, did you manage to solve the problem with nix you had?

**Comment by arthurgreef** [comment]
Hi @iquerejeta my use case is DKG.
Here is the paper. https://eprint.iacr.org/2019/985.pdf
Here is the ETH smart contract I wanted to convert to Plutus. https://github.com/PhilippSchindler/EthDKG/blob/master/contracts/ETHDKG.sol

**Comment by iquerejeta** [comment]
Thanks @arthurgreef . I will try to look into the nix issue soon (we've been building it directly with `cabal` so far). And regarding DKG, that's cool. And the intention of using pairing friendly curve is because you want to generate a key for a pairing friendly primitive? or simply because BLS curve is what is closer to be available in Plutus?

**Comment by arthurgreef** [comment]
@iquerejeta we need BLS12-381 so that organizations can issue verifiable credentials using signatures that comply with the BBS+ selective disclosure protocol. https://w3c-ccg.github.io/ldp-bbs2020/

**Comment by arthurgreef** [comment]
Hi @iquerejeta - we are need BLS12-381 for organizations that sign verifiable credentials compliant with BBS+ https://w3c-ccg.github.io/ldp-bbs2020/

**Comment by jhofp** [comment]
Hi @tdammers , thanks for all the great works from you and your team on bring BLS12-381 to Plutus. 
Is there any timeline that i could expect for this to be on Plutus? thanks!

File: cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs (hunk 12 above)
Comment (iquerejeta) [reply cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs]: thanks for pointing this out @jmhrpr 🙏  I've updated the code. I did un-resolve the comment. I believe, however, that it is now resolved. If you agree, feel free to resolve.

**Comment by iquerejeta** [comment]
@arthurgreef you should be able to make it work with nix. Let me know if that is not the case.

File: cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs (hunk 7 above)
Comment (iquerejeta) [reply cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs]: I changed the function so that now it checks that the length is the expected length. @jmhrpr , if you agree with the changes ([here](https://github.com/input-output-hk/cardano-base/pull/266/files#diff-c03c192225bf84bd0c9cf7d189bc02bf3e266f01e37a9f16c0d0ba0f8e8644aeR802-R810)), could you resolve this conversation?

**Comment by iquerejeta** [comment]
@arthurgreef , nix should work now. If that's not the case, please let me know.

**Comment by arthurgreef** [comment]
`nix-shell` works for me now. thank you.

File: cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs (hunk 7 above)
Comment (jmhrpr) [reply cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs]: Code from Go bindings:

```
func (fr *Scalar) FromBEndian(arr []byte) *Scalar {
	nbytes := len(arr)
	if nbytes < BLST_SCALAR_BYTES ||
		!C.blst_scalar_from_be_bytes(fr, (*C.byte)(&arr[0]), C.size_t(nbytes)) {
		return nil
	}
	return fr
}
```

New code:
```
scalarFromBS bs = unsafePerformIO $ do
  BS.useAsCStringLen bs $ \(cstr, l) -> do
    (success, scalar) <- withNewScalar $ \scalarPtr -> do
      c_blst_scalar_from_be_bytes scalarPtr cstr (fromIntegral l)
    if (success && l == sizeScalar) then
      return $ Right scalar
    else
      return $ Left BLST_BAD_SCALAR
```

The difference is that in the Go bindings the conditional will short-circuit if the bytestring length is too small and the C function won't be called on the short bytestring, whereas in your code we will call the C function with a bytestring which is maybe too small (or at least, the Go bindings are explicitly preventing this from happening so I am just assuming it is "too small") and then check the length after. Maybe the C function can handle it OK, but perhaps it's just better to align with their bindings instead of figuring out what the C function is doing. I can take a closer look at the C function if you want.

File: cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs (hunk 12 above)
Comment (jmhrpr) [reply cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs]: Resolved (I can't mark as resolved)

File: cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs (hunk 7 above)
Comment (iquerejeta) [reply cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs]: yep, you are absolutely right. I've changed it 👍

**Comment by angerman** [comment]
I doubt we actually installed blst at all, the process took 1s, that seems unreasonable.

File: cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs [hunk 19]
```diff
@@ -0,0 +1,875 @@
...
+      c_blst_compress cstrp pp
+      BS.packCStringLen (cstrp, compressedSizeP (Proxy @curve))
+
+serialize :: forall curve. (BLS_P curve, BLS_Curve curve) => P curve -> ByteString
+serialize p = unsafePerformIO $ do
+  withP p $ \pp -> do
+    cstr <- mallocForeignPtrBytes (serializedSizeP (Proxy @curve))
+    withForeignPtr cstr $ \cstrp -> do
+      c_blst_serialize cstrp pp
+      BS.packCStringLen (cstrp, serializedSizeP (Proxy @curve))
```
lehins, line 748: Same here
```suggestion
          BS.fromForeignPtr0 cstr (serializedSizeP (Proxy @curve))
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,875 @@
...
+
+-- | @hash msg mDST mAug@ generates the elliptic curve hash for the given
+-- message @msg@; @mDST@ and @mAug@ are the optional @aug@ and @dst@
+-- arguments.
+hash :: (BLS_P curve, BLS_Curve curve) => ByteString -> Maybe ByteString -> Maybe ByteString -> P curve
+hash msg mDST mAug = unsafePerformIO $
+  BS.useAsCStringLen msg $ \(msgPtr, msgLen) ->
```
lehins, line 755:
```suggestion
  BS.unsafeUseAsCStringLen msg $ \(msgPtr, msgLen) ->
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,875 @@
...
+          withNewPT' $ \ppt ->
+            c_blst_miller_loop ppt ap2 ap1
+  else
+    Left BLST_POINT_NOT_IN_GROUP
+
+---- Utility
+
+withMaybeCStringLen :: Maybe ByteString -> (CStringLen -> IO a) -> IO a
+withMaybeCStringLen Nothing go = go (nullPtr, 0)
+withMaybeCStringLen (Just bs) go = BS.useAsCStringLen bs go
```
lehins, line 871: It would be useful to add some documentation stating that the supplied action must not mutate the pointer. This is already true because of `nullPtr` usage
```suggestion
withMaybeCStringLen (Just bs) go = BS.unsafeUseAsCStringLen bs go
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,875 @@
...
+frFromCanonicalScalar :: Scalar -> IO (Maybe Fr)
+frFromCanonicalScalar scalar
+  | scalarCanonical scalar
+  = Just <$> frFromScalar scalar
+  | otherwise
+  = return Nothing
+
+scalarFromBS :: ByteString -> Either BLSTError Scalar
+scalarFromBS bs = unsafePerformIO $ do
+  BS.useAsCStringLen bs $ \(cstr, l) ->
```
lehins, line 807:
```suggestion
  BS.unsafeUseAsCStringLen bs $ \(cstr, l) ->
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,875 @@
...
+    else
+      return $ Left BLST_BAD_SCALAR
+
+scalarToBS :: Scalar -> ByteString
+scalarToBS scalar = unsafePerformIO $ do
+  cstr <- mallocForeignPtrBytes sizeScalar
+  withForeignPtr cstr $ \cstrp -> do
+    withScalar scalar $ \scalarPtr -> do
+      c_blst_bendian_from_scalar cstrp scalarPtr
+      BS.packCStringLen (castPtr cstrp, sizeScalar)
```
lehins, line 824:
```suggestion
  BS.fromForeignPtr0 (castForeignPtr cstr) sizeScalar
```

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
```diff
@@ -0,0 +1,214 @@
...
+testScalar name =
+  testGroup name
+    [ testProperty "self-equality" $
+        \(a :: BLS.Scalar) -> a === a
+    , testProperty "to/from BS round-trip" $
+        \s -> Right s === (BLS.scalarFromBS . BLS.scalarToBS $ s)
+    , testProperty "non-negative" $
+        \s -> (unsafePerformIO . BLS.scalarToInteger $ s) >= 0
+    , testProperty "to/from Integer round-trip" $
+        \s -> s === ((unsafePerformIO . BLS.scalarFromInteger) . (unsafePerformIO . BLS.scalarToInteger) $ s)
```
lehins, line 67:
```suggestion
        \s -> s === unsafePerformIO (BLS.scalarToInteger s >>= BLS.scalarFromInteger)
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -0,0 +1,875 @@
...
+
+integerAsCStrL :: Int -> Integer -> (Ptr CChar -> Int -> IO a) -> IO a
+integerAsCStrL i n f = do
+  let bs = padBS i $ integerToBS n
+  BS.useAsCStringLen bs $ uncurry f
+
+scalarFromInteger :: Integer -> IO Scalar
+scalarFromInteger n = do
+  withNewScalar' $ \scalarPtr -> do
+    integerAsCStrL sizeScalar (n `mod` scalarPeriod) $ \str _length -> do
```
lehins, line 514: We either need to check that `n` is not negative or use `rem` here instead:
```suggestion
    integerAsCStrL sizeScalar (n `rem` scalarPeriod) $ \str _length -> do
```

This has to do with the way modulo arithmetics work on the negative numbers. Here is an example:
```haskell
> scalarPeriod = 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001 :: Integer
> (1 - scalarPeriod) `mod` scalarPeriod
1
> (1 - scalarPeriod) `rem` scalarPeriod
-52435875175126190479447740508185965837690552500527637822603658699938581184512
```

nix/sources.json
```diff
@@ -38,15 +38,15 @@
         "version": "962ecfed3a4fb656b5a91d89159291e00ed766bc"
     },
     "iohk-nix": {
-        "branch": "master",
+        "branch": "iquerejeta/include-blst",
```
lehins, line 41: Is the `blst` not in `master` yet?

.github/workflows/haskell.yml
```diff
@@ -96,6 +98,23 @@ jobs:
...
+
+    - name: "NON-WINDOWS: Install blst"
+      if: runner.os != 'Windows'
+      run: |
+        mkdir blst-sources && cd blst-sources
+        git clone https://github.com/supranational/blst
+        cd blst
+        git reset --hard $BLST_REF
+        ./build.sh
+        cd ../..
```
lehins, line 116: This looks almost exactly as the `.github/workflows/build-blst.bash` script. Can we reuse it here?

lehins (changes requested):
Here is my first pass at it. I'd like to do a second more thorough review once current comments are addressed.  It's a big PR and it's hard to review the whole thing in just one swoop.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
iquerejeta: what's wrong with 1 - scalarPeriod `mod` scalarPeriod being equal to 1? That's what we'd actually expect, right?

lehins: So, you are saying negative values are allowed as input? If that's the case then it's fine, I just would not have expect it to be the case. This sort of overflows is the constant source of bugs in the bounded integer arithmetic, that's why I always try to be extra careful with this. I am not familiar with BLS12 at all, so it might be totally expected by the users of the library.

Maybe it would be good a idea to add a note about negative numbers on the `mult` function? It is sort of implied by the haddock: "The scalar will be brought into the range of modular arithmetic by means of a modulo operation over the 'scalarPeriod'", but being more explicit about negative numbers could be useful IMHO.

iquerejeta: Sometimes, when interacting with elliptic curve points we want to check things such as `a * A - b * B` (where capital letters represent points). A way to do this would be `a * A + NEG(b * B)`, but if we allow negative numbers, we allow the user to call `a * A + (-b) * B`. 

However, Now that I'm thinking about it, the `NEG` operation on an ecpoint might be cheaper (it's a sign bit flip only). I'll think about this. We have to change the representation of points nonetheless, because @kwxm found a bug that the following:  `neg(A) =? (-1)*A`, sometimes does not hold. I'll come back to this.

kwxm: Hi. I'm extending Plutus Core to include some BLS operations based on this branch, and I'm pretty confident that `mod` is the right thing to use here. Scalar multiplication can be interpreted as repeated addition:
```
0p = 0
np = p + p + ... +p    (n times if n>0)
np = (-p) + (-p) + ... + (-p)   (-n times if n<0)
```
With these definitions you have sensible laws like n(p+q) = np + nq and (mn)p = m(np), and in fact earlier this week I added [tests](https://github.com/input-output-hk/plutus/blob/efd3d1ff5859abbd0abfb3e39bd702487b85532d/plutus-core/untyped-plutus-core/test/Evaluation/Builtins/BLS12_381/HaskellTests.hs#L81) to check such properties: they all pass, for all integral values (although there was an unrelated problem that we have to fix). I'm also paranoid about `rem` and `mod` (the fact that we have both is very confusing and I have to check the documentation every time I use one of those); I changed `mod` to `rem` in the code you quoted above and then the tests started failing for negative numbers ("point not on curve").  I think that this is because all of the functions that `blst` uses to construct scalars (like [this](https://github.com/supranational/blst/blob/d334254a4125cd2676d40b11229df208db8902e0/src/exports.c#L316) and the ones below) expect some unsigned representation of integers (although they don't actually say _what_ they expect anywhere), so if you supply a negative input the signed representation introduces unexpected bit patterns.

The crucial point here is that if we let N be the order of the group (ie, `scalarPeriod`), then `N*p = 0` for every point in G1 or G2, and it follows that (n+kN)*p = np for all integers k and n, positive or negative.  For example (-1)*p = (N-1)*p is an expensive way of calculating -p.  Now (as long as N>0),`a mod N` always gives you a result r with 0 <= r < N (which is thus a safe input to the `blst` scalar conversions), and `mod` also has the property that `(a +kN) mod n = a mod n`, so it's periodic modulo N in exactly the same way that repeated addition is (remember Np is always zero) and you get consistent results.

So I think it's perfectly safe to allow negative `Integer`s here. as long as everything's reduced modulo `scalarPeriod` before being fed to `blst` .  Maybe we don't _want_ to allow negative numbers, but I don't think it'd do any harm. You could argue that `blst` doesn't allow them, but then it also doesn't allow scalars bigger than `scalarPeriod`(although I think you can put any unsigned 256-bit value into a blst scalar, including some that are a bit bigger than the period: it'd be interesting to see what happens if you actually do that).

kwxm: > However, Now that I'm thinking about it, the `NEG` operation on an ecpoint might be cheaper (it's a sign bit flip only).

For G1, scalar multiplcation in Plutus is about 300 times more expensive than negation, and for G2 it's 600 times more expensive: you really don't want to be doing scalar multiplication if you can avoid it!

lehins: >the fact that we have both is very confusing

In Haskell `rem` is a primitive operation, while `mod` is implemented in terms of `rem`. For that reason `rem` is more efficient and thus better when dealing with non-negative numbers only (eg. array indices)

> it's perfectly safe to allow negative Integers here. as long as everything's reduced modulo scalarPeriod before being fed to blst

That's exactly what current implementation of `scalarFromInteger` does, by using `mod x scalarPeriod` negative  numbers are converted to positive, so as far as `c_blst_p2_mult` and `c_blst_p1_mult` are concerned they will be getting positive numbers

> the NEG operation on an ecpoint might be cheaper (it's a sign bit flip only)

Negation doesn't look to be just a bit flip. If I had to guess it also looks like modulo arithmetic.

1. https://github.com/supranational/blst/blob/d334254a4125cd2676d40b11229df208db8902e0/src/e2.c#L91-L95
2. https://github.com/supranational/blst/blob/d334254a4125cd2676d40b11229df208db8902e0/src/fields.h#L78
3. https://github.com/supranational/blst/blob/d334254a4125cd2676d40b11229df208db8902e0/src/asm/add_mod_384-armv8.pl#L468-L516

> Maybe we don't want to allow negative numbers, but I don't think it'd do any harm.

Also my guess would be that this `mul (neg x) p` or this `neg (mul x p)` could be more efficient then `mul x (negate p)`. This is the only thing I can think of, if those are even correct properties that is :smile: 

In any case, this are just my thoughts. I am not even familiar with BLS12-381, so you should decide on how it should behave. Feel free to resolve this comment at any time

iquerejeta: > Negation doesn't look to be just a bit flip. If I had to guess it also looks like modulo arithmetic.

Ah, you are right. The bit flip works with compressed points. In compressed form only the sign of the `y` coordinate is stored, so a negation is simply changing that byte. However, with the full `y` coordinate, we indeed need the modular reduction. 

 > scalar multiplcation in Plutus is about 300 times more expensive than negation
 
 But none of the tricks here save us the scalar multiplication. I think we are really deciding of whether negating a number modulo ~32 bytes (negating a scalar) is faster than negating one modulo ~48 bytes (negating the `y` coordinate). I guess the difference is very small, particularly in comparison with the other operations. I'd leave it as it is, and let the user decide what it wants to do.

iquerejeta: I'll resolve this.

nix/sources.json
lehins: Looks like PR has been merged: https://github.com/input-output-hk/iohk-nix/pull/528

We can go ahead and switch to `iohk-nix/master` now :1st_place_medal:

lehins (approved):
This looks great. :+1: 
I had a few suggestions, but none of them are blocking this PR from being merged.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381.hs
```diff
@@ -0,0 +1,48 @@
...
+  , mult
+  , cneg
+  , neg
+  , compress
+  , serialize
+  , uncompress
+  , deserialize
+  , hash
+  , generator
+  , isInf
```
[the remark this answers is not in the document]
iquerejeta: I don't know what is the Haskell convention here. To be honest, I'm no worried of enforcing qualified imports here. I would keep them short and require qualified imports. @tdammers, do you have a preference here?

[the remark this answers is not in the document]
lehins: We have a words that are used very often, even in this repo: `hash` and `(un)compress`, `(de)serialize`

It is better to disambiguate (even for the purpose of grepping the repo)

Even C functions are prefixed (obviously reasons why don't apply to Haskell), but I  think it would match up nicely.

I strongly suggest a `bls` prefix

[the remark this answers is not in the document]
iquerejeta: Ok, will do 👍

PR #269, opened 2022-03-30
Add a flag to turn off secp256k1 support

michaelpj:
Do so in the windows CI. This should hopefully unbreak people
downstream until we get a better solution.

(Draft until I see the Windows CI pass)

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -92,21 +92,20 @@ jobs:
         echo "LIBSODIUM_PATH=$LIBSODIUM_PATH"
         echo "$LIBSODIUM_PATH" >> $GITHUB_PATH
 
-    - name: Install secp256k1 (Windows)
-      if: matrix.os == 'windows-latest'
-      run: |
-        mkdir secp256k1-sources
-        cd secp256k1-sources
-        git clone https://github.com/bitcoin-core/secp256k1.git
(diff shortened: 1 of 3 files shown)


.github/workflows/haskell.yml
```diff
@@ -95,12 +95,10 @@ jobs:
...
       run: |
-        mkdir secp256k1-sources
-        cd secp256k1-sources
-        git clone https://github.com/bitcoin-core/secp256k1.git
-        cd secp256k1
-        git reset --hard $SECP256K1_REF
-        echo "FIXME: install from sources"
+        echo "FIXME: can't install from sources, so instead libsecp256k1 support is disabled"
+        cat >> cabal.project <<EOF
+        package cardano-crypto-class
```
lehins, line 100: I think it is better to do `cabal configure --flags="-secp256k1-support"'` instead of editing cabal.project

michaelpj: This is what is done further down for configuring libsodium, so I just copied that :shrug:

michaelpj: I also don't think that multiple calls to `cabal configure` will compose in the way we want.

michaelpj:
Oh, it won't pick up my changes to the workflow until this is merged, annoyingly, so we may just have to try it and see.

PR #271, opened 2022-03-31
Fix the SECP256K1 flag actually

michaelpj:
I missed some things, but more importantly I figured out how to test the
workflow without merging into master, so I now know that this does
actually work (on my fork).

### DIFF
--- a/cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
+++ b/cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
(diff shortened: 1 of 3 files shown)


cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
```diff
@@ -126,8 +134,10 @@ tests =
     [ testDSIGNAlgorithm mockSigGen (arbitrary @Message) "MockDSIGN"
     , testDSIGNAlgorithm ed25519SigGen (arbitrary @Message) "Ed25519DSIGN"
     , testDSIGNAlgorithm ed448SigGen (arbitrary @Message) "Ed448DSIGN"
+#ifdef SECP256K1
     , testDSIGNAlgorithm secp256k1SigGen genSECPMsg "EcdsaSecp256k1DSIGN"
     , testDSIGNAlgorithm schnorrSigGen (arbitrary @Message) "SchnorrSecp256k1DSIGN"
+#endif
```
lehins, line 140: @michaelpj Honestly, there is a little too much CPP for my taste, I'd put all secp256k1 related tests into a separate module and imported it with a single CPP `#if` statement. 

However, I am not gonna stand in a way of correct code, so it's up to you.

lehins (approved):
Ah, of course, there is still a test suite :smile:

cardano-crypto-tests/src/Test/Crypto/DSIGN.hs
michaelpj: I'm happy either way, I just wanted to go with the dumbest thing I could get working for now...

We'd still need at least two `#if`s: one for the import and one for the test invocation, but it would indeed be less.

If you don't mind, we can merge this to unblock people and I can make a follow-up with less CPP.

PR #272, opened 2022-04-05
Re-enable secp256k1 on Windows builds

newhoggy:
Change taken from https://github.com/input-output-hk/ouroboros-network/pull/3687


lehins (approved):
:+1: Thank you for getting it done!

michaelpj:
Thanks!

Is there an admin on the repo who could make these checks required so this doesn't happen again in future?

PR #273, opened 2022-04-11
Aeson compatiblity

newhoggy:


lehins:
@newhoggy What is the motivation behind compatibility with aeson 1.x? I am only asking because I think it will require quite a bit more work to make the rest of cardano ecosystsem compatible with both 2.x and 1.x aeson. Is there a plan to do all that work and is that work justified?

newhoggy:
Sorry.  It's just a draft at the moment.  I'm curious what break.

newhoggy:
Looks like `cardano-base` already builds with both version.  😁

PR #274, opened 2022-04-24
Packed bytes with offset

lehins:
It is useful for address deserialization to be able to create a hash from a [Short]ByteString with an offset, therefore avoiding unnecessary copy when creating a hash from bytes located in a larger ByteString.

Also switches to faster `decodeLatin1` for hex encoded hashes.


JaredCorduan (approved):
looks good to me!

nc6 (approved):

PR #276, opened 2022-05-10
Perform serialization of KES and DSIGN sign keys in MLocked memory

tdammers:
We will need this for a KES Agent: in order to send and receive keys over a secure network connection, we have to serialize and deserialize them. The serialized format is still just as confidential as the deserialized form, so we need to treat them with the same care, and that involves storing them in mlocked memory.

We also need to make sure no intermediate values ever reach unprotected memory, so we cannot use the usual Haskell serialization primitives, as these may store intermediates on the Haskell heap. At the same time, mlocked memory is a scarce resource, and we don't want to allocate excessive amounts of it, so in the case of staged KES algorithms (`SumKES` / `CompactSumKES`), we cannot create a new temporary variable for each sub-key at each stage - ideally, we want to allocate a single block of mlocked memory for the final key, and perform all operations on it, in-place.

Hence, we use the following approach:

- We extend the `MLockedSizedBytes` API with a function `mlsbMemcpy`, which, just like the `memcpy` C function, takes two pointers (`MLockedSizedBytes`) and a length, and copies this many bytes from one `MLockedSizedBytes` to the other. However, this function has two additional `offset` parameters, which allow us to address memory locations within the `MLockedSizedBytes`, and, unlike `memcpy`, we add bounds checking to prevent buffer overruns.
- The `DSIGNM` and `KESSignAlgorithm` classes expose two methods for serializing and deserializing their SignKeys; these methods accept a target `MLockedSizedBytes` value that they write to, and an offset into it. This is a slightly awkward signature for user code, but it allows us to delegate the serialization of sub-keys recursively.
- To deal with the awkwardness, then, we provide a pair of convenience wrappers that have more palatable signatures (`SignKey -> IO MLockedSizedBytes` for serialization, and `MLockedSizedBytes -> IO (Maybe SignKey)` for deserialization). These can't be used for recursive ser/deser of composite keys, but they are perfect for user code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ stack-local.yaml
 # ghcid
 **/.ghcid
 **/ghcid.txt
+tags
--- a/cabal.project
+++ b/cabal.project
@@ -14,6 +14,7 @@ packages:
   orphans-deriving-via
   slotting
   strict-containers
+  scrap
 
 -- Disable tests in a dependency (which are currently broken)
 package cardano-crypto
--- a/cabal.project
+++ b/cabal.project
@@ -14,7 +14,6 @@ packages:
   orphans-deriving-via
   slotting
   strict-containers
-  scrap
 
 -- Disable tests in a dependency (which are currently broken)
 package cardano-crypto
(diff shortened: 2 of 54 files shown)


iquerejeta:
closing in favour of #255 and #317

PR #277, opened 2022-05-16
Add `unsafeLinearExtendEpochInfo`.

nc6:
This commit adds a function which can extend an `EpochInfo` linearly
given a starting point. This is needed to put in place a patch for a bug
that arose in the Alonzo era, where the `EpochInfo` was too generous in
translating slots.

### DIFF
--- a/slotting/src/Cardano/Slotting/EpochInfo/Extend.hs
+++ b/slotting/src/Cardano/Slotting/EpochInfo/Extend.hs
@@ -1,0 +1,72 @@
+module Cardano.Slotting.EpochInfo.Extend where
+
+import Cardano.Slotting.EpochInfo.API (EpochInfo (..))
+import Cardano.Slotting.Slot (EpochNo (EpochNo), EpochSize (EpochSize), SlotNo (SlotNo))
+import Cardano.Slotting.Time
+  ( SlotLength (getSlotLength),
+    addRelativeTime,
+    multNominalDiffTime,
+  )
+
+-- | Given a basis point, use it and its slot length to impute a linear
+-- relationship between slots and time in order to extend an 'EpochInfo' to
+-- infinity.
+--
+-- The returned `EpochInfo` may still fail (according to the semantics of the
+-- specified monad) if any of the underlying operations fail. For example, if we
+-- cannot translate the basis point.
+unsafeLinearExtendEpochInfo ::
+  Monad m =>
+  SlotNo ->
+  EpochInfo m ->
+  EpochInfo m
+unsafeLinearExtendEpochInfo basisSlot underlyingEI =
+  let lastKnownEpochM = epochInfoEpoch_ underlyingEI basisSlot
+
+      goSize = \en -> do
+        lke <- lastKnownEpochM
+        if en <= lke
+          then epochInfoSize_ underlyingEI en
+          else epochInfoSize_ underlyingEI lke
+      goFirst = \en -> do
+        lke <- lastKnownEpochM
+        if en <= lke
+          then epochInfoFirst_ underlyingEI en
+          else do
+            SlotNo lkeStart <- epochInfoFirst_ underlyingEI lke
+            EpochSize sz <- epochInfoSize_ underlyingEI en
+            let EpochNo numEpochs = en - lke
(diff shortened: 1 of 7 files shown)


JaredCorduan (approved):
looks good to me

slotting/src/Cardano/Slotting/EpochInfo/Extend.hs
```diff
@@ -0,0 +1,72 @@
...
+            lke <- lastKnownEpochM
+            lkeStart <- epochInfoFirst_ underlyingEI lke
+            EpochSize sz <- epochInfoSize_ underlyingEI lke
+            let SlotNo slotsForward = sn - lkeStart
+            pure . (lke +) . EpochNo $ slotsForward `div` sz
+      goTime = \sn ->
+        if sn <= basisSlot
+          then epochInfoSlotToRelativeTime_ underlyingEI sn
+          else do
+            let basisSlotPred = pred basisSlot
```
lehins, line 58: Is it possible that `basisSlot == 0`? `pred` is a partial function

lehins (changes requested):
Some comments on improvements to the test suite.

slotting/test/Test/Cardano/Slotting/EpochInfo.hs
```diff
@@ -0,0 +1,47 @@
...
+  deriving (Eq, Show)
+
+instance Arbitrary TestSlotNo where
+  arbitrary = TestSlotNo . SlotNo <$> choose (0, 100)
+
+newtype TestEpochNo = TestEpochNo EpochNo
+  deriving (Eq, Show)
+
+instance Arbitrary TestEpochNo where
+  arbitrary = TestEpochNo . EpochNo <$> choose (0, 10)
```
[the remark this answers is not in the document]
nc6: There's not a great deal of point testing larger numbers

slotting/src/Cardano/Slotting/EpochInfo/Extend.hs
nc6: I've added a specific error in this case.

nfrisby: I may have proposed the predecessor (I don't recall). But I think we want successor.

The extrapolation we're doing is determined by the onset of the basis slot and the length of the basis slot. The length of the basis slot is the difference between its onset and the next slot's onset. By using `pred` you're taking the length of the predecessor of the basis slot. That'd be wrong when the basis slot is the first slot of an era that changes the slot length.

On the other hand, we only use this for Alonzo, and Mary -> Alonzo didn't change the slot length. So it's moot for our use-case.

Moreover, using the successor slot runs the risk that the onset of the basis slot is indeed the last time point the underlying `EpochInfo` _can_ translate, so we're unable to translate its successor. On the contrary: we'll always be able to translate the predecessor, assuming it exists.

I haven't thought through whether the onset of the successor of the current block's slot (which is what I assume the ledger passes for `basisSlot`) will always be translate-able. But I _think_ it would be (modulo the odd behavior you mentioned, specifically where the next era's safezone is 0).

A simpler option: add an `epochInfoSlotLength_` field to `EpochInfo`?

slotting/test/Test/Cardano/Slotting/EpochInfo.hs
```diff
@@ -0,0 +1,48 @@
...
+  ( Arbitrary (arbitrary),
+    choose,
+    testProperty,
+    (===),
+  )
+
+baseEpochInfo :: EpochInfo Identity
+baseEpochInfo = fixedEpochInfo (EpochSize 10) (slotLengthFromSec 10)
+
+extendedEpochInfo :: SlotNo -> EpochInfo Identity
```
nfrisby, line 20: Would you add a comment explaining that `extendedEpochInfo` should be `id` when applied to the result of `fixedEpochInfo`?

slotting/src/Cardano/Slotting/EpochInfo/Extend.hs
nc6: I went with the `epochInfoSlotLength_` option. This will also help with test failures in ledger, which tests a lot of things with `SlotNo 0`.

PR #280, opened 2022-05-25
Remove dependency on cardano-prelude

teodanciu:
There still is a dependency on `cardano-prelude-test` in 3 files

### DIFF
--- a/binary/cardano-binary.cabal
+++ b/binary/cardano-binary.cabal
@@ -58,7 +58,9 @@ library
                       , cborg              >= 0.2.2 && < 0.3
                       , containers
                       , data-fix
+                      , deepseq
                       , formatting
+                      , ghc-prim
                       , nothunks
                       , primitive
                       , recursion-schemes  >= 5.1   && < 5.3
@@ -102,3 +104,4 @@ test-suite test
 
   ghc-options:          -threaded
                         -rtsopts
+
--- a/binary/cardano-binary.cabal
+++ b/binary/cardano-binary.cabal
@@ -54,7 +54,6 @@ library
   build-depends:        base
                       , aeson
                       , bytestring
-                      , cardano-prelude
                       , cborg              >= 0.2.2 && < 0.3
                       , containers
                       , data-fix
@@ -87,7 +86,6 @@ test-suite test
   build-depends:        base
                       , bytestring
                       , cardano-binary
-                      , cardano-prelude
                       , cardano-prelude-test
                       , cborg
                       , containers
--- a/binary/cardano-binary.cabal
+++ b/binary/cardano-binary.cabal
@@ -24,7 +24,6 @@ common base                         { build-depends: base                       
 
 common project-config
   default-language:     Haskell2010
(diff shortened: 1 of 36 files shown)


lehins (approved):

teodanciu:
Hey @lehins , thanks for reviewing!  Do you mind merging this when you can?  I don't have rights on this repo to do it. Thanks!

lehins:
@L-as We are still relying on `cardano-prelude-test`, which itself depends on `cardano-prelude`. That's why we need it in `cabal.project`. I am not sure what would entail to get rid of `cardano-prelude-test`, but figuring it out is on our long TODO list.

L-as:
The current state of the ecosystem is problematic.
cardano-base has now copied over much of the code of cardano-prelude, yet many libraries still depend on cardano-prelude-test, hence there will very often be two HeapWords.
What is the plan for cardano-prelude-test? Why is this a thing? Why is cardano-prelude no longer a thing? What is supposed to happen in cardano-ledger?

Context: I am currently trying to port cardano-ledger to GHC 9.2.4, and am facing difficulties because of decisions like this.

L-as:
Found https://github.com/input-output-hk/cardano-ledger/pull/2992

lehins:
@L-as Very good point and reasonable questions.

> cardano-base has now copied over much of the code of cardano-prelude

Not much. From top of my head just these:

* two functions: `toCborError` + `cborError`. Latter two were moved into `cardano-binary`, where they really belong and should be removed from `cardano-prelude`
* and `HeapWords` moved into its own package `heapwords` that is part of this repo. In other words `HeapWords` from cardano-prelude` should simply be removed in favor of `heapwords` package.

Since everyone depends on `cardano-base` anyways they should use `heapwords` and `cardano-binary` instead for that functionallity.

> What is the plan for cardano-prelude-test? 

Not sure yet. As far as ledger code goes, mostly only byron tests depend on it.

> Why is cardano-prelude no longer a thing?

Two main reasons I can think of:

* `protolude` is very opinionated prelude, as most preludes are, and just not everyone agrees with it. I personally don't care for most of the functionality in that prelude.

* In my opinion `cardano-prelude` is improperly structured. It seems more like a dumping ground for sporadically useful functions, rather than a prelude. A proper prelude should bring best practices and libraries together for everyone to use. There should be curation of libraries with proper maintainership, neither of which is the case for `cardano-prelude`.

> What is supposed to happen in cardano-ledger?

Ledger code hasn't used `cardano-prelude` since Byron era. DISCLAIMER: Decision to stop using `cardano-prelude` was made even before my time at IOHK, but I fully support it. At my previous place of work I participated in development of [`rio`](https://hackage.haskell.org/package/rio), which is also an alternative prelude and a way better one than `protolude`. However, I am not advocating on switching to `rio`, I am just pointing out that we could do much better than `protolude` and overall idea of organization wide `cardano-prelude` is a good one, but with so many engineers in the company it will be hard to reach consensus on what a good prelude is.

L-as:
Thank you for the very in depth response.

PR #281, opened 2022-06-06
Disable VRF10 until the audit is complete

lehins:


iquerejeta (approved):
Great, def the best way to temporarily remove the requirement of the fork until the audit is finished.

PR #283, opened 2022-06-28
Cabal wrapped

deepfire:
Cabal is missing a feature to conditionalise the `source-repository-package` clauses -- this problem has been known for 1.5 years now.

Because of this missing feature, these clauses makes Cabal ignore the dependencies already provided by Nix, and so it has to get them and build on its own. ..and sometimes Cabal fails for no good reason.

The `cabalWrapped` shell wrapper creates a filtered cabal.project without these clauses: https://github.com/input-output-hk/iohk-nix/blob/master/overlays/utils/default.nix#L32-L78

This cuts down the amount of dependency supplementation work that Cabal has to do to `0`
Moreover, it ensures that Cabal doesn't deviate from Nix -- and hence from CI.

Which is how it is supposed to be.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dist-newstyle/
 .ghc.environment.x86_64-linux-8.6.5
 stack.yaml.lock
 cabal.project.local
+.nix-shell-cabal.project
 
 # Editors
 TAGS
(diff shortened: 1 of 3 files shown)


lehins:
Changes in this PR are unnecessary because packages do build fine inside nix-shell.
@deepfire could you please open an issue with an error you are getting, so we have a record of it and we can properly fix the issue.

deepfire:
@lehins, I think it's unfair to close this PR while denying the issue.

lehins (approved):

PR #284, opened 2022-06-30
Ts haddock annotated

TimSheard:
Add additional Haddock documentation to Annotated.hs

There was a problem in building this repo using nix-shell, Serge Kosyrev made some fixes which involved using cabalWrapped in the nix shell file that allowed things to work.  This was  done in https://github.com/input-output-hk/cardano-base/pull/283
I have made this PR on top of that.  Someone who knows this repository should make sure that Serges fixes are approriate.

### DIFF
--- a/binary/src/Cardano/Binary/Annotated.hs
+++ b/binary/src/Cardano/Binary/Annotated.hs
@@ -8,6 +8,32 @@
 {-# LANGUAGE TypeFamilies       #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
+-- | The CBOR class 'FromCBOR' does not support access to the original bytestring that is being deserialized.
+--   The 'Annotated' module recovers this ability by introducing several newtypes types that,
+--   along with some new operations, recover this ability.
+--
+-- 1. 'ByteSpan'  A pair of indexes into a bytestring, indicating a substring.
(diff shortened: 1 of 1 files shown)


JaredCorduan (approved):
The haddocks themselves look great to me!

I don't know enough about the nix changes, though. (Or if there is another way to get the haddocks besides exposing the module).

JaredCorduan:
This PR resolves https://github.com/input-output-hk/cardano-ledger/issues/2736

lehins (changes requested):
Haddock looks good, although examples are totally broken. In an ideal world we'd have haddock examples that are executed by doctests that are run in CI (opened #285 so we can one day get to the ideal world). Few things we need to do before merging this PR:

1. Regardless of doctest, some examples is better then no examples, but we do need to make sure that examples we do add to the haddock at least initially are correct. So, I'd like to recommend validating any haddock example locally in the future before adding them to documentation. Because no examples is actually better then _broken_ examples. As far as examples in this PR are concerned, I ran them locally and when all suggestions are applied the should be ok.
2. Removing commits that adjust nix configuration, that fix is not the correct one for the problem you had experience. In fact the PR #283 that introduced them has been closed.
3. Remove all trailing white space in all files that have been touched by this PR. It can be achieved `M-x whitespace-cleanup` in emacs

lehins (approved):
:+1: Looks great.

PR #286, opened 2022-07-01
Remove unused cardano-crypto dependency

lehins:
This is more of a cleanup PR:
* `cardano-crypto` is no longer used, since there was a full switch to libsodium a while back
* Update broken link and use suggestion from the documentation: remove
  `packages` from shell.nix, since default is all the project's local packages
* Bump up the cabal version to the latest one.


iquerejeta:
Can it be that plutus still relies on cardano-crypto?
cc: @ kwxm

kwxm:
> Can it be that plutus still relies in cardano-crypto? cc: @kwxm

Yes.

lehins:
> Can it be that plutus still relies in cardano-crypto?

@iquerejeta the fact that plutus uses the package is totally orthogonal to this PR, because this PR removes it from the cabal project in this repo only, not in plutus. Naturally, it is removed from the project only because `cardano-crpyto` is completely unused in the repo.

erikd (approved):

PR #287, opened 2022-07-05
Ts haddock annotated

TimSheard:
This is a modified version of https://github.com/input-output-hk/cardano-base/pull/284 which adds haddock docs to Annotated.hs
It is built on top of https://github.com/input-output-hk/cardano-base/pull/286 which fixes a problem, first identified in 
https://github.com/input-output-hk/cardano-base/pull/283, which might not be the best way to fix it.


lehins:
Closing as duplicate of #284

@TimSheard thank you for confirming that issue with cabal was caused by the wrong setting for the certificate authority

PR #288, opened 2022-07-12
Update index state to 2022-08-15T00:00:00Z

newhoggy:


lehins (approved):

L-as:
This broke master.

lehins:
It sure did. sources need to be updated too.

angerman:
Why didn't this break cardano-base's CI?

lehins:
@angerman That is a very good question. Could it be due to the `cabal-wrapper`?

hamishmack:
It looks like almost everything broke except the `required` job (the one that should aggregate everything we expect to work) https://hydra.iohk.io/eval/1267144

PR #293, opened 2022-07-28
Fix deserialization bug with Schnorr signatures, put repro test in place

kozross:
This is in relation to a test failure [in this PR](https://github.com/input-output-hk/plutus/pull/4786), as evidenced [here](https://github.com/input-output-hk/plutus/pull/4786#issuecomment-1197479222). The cause of this issue was that `secp256k1_xonly_pubkey_parse` does not (and cannot) check the length of the input it is given, which led it to read 'off the end' of the `ByteString` it was passed. Strangely, one specific case (the literal `"m"`) caused it to parse successfully: this has been included as a regression case to make sure this kind of issue does not arise again.

This partly came about as a result of the way we currently test serialization and deserialization. If we observe:

```haskell
prop_raw_serialise :: (Eq a, Show a)
                   => (a -> ByteString)
                   -> (ByteString -> Maybe a)
                   -> a -> Property
prop_raw_serialise serialise deserialise x =
    case deserialise (serialise x) of
      Just y  -> y === x
      Nothing -> property False
```

This verifies that _if_ a serialized form is valid, _then_ it will deserialize. However, it does _not_ check the other direction, so this does not verify the 'if and only if' this truly ought to be. The CBOR encoding checks are also vulnerable to this.

This should be addressed with a much more thorough set of tests, both property and unit, to ensure we don't get bitten by this again. I haven't included such tests here, as this PR is urgent.

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
@@ -186,14 +186,17 @@ instance DSIGNAlgorithm SchnorrSecp256k1DSIGN where
   rawSerialiseSignKeyDSIGN (SignKeySchnorrSecp256k1 skPSB) = psbToByteString skPSB
   {-# NOINLINE rawDeserialiseVerKeyDSIGN #-}
   rawDeserialiseVerKeyDSIGN bs = 
-    unsafeDupablePerformIO . unsafeUseAsCStringLen bs $ \(ptr, _) -> do
-      let dataPtr = castPtr ptr
-      (vkPsb, res) <- psbCreateSizedResult $ \outPtr ->
-        withForeignPtr secpCtxPtr $ \ctx -> 
-          secpXOnlyPubkeyParse ctx outPtr dataPtr
-      pure $ case res of 
-        1 -> pure . VerKeySchnorrSecp256k1 $ vkPsb
-        _ -> Nothing
+    unsafeDupablePerformIO . unsafeUseAsCStringLen bs $ \(ptr, len) ->
+      if len /= 32
+      then pure Nothing
+      else do
+        let dataPtr = castPtr ptr
+        (vkPsb, res) <- psbCreateSizedResult $ \outPtr ->
+          withForeignPtr secpCtxPtr $ \ctx -> 
+            secpXOnlyPubkeyParse ctx outPtr dataPtr
+        pure $ case res of 
+          1 -> pure . VerKeySchnorrSecp256k1 $ vkPsb
+          _ -> Nothing
   rawDeserialiseSignKeyDSIGN bs = 
     SignKeySchnorrSecp256k1 <$> psbFromByteStringCheck bs
   rawDeserialiseSigDSIGN bs = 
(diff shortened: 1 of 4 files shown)


cardano-crypto-tests/src/Test/Crypto/Regressions.hs
```diff
@@ -0,0 +1,29 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE OverloadedStrings #-}
+#ifdef SECP256K1_ENABLED
+{-# LANGUAGE TypeApplications #-}
```
JaredCorduan, line 4: Why do we need to dance around `TypeApplications`?

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -186,14 +186,17 @@ instance DSIGNAlgorithm SchnorrSecp256k1DSIGN where
...
-    unsafeDupablePerformIO . unsafeUseAsCStringLen bs $ \(ptr, _) -> do
-      let dataPtr = castPtr ptr
-      (vkPsb, res) <- psbCreateSizedResult $ \outPtr ->
-        withForeignPtr secpCtxPtr $ \ctx -> 
-          secpXOnlyPubkeyParse ctx outPtr dataPtr
-      pure $ case res of 
-        1 -> pure . VerKeySchnorrSecp256k1 $ vkPsb
-        _ -> Nothing
+    unsafeDupablePerformIO . unsafeUseAsCStringLen bs $ \(ptr, len) ->
+      if len /= 32
```
JaredCorduan, line 190: I think there is a way to get `32` out of `SchnorrSecp256k1DSIGN`, I think it's `SizeVerKeyDSIGN`.

cardano-crypto-tests/src/Test/Crypto/Regressions.hs
kozross: I get warnings otherwise.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
kozross: That involves `natVal`ing a `Proxy`. I'm happy to make this change though.

JaredCorduan: if it's not a huge inconvenience,  that would be grand!

JaredCorduan (approved):
Looks good to me. This is a very small and contained PR, and it removes only an obviously wrong case (where the length is < 32 bytes), and adds a test.

iquerejeta:
I guess that the correct parsing is not so much from the literal "m", but instead from the bytes "off the end". Are we making these checks for the other deserialisation functions? might be that the secp serialisation fails due the "m" literal not being a valid tag, and not because it has incorrect size. Are we sure the deserialisation of secp does not read bytes "off the end"? A nice additional test for secp would be to have a small literal, but which is a valid tag, such as for example `0x02`, `0x03` or `0x04`.

PR #294, opened 2022-08-03
Clear out psbFromBytes, quasiquoter

### LINKED ISSUE #291: Clear away psbFromBytes
As observed [here](https://github.com/input-output-hk/cardano-base/pull/289#discussion_r923916441), this function has some severe issues with referential transparency, and is thus generally not safe. While [this PR](https://github.com/input-output-hk/cardano-base/pull/289) deprecates both `psbFromBytes` (and `psbZero`) using a `DEPRECATED` pragma, as well as removing its uses internal to `cardano-base` itself, this function is still used internally to `cardano-crypto-class`:

* `psbFromByteString`
* `IsString (PinnedSizedBytes n)`

While these should both be modified to _not_ use `psbFromBytes`, the `IsString` instance is rather horrifying: much like the same instance for `ByteString`, is hugely problematic for mostly the same reasons. I would propose removing this `IsString` instance wholesale, and replacing it with a quasi-quoter.

kozross:
This attempts to resolve #291. In particular:

* All functions using `psbFromBytes` are gone.
* `IsString PinnedSizedBytes` is gone.
* Quasi-quoter `psbHex` has been added for compile-time construction.
* Small tests for quasi-quoter logic.

Some refactors to enable the quasiquoter have also been added.

<s>Currently, I'm submitting this as a draft, as I can't (quite) figure out how to stop the quasiquoter making two identical hex strings construct `PinnedSizedBytes` which share the same address. This is reproduced by one of the tests, which is currently failing. I attempted to address this by porting over `primitive`'s `Lift` instance for `ByteArray`, but that doesn't seem to help any.</s>

Upon further attempts, I'm basically giving up on this, and leaving a note about not mutating the outcomes of the quasiquoter. This is basically for two reasons:

* No matter _how_ we do this, GHC can work against us and let-float things, then CSE its way to victory. While there _was_ some discussion to allow explicit blocking of this in GHC itself, we basically don't have it in any reliable way.
* If you're using a quasiquoter, you're probably defining a constant anyway: I see no reason to want to mutate it, unless it's a bug.

@lehins - tagging you for a review.


cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
```diff
@@ -0,0 +1,138 @@
...
+import Data.Functor (($>))
+import Data.Foldable (traverse_)
+
+-- | Constructs (and verifies) a compile-time-known hex string as a
+-- 'PinnedSizedBytes', as well as computing its length at the type level.
+--
+-- The syntax accepted by this quasiquoter only works in an expression context.
+-- A valid input consists of the literal sequence @0x@ followed by either:
+--
+-- * A single underscore (@_@); or
```
tdammers, line 54: This seems awkward and unnecessary to me.

The only thing that this quasiquoter will ever accept is hex-encoded byte arrays, so there is no need for the leading "0x" (though you could of course make it optional, just to cater for people's motor memory), and the underscore is needlessly complex, both for the user and for the parser. I would propose the following syntax:

- Empty byte array: `[psbHex| |]` or `[psbHex| 0x |]`
- Non-empty byte array: `[psbHex| deadfood |]` or `[psbHex| 0xdeadfood |]`

tdammers (changes requested):

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
```diff
@@ -0,0 +1,138 @@
...
+-- * A positive, even number of hexits (digits @0@ through @9@, or letters @a@
+-- through @f@, lower-case only).
```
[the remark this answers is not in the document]
kozross: Should we allow 'mixing'? For example, should `0xAb12cD` be acceptable?

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
kozross: `0x` is technically correct, but my sentiment is that it looks like a typo, which is why I had the underscore to indicate specifically 'I want an empty `PinnedSizedBytes` literal'. I'm happy to have it be `0x` for that meaning.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/Internal.hs
```diff
@@ -0,0 +1,115 @@
...
+-- = Note
+--
+-- The 'Ptr' given to the function argument /must not/ be used as the result of
+-- type @r@.
+psbUseAsCPtr :: 
+  forall (n :: Nat) (r :: Type) .
+  PinnedSizedBytes n -> 
+  (Ptr Word8 -> IO r) -> 
+  IO r
+psbUseAsCPtr (PSB ba) = runAndTouch ba
```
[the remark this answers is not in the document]
kozross: Yeah, but we need some way to read the bytes as a `Ptr` to use with the FFI. We could add a 'please do not mutate me' note, but without functionality such as this, it's nearly impossible to use `PinnedSizedBytes` for its main intended purpose.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
kozross: I want to keep the `0x` as it makes copy-pasting literals easier from other sources, such as files or examples. This is necessary for me, as I have outstanding test work for specific SECP-related test vectors.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
```diff
@@ -0,0 +1,138 @@
...
+--
+-- * @['psbHex'| 0x_ |]@ produces a 'PinnedByteString' of length 0.
+-- * @['psbHex' | 0xafbc |]@ produces a 'PinnedByteString of length 2; the 
+-- first byte is @0xaf@, and the second is @0xbc@.
+--
+-- = Notes
+--
+-- To use this quasiquoter, @DataKinds@ and @TypeApplications@ need to be on.
+--
+-- Anything produced by this quasiquoter should be seen as immutable; this is
```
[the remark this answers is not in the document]
kozross: Fair point, removing.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/Internal.hs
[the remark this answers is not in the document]
tdammers: In that case, I would suggest adding both the "do not mutate" note (in Haddock, not just inline comment), and also renaming the function to `unsafeUsePSBAsCPtr` or something equally awkward and scary-looking, in good old Haskell tradition.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -1,24 +1,17 @@
...
 
--- for pinnedByteArrayFromListN
-{-# OPTIONS_GHC -Wno-missing-local-signatures #-}
 module Cardano.Crypto.PinnedSizedBytes
   (
     PinnedSizedBytes,
-    -- * Initialization
-    psbZero,
+    -- * Quasiquoter
+    psbHex,
```
tdammers, line 12: Personally, I would not re-export `psbHex` from the main module; most users will *not* need the TH functionality, and for those who do, explicitly importing the `.TH` module won't be prohibitive (and might actually be useful as a reminder that we're using the TH functionality).

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
```diff
@@ -0,0 +1,124 @@
...
+  quotePat = qqUseError "pattern",
+  quoteType = qqUseError "type",
+  quoteDec = qqUseError "declaration"
+  }
+
+-- Helpers
+
+mkHexLiteral :: String -> Q Exp
+mkHexLiteral input = do
+  let asStrippedText = Text.strip . Text.pack $ input
```
tdammers, line 88: This may be one of the rare cases where going through `Text` instead of keeping things as `String` isn't helpful. We're traversing the string exactly once, sequentially, and at each step, we just unpack a 2-character substring and feed it to `readMaybe`. Might as well do it all on a `String`. This isn't going to be performance critical anyway, since it runs at compile time, and getting rid of the `Text.{un}pack` bits would reduce visual clutter.

tdammers (commented):
Looks good to me, apart from the one remark on the previous review.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
kozross: The flip side is that stripping whitespace (leading and trailing) would get more difficult. I don't mind making this change if you insist, but I personally think it's easier to work with `Text` in this case.

tdammers: Fair enough.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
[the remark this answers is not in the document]
tdammers: Correct me if I'm wrong, but it looks to me as if the code already supports uppercase hexits, however this comment suggests it doesn't.

tdammers (changes requested):

kozross:
@tdammers - your comment has already been addressed a few commits ago. I can't reply to it, as it's marked 'outdated'.

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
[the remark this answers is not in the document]
tdammers: Has been fixed.

tdammers (approved):

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes/TH.hs
```diff
@@ -0,0 +1,125 @@
...
+natLiteral = LitT . NumTyLit
+
+decodeAndCount :: Text -> StateT Integer Q Word8
+decodeAndCount chunk
+  | Text.length chunk /= 2 = do
+      len <- get
+      fail $ "Odd number of hexits (" <> show (2 * len + 1) <> ")."
+  | otherwise = case readMaybe ("0x" <> Text.unpack chunk) of 
+      Nothing -> fail $ "Invalid hexit pair (" <> Text.unpack chunk <> ")."
+      Just w8 -> modify (+ 1) $> w8
```
lehins, line 118: Why are you reimplementing from hex conversion. Why not just use bytestring-base16

lehins (changes requested):
I don't particularly like the reimplementation of hex conversion. For performance, maintainability and correctness reasons. How can we change that?

lehins:
@kozross Thank you for doing all the work, but I'd rather implement this feature in a different way. Please review this PR if you can: #299

lehins:
Closing in favor of #299 

@kozross Thank you on your effort on this, but quasiquoter only allows decoding into a ByteString, thus it is less powerful than the TH solution supplied in #299, which allows static check of the length as well, thus allows decoding into Hash and PinnedSizedBytes directly.

PR #295, opened 2022-08-07
Additional deserialization tests, fix bug in Ed25519 deserialization

kozross:
In a bid to prevent future issues similar to the one addressed [by this PR](https://github.com/input-output-hk/cardano-base/pull/293), I have added some 'negative' deserialization tests. More precisely, the tests work as follows:

* Generate a serialized form of the wrong length
* Attempt to deserialize it
* Confirm you get `Nothing`

As part of this, I fixed a similar issue to #293 in the Ed25519 code where a length check had been omitted.

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Ed25519.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Ed25519.hs
@@ -32,7 +32,7 @@ import GHC.Generics (Generic)
 import NoThunks.Class (NoThunks)
 import System.IO.Unsafe (unsafeDupablePerformIO)
 import GHC.IO.Exception (ioException)
-import Control.Monad (unless)
+import Control.Monad (unless, guard)
 import Foreign.C.Error (errnoToIOError, getErrno)
 import Foreign.Ptr (castPtr, nullPtr)
 import qualified Data.ByteString as BS
@@ -181,7 +181,9 @@ instance DSIGNAlgorithm Ed25519DSIGN where
     rawSerialiseSigDSIGN      (SigEd25519DSIGN sig) = psbToByteString sig
 
     rawDeserialiseVerKeyDSIGN  = fmap VerKeyEd25519DSIGN . psbFromByteStringCheck
-    rawDeserialiseSignKeyDSIGN = Just . genKeyDSIGN . mkSeedFromBytes
+    rawDeserialiseSignKeyDSIGN bs = do
+      guard (BS.length bs == 32)
+      pure . genKeyDSIGN . mkSeedFromBytes $ bs
     rawDeserialiseSigDSIGN     = fmap SigEd25519DSIGN . psbFromByteStringCheck
 
 
--- a/cardano-crypto-tests/src/Test/Crypto/Util.hs
+++ b/cardano-crypto-tests/src/Test/Crypto/Util.hs
@@ -4,6 +4,10 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE RoleAnnotations #-}
+{-# LANGUAGE TupleSections #-}
 
 module Test.Crypto.Util
   ( -- * CBOR
@@ -15,6 +19,7 @@ module Test.Crypto.Util
   , prop_cbor_valid
   , prop_cbor_roundtrip
   , prop_raw_serialise
+  , prop_raw_deserialise
   , prop_size_serialise
   , prop_cbor_direct_vs_class
 
@@ -32,14 +37,40 @@ module Test.Crypto.Util
 
    -- * test messages for signings
   , Message(..)
+
+    -- * Test generation and shrinker helpers
+  , BadInputFor
+  , genBadInputFor
+  , shrinkBadInputFor
+  , showBadInputFor
   )
 where
 
-import Cardano.Binary (FromCBOR (..), ToCBOR (..),
-                       Encoding, Decoder, Range (..),
-                       decodeFullDecoder, serializeEncoding, szGreedy, szSimplify)
-import Codec.CBOR.FlatTerm
-import Codec.CBOR.Write
+import Numeric (showHex)
+import GHC.Exts (fromListN, fromList, toList)
+import Text.Show.Pretty (ppShow)
+import Data.Kind (Type)
+import Cardano.Binary (
+  FromCBOR (fromCBOR), 
+  ToCBOR (toCBOR),
+  Encoding, 
+  Decoder, 
+  Range (Range),
+  decodeFullDecoder, 
+  serializeEncoding, 
+  szGreedy, 
+  szSimplify,
+  lo, 
+  hi,
+  encodedSizeExpr
+  )
+import Codec.CBOR.FlatTerm  (
+  validFlatTerm,
+  toFlatTerm
+  )
+import Codec.CBOR.Write (
+  toStrictByteString
+  )
 import Cardano.Crypto.Seed (Seed, mkSeedFromBytes)
 import Cardano.Crypto.Util (SignableRepresentation(..))
 import Crypto.Random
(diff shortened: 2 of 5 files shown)


cardano-crypto-tests/src/Test/Crypto/Util.hs
```diff
@@ -174,6 +204,20 @@ prop_raw_serialise serialise deserialise x =
       Just y  -> y === x
       Nothing -> property False
 
+prop_raw_deserialise ::
```
tdammers, line 207: This name is a bit unintuitive - it suggests a "positive" property of deserialization, similar to how `prop_raw_serialize` tests a positive property of serialization; however, `prop_raw_deserialise` does not verify that deserialization succeeds when it should, but rather that it does not succeed in two specific scenarios (input too long, and input too short).

cardano-crypto-tests/src/Test/Crypto/Util.hs
```diff
@@ -174,6 +204,20 @@ prop_raw_serialise serialise deserialise x =
...
+prop_raw_deserialise :: 
+  forall (a :: Type) .
+  (Show a) =>
+  (ByteString -> Maybe a) -> 
+  BadInputFor a -> 
+  Property
+prop_raw_deserialise deserialise (BadInputFor (forbiddenLen, bs)) = 
+  checkCoverage . 
+  cover 50.0 (BS.length bs > forbiddenLen) "too long" .
+  cover 50.0 (BS.length bs < forbiddenLen) "too short" $
```
tdammers, line 216: This isn't fully orthogonal:

1. We're conflating two error conditions into one here ("too long", and "too short")
2. There is a potential third error condition that can interfere, at least from a black-box perspective: if the input is too long, *and* contains an invalid serialization, the test may successfully fail even without a length test, and since we're neither forcing the input to be well-formed, nor checking why the deser failed, we cannot tell this from the correct failure (reject due to input being too long). I I think we'll get away with this one here, because any bytestring of the correct size is a valid input (after all, the serialization format is just a literal dump of the key bytes, which themselves are generated by just taking an appropriate number of bytes from an entropy source), but I still think it's a good habit to avoid these gotchas, and it would make it more obvious what it is we're testing here.

What I would suggest:

- Split the property up into two separate ones, one for the "too short" case, and one for "too long". (This should also allow us to get rid of the awkward coverage check).
- In both cases, instead of passing a random input, generate a known-valid input, and then make it invalid by adding or removing bytes at the end.

tdammers (approved):
Fine by me; I think the tests could be improved (see comments), but I don't think this is prohibitive, and if deemed necessary, we can always do that after merging.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Ed25519.hs
```diff
@@ -181,7 +181,9 @@ instance DSIGNAlgorithm Ed25519DSIGN where
     rawSerialiseSigDSIGN      (SigEd25519DSIGN sig) = psbToByteString sig
 
     rawDeserialiseVerKeyDSIGN  = fmap VerKeyEd25519DSIGN . psbFromByteStringCheck
-    rawDeserialiseSignKeyDSIGN = Just . genKeyDSIGN . mkSeedFromBytes
+    rawDeserialiseSignKeyDSIGN bs = do
+      guard (BS.length bs == 32)
```
lehins, line 185: We should do better than just hardcode `32`.
```suggestion
      guard (fromIntegral (BS.length bs) >= seedSizeDSIGN (Proxy @Ed25519DSIGN)))
```
Also it should have at least 32 bytes, not exactly 32. `genKeyDSIGN` documentation says so and when you look at its implementation it actually slices the supplied bytes and only uses the first 32:
https://github.com/input-output-hk/cardano-base/blob/18be4924a2e21261cb1928d8e48384265770f9df/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Ed25519.hs#L162

lehins (changes requested):
I haven't delved too deep into the test suite changes yet, I'll look at them next pass. Came back from vacation yesterday, still catching up

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Ed25519.hs
kozross: I don't think blindly deserializing more than the expected size and then dropping the rest is good behaviour here.

lehins: I actually completely agree with you that dropping is not a good behavior, but it was the original design and we need to keep in mind that this is a breaking change. Some code that has worked before might no longer work in the same way. If it was a compilation error I'd be fine, but that will be a different runtime behavior. So we need to investigate what sort of affect it has on downstream users of this function.

It was previously protected for `< 32`, it would just throw a run time error, which is ugly, but it is better than segfault. However it did work fine for `>= 32` bytes, which is a documented feature. So, we can't just change this behavior because we believe it is a good idea, we need to be sure we don't cause any problems while we do that

lehins: I've doubled checked the cardano-ledger and I believe it is ok to make it more strict and enforce `== 32`, because we never deserialize any signing keys. Although we still do have to change it to use `seedSizeDSIGN`.

@iquerejeta and @michaelpj do you guys have any opinions on this problem? I don't think it should cause any issues for node. But I am just not sure if it is used anywhere else, like in Plutus for example (in particular deserialization of Ed25519 signing keys).

iquerejeta: Good catch @lehins. Might affect the node as this deserialisation is used in KES, and I wouldn't be surprised if KES key serialisation is > 32 bytes as it not only contains the ed25519 seed, but also the period. It seems that the byte string is split with the correct size, but could someone check?
https://github.com/input-output-hk/cardano-base/blob/1ea947d123981193a6055a193bc2b0e5fd5001ba/cardano-crypto-class/src/Cardano/Crypto/KES/Simple.hs#L177

For Plutus I believe it is fine, as we are still using cardano-crypto there.

lehins: Yep, that's exactly what I was afraid we could be overlooking. It does affect `rawDeserialiseSignKeyKES` because implementation does chunk up the ByteString into 32 byte long seeds, except the last one, which might be bigger than 32 bytes. We should fix that implementation regardless if we enforce the stricter check or not in `rawDeserialiseSignKeyDSIGN`

lehins: Actually I take it back `rawDeserialiseSignKeyKES` is not affected, because it will produce exactly `duration` number of 32 byte chunks and if it there is leftover it will produce Nothing. In other words it will fail if the supplied ByteString is not the exact length `sizeVerKeyDSIGN d * t`

iquerejeta: Glad to see it does not affect KES. But looking at the KES code, it seems that a `SignKeySimpleKES` holds all the remaining secret keys. Am I interpreting that right?
https://github.com/input-output-hk/cardano-base/blob/master/cardano-crypto-class/src/Cardano/Crypto/KES/Simple.hs#L58
@tdammers

iquerejeta: (not that the above affects these changes, just trying to understand how the KES key is formed)

lehins: KES key is formed by splitting input bytestring into exactly `duration` number of 32 byte chunks and each individual is converted to `VerKeyDSIGN`, all of which afterwards become `VerKeySimpleKES`. If the supplied ByteString has at least one more or one less byte than expected it will return `Nothing` thanks to this check: `length vkbs == duration`

iquerejeta: I take that you used the notation of `VerKeySimpleKES`, but the same holds for `SignKeySimpleKES` (so a sign key is converted into exactly `duration` `VerKeyDSIGN` keys).

lehins: My bad, I meant `SignKeySimpleKES` and `SignKeyDSIGN`. But yeah, `VerKeySimpleKES` has the same logic

iquerejeta: @lehins , I believe then that we are good to proceed with the strict criteria of size == 32. Or you still want more guarantees that this is not breaking elsewhere?

lehins: Yeah, I think so. This PR still need this change: `32` -> `seedSizeDSIGN (Proxy @Ed25519DSIGN)`

Other than that if you say its good it is good from my side.

PR #297, opened 2022-08-11
Newtype deriving type classes for `StrictFingerTree`.

jorisdral:
The `FingerTree` datatype has a number of type class instances that we
could newtype-derive for `StrictFingerTree`. This PR auto-derives
these instances for `StrictFingerTree`.

### DIFF
--- a/strict-containers/src/Data/FingerTree/Strict.hs
+++ b/strict-containers/src/Data/FingerTree/Strict.hs
@@ -73,11 +73,8 @@ infixl 5 |>
 -- wrap the original "Data.FingerTree" functions while forcing the provided
 -- value to WHNF.
 newtype StrictFingerTree v a = SFT {fromStrict :: FT.FingerTree v a}
-  deriving (Eq, Ord, Show)
-
-deriving newtype instance Foldable (StrictFingerTree v)
-
-deriving newtype instance (Measured v a) => Measured v (StrictFingerTree v a)
+  deriving stock (Eq, Ord, Show)
+  deriving newtype (Foldable, Semigroup, Monoid, Measured v)
 
 instance NoThunks a => NoThunks (StrictFingerTree v a) where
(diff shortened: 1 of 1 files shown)


strict-containers/src/Data/FingerTree/Strict.hs
```diff
@@ -73,9 +73,8 @@ infixl 5 |>
 -- wrap the original "Data.FingerTree" functions while forcing the provided
 -- value to WHNF.
 newtype StrictFingerTree v a = SFT {fromStrict :: FT.FingerTree v a}
-  deriving (Eq, Ord, Show)
-
-deriving newtype instance Foldable (StrictFingerTree v)
+  deriving (Eq, Ord, Show, Generic)
```
amesgen, line 76: As `DerivingStrategies` is already enabled:
```suggestion
  deriving stock (Eq, Ord, Show, Generic)
```

amesgen, line 76: Deriving `Generic` allows to break the abstraction barrier, in particular get an iso `StrictFingerTree ↔ FingerTree` (which allows thunks in the values to sneak back in). Are you intending to extensively use the `Generic` instance?

strict-containers/src/Data/FingerTree/Strict.hs
```diff
@@ -73,9 +73,8 @@ infixl 5 |>
...
+  deriving newtype (Foldable, Semigroup, Monoid)
 
 deriving newtype instance (Measured v a) => Measured v (StrictFingerTree v a)
```
amesgen, line 79: What about moving `Measured v` into the `deriving newtype (...)` just above like `Foldable`?

amesgen (changes requested):

amesgen (approved):

PR #308, opened 2022-09-20
Rename `strict-containers` to `cardano-strict-containers`.

jonathanknowles:
Hackage currently has a package called [`strict-containers`](https://hackage.haskell.org/package/strict-containers), which aims to provide:

- fully-strict versions of the data structures exported by the [`containers`](https://hackage.haskell.org/package/containers) package
- strong API compatibility guarantees with the [`containers`](https://hackage.haskell.org/package/containers) package.

This commit renames the `strict-containers` library in this repository so that its name does not collide with the above Hackage package, thus making it possible for downstream projects to consume both the version provided by `cardano-base` and also the version provided by the `strict-containers` library on Hackage.

(Without this renaming, it is difficult to consume both libraries, as cabal does not provide functionality to disambiguate identically-named packages.)


michaelpj (approved):

michaelpj:
cc @lehins

lehins (approved):
Very nice rename. Thank you for submitting this PR

jonathanknowles:
@lehins wrote:
> Very nice rename. Thank you for submitting this PR

You're welcome, thanks for merging!

andreabedini:
> cc @andreabedini this one has snuck into https://github.com/input-output-hk/cardano-haskell-packages/tree/main/_sources/strict-containers/0.1.0.0 already :/

should we just drop it? I think we will do many more little mistakes and perhaps we need to get comfortable with little breaking. Worse case scenarios are 1) someone added CHaP to a project where strict-containers was used and didn't realise it was getting a different package (unlikely) 2) someone add CHaP to a project and depended on strict-containers (cardano version) intentionally and now it has to rename like here.

michaelpj:
Well, the node depends on it (and a bunch of other stuff like ouroboros-network), so we can't drop it until the rename has propagated.

PR #310, opened 2022-09-27
Fix nix setup

michaelpj:
- Fix CHaP URL
- Add CHaP to niv and pass in source
- Update haskell.nix to handle additional repositories
- Fix overriding of hackage.nix and update hackage.nix
- Simplify shell and include cabal

It would be nice if this was being tested in CI, but :shrug:


michaelpj:
Wait, it _is_ in CI. How did the previous PR actually pass? :thinking: It had an unknown Hackage index state and a missing package repository?

michaelpj:
Wow, the `release.nix` is totally broken and doesn't actually depend on any of the Haskell packages. I'm not going to fix that now, maybe you want to take a look @angerman

newhoggy (approved):

PR #316, opened 2022-10-12
Improve Praos conversion safety and performance

lehins:
There are a few different improvements in this PR to Praos implementation:

* Expose internal types (but not constructors)
* Avoid redundant copy in `copyFromByteString`
* Make conversion function fail with `Maybe`, thus making CBOR decoders total
* Fix return types for two FFI calls (they never returned an Int, if it was some future compatibility it was wrong)
* Remove useless `io_crypto_vrf_publickeybytes`, `io_crypto_vrf_secretkeybytes` - not used anywhere on github, thus safe to remove
* Remove internal exports: `crypto_vrf_proofbytes`, `crypto_vrf_publickeybytes` , `crypto_vrf_secretkeybytes` , `crypto_vrf_seedbytes` and `crypto_vrf_outputbytes`
* Remove nowhere to be used `unsafeRawSeed`


cardano-crypto-praos/src/Cardano/Crypto/VRF/Praos.hs
```diff
@@ -1,36 +1,23 @@
...
 module Cardano.Crypto.VRF.Praos
   (
   -- * VRFAlgorithm API
     PraosVRF
 
-  -- * Low-level size specifiers
-  --
-  -- Sizes of various value types involved in the VRF calculations. Users of
-  -- this module will not need these, we are only exporting them for unit
-  -- testing purposes.
```
JaredCorduan, line 24: so these were not being used in unit tests, despite the comment?

JaredCorduan (approved):
LGTM

cardano-crypto-praos/src/Cardano/Crypto/VRF/Praos.hs
lehins: Exactly. There are also no real reasons to export them either since there are more natural ways to get to those values: `sizeVerKeyVRF`, `sizeSignKeyVRF`,  `sizeCertVRF`, ` sizeOutputVRF` and `seedSizeVRF`

PR #318, opened 2022-10-14
Remove un-needed constraint on time

erikd:


andreabedini:
Why is it unneded? a downstream consumer needs toknow what versions of time are compatible with cardano-slotting.

erikd:
`ledger` had `allows-newer: cardano-slotting:time` so its not needed.

andreabedini:
> ledger had allows-newer: cardano-slotting:time so its not needed.

That tells me that ledger is ignoring that constraint but what about other users?

We are publishing cardano-slotting to CHaP so that more people can use our libraries. If anything we need to put back all the bounds that are currently missing.

erikd:
> That tells me that ledger is ignoring that constraint but what about other users?

Other users can set whatever `time` constraint they want and it will JustWOrk(tm). This is a vastly better solution than what we had which was setting an un-necessarily narrow `constraint` which could clash with upstream constraints requiring hacks like `allow-newer`.

andreabedini:
> Other users can set whatever time constraint they want and it will probably JustWork(tm).

It will not. The `time` package will have new major releases with breaking changes, which will break our code.

> If we do that to all our packages we will likely end up spending 100 hours a week just maintaining version constraints in out bazillion packages.

Without constraints, it will be our users spending 100 hours a week trying to figure out what set dependencies work together. Even if we won't see it.

erikd:
We should encourage the use if `index-state` statements in the `cabal.project` file.

Using `index-state` like this means that a new version of `time` **cannot** break anything unless `index-state` is changed.

andreabedini:
> We should encourage the use of index-state statements in the cabal.project file.

plutus uses [index-state: 2022-09-26T00:00:00Z](https://github.com/input-output-hk/plutus/blob/1f58c2650c8d7407a7d248eea87969f708efb770/cabal.project#L16)
ouroboros-network uses [index-state: 2022-07-01T00:00:00Z](https://github.com/input-output-hk/ouroboros-network/blob/0dcda09b6958205fe362524a7e5ea35dea320196/cabal.project#L1)

Which one should I use?

Also this means that downstream consumers will never be able to release a library, since using that new library would require bumping the index-state.

The critical information (i.e. what works with what) is in the dependency version bounds. It's not stored anywhere else (perhaps other than the mind of the release manager). Even making CHaP @michaelpj and I are pulling our hair trying to figure out the missing "what works with what" information.

erikd:
> Which one should I use?

The latest one that works.

> Also this means that downstream consumers will never be able to release a library, since using that new library would require bumping the index-state.

They can bump and if they run into trouble with something like `time` they can add `constraints: time < <newest-version>`.

Per-project constraints in the `<project>.cabal` file worked reasonably well for quite some time, but IMO they have become increasing painful over time.

michaelpj:
I think bounds is a big topic. For now I think what is uncontroversial is adding bounds that _we know exclude breakage_. Clearly we didn't know that the upper-bound on `time` was necessary, so it got ignored. The lower-bound may well be necessary. If we discover that we need an upper-bound we can put it in or revise it.

I'll start a discussion about the bigger topic elsewhere.

lehins (approved):
@erikd Thank you for finding and removing these invalid bounds. 

@andreabedini We rely on index-states which means bounds only need to be specified whenever packages is definitely known to be broken with a particular version of a dependency

andreabedini:
index-state is not part of the package description and doesn't get propagated to the consumers of the library. I had to spend a few hours trying to figure out why cardano-crypto-class wasn't compiling "anymore", where "anymore" meant someone had a different index-state. The culprit was a lacking of upper bound on vector.

PR #319, opened 2022-10-14
Bump up the version and add changelog for `cardano-crypto-praos`

lehins:
We've switched recently to CHaP, but recent PR #316 did not update changelog and did not bump up the version


JaredCorduan (approved):
:raised_hands:

PR #324, opened 2022-10-25
Update cabal.project

angerman:
This is here to see if what we have in 2.0.0.1 actually builds today.


lehins:
Looks like an ephemeral PR that was mistakenly left opened. Feel free to re-open if my assumption is wrong

PR #325, opened 2022-10-25
Add upper version bound on vector

andreabedini:
After spending a couple of hours trying to figure out why cardano-crypto-class was not compiling in certain repos.

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -111,7 +111,7 @@ library
     , th-compat
     , text
     , transformers
-    , vector
+    , vector < 0.13
 
   if impl(ghc < 9.0.0)
     build-depends:


michaelpj:
Comment explaining why this helps, or do we not know?

michaelpj (approved):

michaelpj:
Like, did vector do something weird in this version?

andreabedini:
Well, nothing weird other than a breaking change (justifying the 0.12.x.y -> 0.13.0.0 transition). To be honest I don't understand the problem very well. The error is

```
src/Cardano/Crypto/KES/Simple.hs:195:10: error:
Error:     • Could not deduce (NoThunks (Vector (SignKeyDSIGN d)))
        arising from a use of ‘NoThunks.Class.$dmwNoThunks’
      from the context: DSIGNAlgorithm d
        bound by the instance declaration
        at src/Cardano/Crypto/KES/Simple.hs:195:10-66
    • In the expression:
        NoThunks.Class.$dmwNoThunks @(SignKeyKES (SimpleKES d t))
      In an equation for ‘NoThunks.Class.wNoThunks’:
          NoThunks.Class.wNoThunks
            = NoThunks.Class.$dmwNoThunks @(SignKeyKES (SimpleKES d t))
      In the instance declaration for
        ‘NoThunks (SignKeyKES (SimpleKES d t))’
    |
195 | instance DSIGNAlgorithm d => NoThunks (SignKeyKES (SimpleKES d t))
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The [changelog](https://hackage.haskell.org/package/vector-0.13.0.0/changelog) mentions a change in the role signatures but I am not sure how that could lead to the problem.

michaelpj:
Hmm, so maybe it's a `nothunks` issue we should raise that the vector instances don't seem to work with `vector-
0.13`

lehins:
Yes, the issue should be fixed. Adding an upper bound is simply masking a problem and delaying someone finding a solution. That is exactly why adding restrictive upper bounds proactively is a bad thing. The weird part is that `nothunks` already has an upper bound on vector (`vector >=0.12 && <0.13`), so I have no idea how vector-0.13 was used in the first place.

In any case roles indeed have changed in latest vector and it does prevent instance deriving (for good reasons) and I guess affects default implementation using generics as well. 

I've opened a ticket to solve this issue properly: #327

andreabedini:
I want the developers to fix the problem of supporting newer versions of vector.

I don’t want the users to fix the problem of why cardano-crypto-class doesn’t compile anymore.

andreabedini:
> The weird part is that nothunks already has an upper bound on vector (vector >=0.12 && <0.13), so I have no idea how vector-0.13 was used in the first place.

This is confusing indeed. I’ll have another look.

michaelpj:
> Adding an upper bound is simply masking a problem and delaying someone finding a solution.

In this case I disagree: adding an upper bound here means that the developers of `cardano-crypto-class` (i.e. us) have noticed there is a problem and can pursue a fix; but it prevents any _users_ of `cardano-crypto-class` from re-discovering the same problem repeatedly until we fix it. If we _know_ a dependency doesn't work, then it's just _true_ to add the upper bound. But I agree that we should mainly try and fix the problem.

lehins:
> Adding an upper bound is simply masking a problem and delaying someone finding a solution.

@michaelpj and @andreabedini My apologies, looks like I worded it poorly and I can now see how unclear was the point that I was trying to make. I was trying to say that we would have not discovered this issue if the upper bound was there in the first place. Moreover, I was reading comments on this PR in my email and did no realize the the comment I was addressing was not even in this thread: https://github.com/input-output-hk/cardano-base/pull/318#issuecomment-1290158102 So, the culprit was not the lack of upper bound, but the fact that `cardano-crypto-class` was not compatible with `vector-0.13`

In other words the proper fix should be making the library compatible. That beeing said, I do agree that in such a situation if a fix is not feasible due to some reasons like complexity of a fix or lack of resources for the task, then adding an upper bound is a totally sensible thing to do and I would always encourage it. So I am definitely not against the interim fix of adding an upper bound, as was done in this PR.

PR #336, opened 2022-11-03
Mempool

lehins:
This is a minimal viable implementation of a lockfree threadsafe mempool that can be used with libsodium's mlock functionality


cardano-mempool/tests/Common.hs
```diff
@@ -0,0 +1,7 @@
+module Common
+  ( module X
+  ) where
+
+import Test.Hspec as X
```
tdammers, line 5: So far we've been using Tasty in cardano-base; if it's not prohibitive, I think it would be a good idea to rewrite this to use Tasty as well, for consistency and a smaller dependency footprint.

cardano-mempool/tests/Test/Cardano/Memory/PoolSpec.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Test.Cardano.Memory.PoolSpec (spec) where
```
tdammers, line 5: Some additional things I'd like to see tested:

- Calling finalizers explicitly (`finalizeForeignPtr`)
- Interleaving deallocations with allocations (as discussed, this is more likely to hit code paths where pages get fragmented, and blocks from within an existing page are returned)

cardano-mempool/src/Cardano/Memory/Pool.hs
```diff
@@ -0,0 +1,217 @@
...
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UnboxedTuples #-}
+
+module Cardano.Memory.Pool where
```
tdammers, line 11: Could use a little bit more documentation: a summary of what this module is about and how the lock-free algorithm works, and some explanations / summaries on at least the functions and types that are intended as the API for this module (`Block`, `Pool`, `initPool`, `grabNextPoolForeignPtr`).

I would also consider explicitly exporting only the intended API, though I understand it may be necessary to expose the rest for testing purposes, and an `Internal.hs` setup isn't everyone's idea of elegance.

cardano-mempool/src/Cardano/Memory/Pool.hs
```diff
@@ -0,0 +1,217 @@
...
+ixBitSize = finiteBitSize (0 :: Word)
+
+initPool ::
+     forall n. KnownNat n
+  => Int
+  -- ^ Number of groups per page. Must be a posititve number, otherwise error. One group
+  -- contains as many blocks as the operating system has bits. A 64bit architecture will
+  -- have 64 blocks per group. For example, if program is compiled on a 64 bit OS and you
+  -- know ahead of time the maximum number of blocks that will be allocated through out
+  -- the program, then the optimal value for this argument will @maxBlockNum/64@
```
tdammers, line 61: I think it would be better to have the caller pass in a page size, rather than "groups per page" here.

"Groups", as a concept, are an implementation detail, meaningless to the API consumer. What we care about on the outside is block size and page size (i.e., allocation granularity), and it's kind of clumsy to have to reverse the math to get the correct number of groups per page to pass to this function.

cardano-mempool/README.md
```diff
@@ -0,0 +1 @@
+# primal-mempool
```
tdammers, line 1: This looks a bit spartan. Could use a little bit more information here.

Also, it looks like this is still the old project name.

tdammers (changes requested):

cardano-mempool/tests/Test/Cardano/Memory/.#PoolSpec.hs
```diff
@@ -0,0 +1 @@
+lehins@lehins-xps.30789:1667459952
```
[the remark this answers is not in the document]
lehins: Yeah, emacs temp file, I was sure I had it in my gitignore. I do now :smile:

cardano-mempool/README.md
lehins: Renamed and added a little bit of explanation what this project is about

cardano-mempool/src/Cardano/Memory/Pool.hs
lehins: Unfortunately page size is a not a good idea, because if the value supplied is not divisible by block size exactly we will end up allocating extra memory that will never be used.

Using group size gives the optimal utilization of the page, because it ensures that the page size is divisible exactly by the (number of blocks * width of atomic element in the bit array). So we utilize the page and the bit array more effectively.

That being said, we could switch to block count, rather than the group size. This will require masking off the last `Int` in the bit array, but it is doable. It would also provide a finer granularity of the page allocator. Let me know if you'd like me to switch to block count.

cardano-mempool/tests/Common.hs
lehins: Not a problem, will switch to tasty

cardano-mempool/tests/Test/Cardano/Memory/PoolSpec.hs
lehins: Totally agree. I'll try to come up with a few tests like that

cardano-mempool/src/Cardano/Memory/Pool.hs
lehins: :100:

lehins: Done. No need for internal module, since internal functions aren't testable in isolation

cardano-mempool/tests/Common.hs
lehins: Converted to `tasty`, despite that `hspec` is way better :wink:

cardano-mempool/tests/Test/Cardano/Memory/PoolSpec.hs
lehins: Done. See new `propPoolAllocateAndFinalize` property

lehins:
@tdammers This PR is ready. I've addressed all comments, except one, which I need some feedback on: namely the segment count vs bytes.

Please resolve the ones that you agree are done, or comment otherwise.

cardano-mempool/src/Cardano/Memory/Pool.hs
tdammers: Ah, I see where the confusion stems from.

In our use case, any leftover memory in an OS page that we don't use is going to be "wasted" anyway, because mlock() can only operate at OS page granularity, and our limiting factor is not overall memory use, but mlocked memory. So I was basically operating under the wost-case assumption that each individually mlocked allocation would eat up a whole page of mlocked RAM - this doesn't always happen, but depending on what else is allocating in between, it is pretty likely that most of the things that we need to mlock will occupy a full page.

But I can see how that would be wasteful in a more typical application of a mempool.

In this light, I think we should just keep the code as is, leak the "number of groups" implementation detail, and do the mlock-usecase-specific calculation on the user side, as I have it now.

I don't think it's changing the pool implementation to support arbitrary group sizes is worth it.

cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
```diff
@@ -0,0 +1,174 @@
...
+import System.Random.Stateful
+import Test.QuickCheck.Monadic
+
+poolTests :: TestTree
+poolTests = do
+  testGroup
+    "Pool"
+    [ testProperty "findNextZeroIndex" $ propFindNextZeroIndex,
+      poolProps (Block :: Block 32),
+      poolProps (Block :: Block 64)
```
tdammers, line 34: We're not going to use this, but might be a good idea to test some non-power-of-2 block sizes.

Would be even better if we could let quickcheck pick them, but I don't think this is possible with the type-safe block sizes.

tdammers (approved):
Looks great, thanks.

cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
lehins: > I don't think this is possible with the type-safe block sizes.

 Sure it is possible :wink: 

> Would be even better if we could let quickcheck pick them

Done

PR #342, opened 2022-11-14
Log mlocked pool allocations for debugging purposes

tdammers:
This is only to trigger a CI run, do not merge.


tdammers:
This will not be needed anymore.

PR #343, opened 2022-11-14
Remove the chance of false positive test failures in mempool

lehins:
Make the test case more resilient to failures by waiting until GC executes all finalizers for a full second, while checking if all ForeignPtrs got GCed every 10ms. Also speeds up the test suite.

### DIFF
--- a/cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
+++ b/cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
@@ -4,7 +4,6 @@
 
 module Test.Cardano.Memory.PoolTests (poolTests) where
 
-
 import Cardano.Memory.Pool
 import Common
 import Control.Concurrent (threadDelay)
@@ -71,8 +70,8 @@ propFindNextZeroIndex w = monadicIO . run $
             ("Expected found index to be different, but got same: " ++ show ix)
             (ix' /= ix)
           assertBool
-            ("Expected the bit under index: " ++ show ix' ++ " to not be set")
-            $ not (testBit w ix')
+            ("Expected the bit under index: " ++ show ix' ++ " to not be set") $
+            not (testBit w ix')
 
 -- We allow one extra page be allocated due to concurrency false positives in block
 -- reservations
@@ -89,9 +88,12 @@ checkNumPages pool n numBlocks = do
        ])
     (numPages <= estimatedUpperBoundOfPages)
 
-
 checkBlockBytes ::
-  (KnownNat n, Storable a, Eq a, Show a) => Block n -> a -> Ptr b -> Assertion
+     (KnownNat n, Storable a, Eq a, Show a)
+  => Block n
+  -> a
+  -> Ptr b
+  -> Assertion
 checkBlockBytes block byte ptr =
   let checkFillByte i =
         when (i >= 0) $ do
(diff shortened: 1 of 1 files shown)


cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
```diff
@@ -106,85 +108,107 @@ mallocPreFilled preFillByte bc = do
   withForeignPtr mfp $ \ptr -> setPtr (castPtr ptr) bc preFillByte
   pure mfp
 
+-- | Wait for all items to be GCed by periodicaly trigger garbage collector for certain
+-- number of times before failing, unless the counter action was executed the exact number
+-- of times.
```
tdammers, line 113: I would consider rewriting this as 2 separate functions: `ensureAllGCedWith`, which takes number of iterations and delay between iterations as arguments, and then `ensureAllGCed` defined in terms of that. This would make these two parameters explicit, instead of burying them in the body of the function and mentioning them in the Haddock as "periodically" and "a certain number of times". IMO this would be better than just documenting the values, because `ensureAllGCed` would be very small (just 1 line, really), reducing the odds of "documentation rot".

I would also try to come up with more descriptive names for `n` and `f`.

So something like this:

```haskell
-- | @ensureAllGCedWith iterations delay expectedCount registerCounter@ waits
-- for all items to be GCed by triggering garbage collection @iterations@
-- times, once every @delay@ milliseconds. After @iterations@ attempts, if the
-- counter hook was not called exactly @expectedCount@ times, a test failure is
-- raised via 'assertFailure'. Garbage collection is tracked via finalizers
-- (see below).
ensureAllGCedWith ::
       Int
       -- ^ Number of GC attempts to make before failing
    -> Integer
       -- ^ Delay between attempts, in milliseconds
    -> Int
       -- ^ Expected number of counter hook firings (in practice: individual
       -- garbage collections on 'ForeignPtr's, as per their finalizers).
    -> (IO () -> IO a)
       -- ^ Function for registering the counter hook. The argument to this
       -- function should be attached to each 'ForeignPtr' we're interested in
       -- as a finalizer.
    -> IO a
ensureAllGCedWith iterations delay expectedCount registerCounter = do
  -- existing impl here

-- | 'ensureAllGCedWith' with default values: 100 iterations, 10ms delay.
ensureAllGCed :: Int -> (IO () -> IO a) -> IO a
ensureAllGCed = ensureAllGCedWith 100 10
```

cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
```diff
@@ -106,85 +108,107 @@ mallocPreFilled preFillByte bc = do
...
+            setPtr (castPtr ptr) (blockByteCount block) emptyByte
+            countOneBlockGCed
+      -- allocate and finalize blocks concurrently
+        pool <$
+          concurrently_
+            (do replicateConcurrently_ numBlocks $ do
+                  fp <- grabNextBlock pool
+                  withForeignPtr fp (checkBlockBytes block emptyByte)
+                  writeChan chan (Just fp)
+              -- place Nothing to indicate we are done allocating blocks
```
tdammers, line 201: Indentation looks off

cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
```diff
@@ -106,85 +108,107 @@ mallocPreFilled preFillByte bc = do
...
+                writeChan chan Nothing)
+            (fix $ \loop -> do
+               mfp <- readChan chan
+               forM_ mfp $ \fp -> do
+                 withForeignPtr fp $ \ptr
+                  -- fill the newly allocated block
+                  -> setPtr (castPtr ptr) (blockByteCount block) fullByte
+                -- manually finalize every other block and let the GC to pick the rest
```
tdammers, line 209: Indentation

tdammers (changes requested):
Ergonomics nitpicks, looks good otherwise.

cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
lehins: I am not sure I agree with all these comments. This is a helper function that is used within this test module only. Haddock will never be generated for this function and is here merely for the purpose of code comment, rather than documentation. Moreover I don't like having the size of comments to be more than the size of actual code. Anyone reading this function should just look at the code rather than reading lengthy documentation to see what that function is doing.

Adding an extra function adds an unnecessary layer of indirection and indicates that it can be used on its own as well, which is not the case here.

If this was an exported function that was part of the library, I would have agreed more with most of your comments.

Despite all that I will change it to your suggested version, just because with test suites I care more about correctness, rather than readability.

cardano-mempool/tests/Test/Cardano/Memory/PoolTests.hs
lehins: Thank you. The hindent automatic formatter I use is pretty buggy and can't handle comments properly.

tdammers (approved):

PR #344, opened 2022-11-21
Add conversion Praos to PraosBatchCompat

### LINKED ISSUE #345: Create a conversion function from Praos to PraosBatchCompat types
Some of the new BatchCompat VRF types are compatible with the existing Praos types. In this PR we create conversion functions, to allow for a smooth transition between the compatible types.

In particular, we need the following conversion functions:

* `skToBatchCompat :: SignKeyVRF PraosVRF -> SignKeyVRF BC.PraosBatchCompatVRF`
* `vkToBatchCompat :: VerKeyVRF PraosVRF -> VerKeyVRF BC.PraosBatchCompatVRF`
* `outputToBatchCompat :: OutputVRF PraosVRF -> OutputVRF BC.PraosBatchCompatVRF`

We must include tests to guarantee the compatibility of these converted types.

iquerejeta:
Some of the new `BatchCompat` VRF types are compatible with the existing `Praos` types. In this PR we create conversion functions, to allow for a smooth transition between the compatible types. 

In particular, we create the following conversion functions: 
* `skToBatchCompat :: SignKeyVRF PraosVRF -> SignKeyVRF BC.PraosBatchCompatVRF`
* `vkToBatchCompat :: VerKeyVRF PraosVRF -> VerKeyVRF BC.PraosBatchCompatVRF`
* `outputToBatchCompat :: OutputVRF PraosVRF -> OutputVRF BC.PraosBatchCompatVRF`

We include tests to guarantee the compatibility of these converted types. 

This closes #345


cardano-crypto-praos/src/Cardano/Crypto/VRF/Praos.hs
```diff
@@ -418,6 +423,26 @@ prove sk msg =
             0 -> return $ Just $! proof
             _ -> return Nothing
 
+-- | Construct a BatchCompat vkey from praos, non-batchcompat
+vkToBatchCompat :: VerKeyVRF PraosVRF -> VerKeyVRF BC.PraosBatchCompatVRF
+vkToBatchCompat praosVk =
+  case rawDeserialiseVerKeyVRF (rawSerialiseVerKeyVRF praosVk) of
+    Just vk -> vk
+    Nothing -> error "All valid Praos VKs are valid BatchCompat VKs."
```
tdammers, line 431: Wouldn't it make more sense to word this the other way around? That is, the error condition is that the given VK is *not* a valid batch-compatible VK for some reason. I.o.w., the error message should state the error condition, not the unmet assertion. (This is unlike property check names, where we do name property check functions after the desired property.)

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -26,10 +26,15 @@ import Test.QuickCheck
          ((==>), (===), Arbitrary(..), Gen, Property,  NonNegative(..),
           counterexample)
 import Test.Tasty (TestTree, testGroup)
-import Test.Tasty.QuickCheck (testProperty)
+import Test.Tasty.QuickCheck (testProperty, vectorOf)
 
 {- HLINT IGNORE "Use <$>" -}
 
+--
+-- New type to generate random ByteStrings of size 32
```
tdammers, line 34: I'd spell it as "newtype" rather than "new type" (we specifically need this to be a Haskell `newtype` construct, not just a new type in the more general sense).

tdammers, line 34: Suggest making this into a proper Haddock comment (`-- | (comment)` instead of just `-- (comment)`), so that it shows up in generated documentation.

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -208,6 +222,84 @@ prop_naturalToBytes (NonNegative sz) n =
     sz >= 8 ==>
       bytesToNatural (naturalToBytes sz (fromIntegral n)) == fromIntegral n
 
+--
+-- Praos <-> BatchCompatPraos VerKey conversion
+--
+prop_pubKeyToBatchComopat :: VerKeyVRF PraosVRF -> Bool
+prop_pubKeyToBatchComopat vk =
+  rawSerialiseVerKeyVRF (vkToBatchCompat vk) == rawSerialiseVerKeyVRF vk
```
tdammers, line 230: Consider changing the return type to `Property`, and using `===` instead of `==`. This will produce better reporting of test failures, listing not only the name of the test and the inputs, but also the exact operands to the failed comparison.

tdammers (changes requested):

lehins (approved):
Looks good. A couple of minor suggestions

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -230,3 +321,11 @@ instance (VRFAlgorithm v,
...
     return $ snd $ evalVRF () a sk
   shrink = const []
+
+instance VRFAlgorithm v => Arbitrary (OutputVRF v) where
+  arbitrary = do
+    bytes <- BS.pack <$> vectorOf (fromIntegral (sizeOutputVRF (Proxy :: Proxy v))) arbitrary
+    return $ OutputVRF bytes
+
+instance Arbitrary BS32 where
+    arbitrary = BS32 . BS.pack <$> vectorOf 32 arbitrary
```
tdammers, line 331: `vectorOf 32 arbitrary` == `vector 32`

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -25,10 +26,13 @@ import Test.QuickCheck
...
           counterexample)
 import Test.Tasty (TestTree, testGroup)
-import Test.Tasty.QuickCheck (testProperty)
+import Test.Tasty.QuickCheck (testProperty, vectorOf)
 
 {- HLINT IGNORE "Use <$>" -}
 
+-- | Newtype to generate random ByteStrings of size 32
+newtype BS32 = BS32 { unBS32 :: BS.ByteString } deriving (Show)
+
```
tdammers, line 35: Instead of newtyping `ByteString`, you can also newtype `Seed`, and use the existing `arbitrarySeedOfSize` function to define it. You could even generalize this with a type-level Nat, like so:

```haskell
newtype SizedSeed (n :: Nat) = SizedSeed { unSizedSeed :: Seed } deriving Show

instance (KnownNat n) => Arbitrary (SizedSeed n) where
    arbitrary = SizedSeed <$> arbitrarySeedOfSize (fromIntegral $ natVal (Proxy @n))
```

And then instead of having `BS32` arguments and converting them to seeds when needed, you would use `SizedSeed 32` and just `unSizedSeed` (or `coerce`, because they're newtypes) them.

For this PR, this doesn't really matter much, but in the interest of avoiding code duplication, I think having the `SizedSeed` generalization in Test.Crypto.Util would be good.

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -44,6 +48,15 @@ tests =
...
       , testProperty "naturalToBytes" prop_naturalToBytes
       ]
+    , testGroup "ConvertingTypes"
+      [ testProperty "pubKeyToBatchCompat" prop_pubKeyToBatchComopat
+      , testProperty "signKeyToBatchCompat" prop_signKeyToBatchCompat
+      , testProperty "outputToBatchCompat" prop_outputToBatchComat
+      , testProperty "compatibleVerKeyConversion" prop_verKeyValidConversion
+      , testProperty "compatibleOutputConversion" prop_outputValidConversion
+      , testProperty "compatibleSignKeyConversion" prop_signKeyValidConversion
+      , testProperty "compatibleFullConversion" prop_fullValidConversion
```
tdammers, line 58: Suggest additional tests:

- Generate keys for both Praos and BatchCompat from the same seed; sign a message with Praos, verify with BatchCompat.
- Same, but the other way around.

We've already tested that all the keys and outputs are the same, but I think it would be good to make this compatibility explicit too.

tdammers (approved):
Looks fine by me, none of the remarks are blockers.

cardano-crypto-tests/src/Test/Crypto/VRF.hs
iquerejeta: The problem is that the signatures themselves are not compatible across types. Its only the keys (and outputs).

cardano-crypto-tests/src/Test/Crypto/VRF.hs
iquerejeta: I've added the SizedSeed at type level as you suggested.

tdammers (changes requested):
Looks good to me, except for those parentheses.

cardano-crypto-tests/src/Test/Crypto/VRF.hs
tdammers: Oh, I see. In that case you're right, that test wouldn't make sense.

lehins (approved):
LGTM

tdammers (approved):

PR #348, opened 2022-11-25
Minor improvement to memory pool allocator

lehins:
We don't want to rely on GC to cleanup redundantly allocated pages that use scarce memory, thus it is better to clean it up eagerly

### DIFF
--- a/cardano-mempool/src/Cardano/Memory/Pool.hs
+++ b/cardano-mempool/src/Cardano/Memory/Pool.hs
@@ -40,7 +40,6 @@ import Foreign.ForeignPtr
 import GHC.ForeignPtr
 import Control.Applicative
 import Data.Bits
-import Data.Maybe
 import GHC.TypeLits
 import Data.Primitive.PrimArray
 import Data.Primitive.PVar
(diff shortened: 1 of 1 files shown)


tdammers (changes requested):
Are there any tests that verify this behavior?

lehins:
@tdammers This is not a user facing feature, which cannot be tested. In other words this is a mere implementation detail that has no impact whatsoever on the semantic behavior of this library, so I do not think designing some sort of test is relevant.

tdammers:
I'd say that "not running out of mlocked memory in a long-running process" *is* a user-facing "feature", after all it can make the difference between code working and, well, not working. 

It's probably not super relevant to our use case, because the expectation is that once we have allocated all the keys we need, memory use will remain stable, and deallocating unused pages isn't going to make much of a difference, but if we were to use it for other purposes, where pooled memory is used in infrequent bursts, cleaning up unused pages would be absolutely vital.

Can we maybe come up with a simple test that allocates X blocks (enough to fill a couple pages), releases them again, and then verifies that the number of pages in the pool is down to 1 again? That shouldn't be too difficult, and I'd feel better knowing that this actually happens.

lehins:
> Can we maybe come up with a simple test that allocates X blocks (enough to fill a couple pages), releases them again, and then verifies that the number of pages in the pool is down to 1 again? 

There is no deallocation of pages in current implementation. Once a page has been allocated it will be there until the memory `Pool` is garbage collected. This was never an intended feature. From what I gathered C implementation did not have such feature either, right? If we are to implement Page cleanup we are essentially rolling out our own mini garbage collector. I am not saying it is impossible to do it, but if we are to add this feature the implementation will become much more complex.

This PR merely finalizes a page that was allocated concurrently with another one and therefore will not be used. So, instead of waiting for garbage collector to release the memory for a page that can't be used we proactively cleaning it up.

tdammers:
Oh, I see, I misunderstood the purpose of this then. In that case, yeah, let's just keep it this way.

tdammers (approved):

PR #349, opened 2022-11-30
mlock violations flag

tdammers:
This adds a build flag, `allow-mlock-violations`, which controls the behavior of some operations on mlocked memory.

When `allow-mlock-violations` is set, secrets stored in mlocked memory can be leaked to the Haskell heap through `mlsbToByteString`, `traceMLSB`, `traceMLockedForeignPtr`, and the `Show` instance for `MLockedSizedBytes`; these operations create various representations of mlocked memory in regular heap memory.

When `allow-mlock-violations` is cleared, these operations are implemented in such a way that no secrets can be leaked. Specifically:

- `show @MLockedSizedBytes` will produce a generic string (`"MLSB <%i bytes>"`, where `%i` is the type-level size of the MLSB)
- `traceMLSB` and `traceMLockedForeignPtr` will produce similar generic strings
- `mlsbToByteString` will fail with a runtime error; attempts at compiling code that uses `mlsbToByteString` will also trigger a compiler warning.

Currently, both cardano-node and the test suite depend on such unsafe operations (cardano-node for sign key ser/deser, the test suite for comparing and inspecting values stored in mlocked memory), so we are defaulting the flag to `True`; however, dependants that do not need the unsafe operations can disable it, and both cardano-node and the test suite should be modified to not depend on them either.


tdammers:
Will not do; KES operations that may leak secrets to the GHC heap now have their own typeclass, `UnsoundKESAlgorithm`, which should hopefully be enough of a warning signal.

PR #350, opened 2022-12-01
KES forgetting generalized

tdammers:
Prior to this, `Ed25519ML` only had a `DSIGNMAlgorithm` instance for `IO`.  This was the last bit of code that prevented us from instantiating the various KES algorithms for monads other than `IO`, including `IOSim` and monad stacks built on top of either.

With this change, `Ed25519ML` has generalized `DSIGNMAlgorithm` instances, constrained only on:

- `MonadSodium`, providing access to mlocked memory
- `MonadST`
- `MonadThrow`

Hence it remains compatible with `IO` out of the box, and can be made compatible with `IOSim` by implementing an `instance MonadSodium IOSim`.  Since we don't use `IOSim` to handle actual production secrets, we can choose to mock out the mlocking there, or simulate mlocking constraints, or trace mlocked allocations at the `IOSim` level, which would allow us to get rid of the somewhat awkward ad-hoc tracing we have now.

This depends on #255, and should not be merged before it.


tdammers:
These changes have been merged into tdammers/kes-agent-requirements; we will not need this PR anymore.

PR #356, opened 2023-01-01
Use setup-haskell action

newhoggy:


.github/workflows/haskell.yml
```diff
@@ -37,6 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         ghc: ["8.10.7", "9.2.4"]
+        cabal: ["3.6.2.0"]
```
lehins, line 40: Is there a reason we should be running our CI with different versions of cabal? I think this setting should go directly into `haskell/actions`

.github/workflows/haskell.yml
```diff
@@ -49,84 +50,22 @@ jobs:
...
-          echo "cabal-store=$(dirname "$cabal_config_file")/store" | tee -a "$GITHUB_OUTPUT"
-        else
-          echo "cabal-store=C:\\cabal\\store" | tee -a "$GITHUB_OUTPUT"
-        fi
+    - name: Install Haskell
+      uses: input-output-hk/setup-haskell@v1
+      id: setup-haskell
+      with:
+        ghc-version: ${{ matrix.ghc }}
+        cabal-version: ${{ matrix.cabal }}
```
lehins, line 59: It is safe to hard code it here, instead of using a variable. Also using the latest version makes more sense, unless there is some reason we are forced to use 3.6.2?
```suggestion
        cabal-version: "3.8.1.0"
```

In fact, I wouldn't even oppose removing the setting and just using the latest cabal. I think it makes sense to see early if our tooling breaks upon a cabal upgrade

lehins (dismissed):
I like the overall simplification, but the cabal version addition changes the name of the action which affects the requirement on branch protection rules. For this reason we should get rid of cabal version in the matrix

.github/workflows/haskell.yml
newhoggy: Done!

.github/workflows/haskell.yml
newhoggy: Using `cabal-version: latest`

lehins:
@newhoggy Windows CI now fails for some strange reason

lehins (approved):
Beautiful

PR #360, opened 2023-01-19
Fix index-state syntax in cabal.project

andreabedini:
The index-state stanza in cabal.project is "last wins", so 
```
index-state: 2022-12-24T00:00:00Z
index-state: cardano-haskell-packages 2022-11-02T15:34:17Z
```
would set the index state of hackage to HEAD.

The correct syntax to set the index-state of both repositories is
```
index-state:
  , hackage.haskell.org 2022-12-24T00:00:00Z
  , cardano-haskell-packages 2022-11-02T15:34:17Z
```
haskell.nix has trouble parsing the index-state syntax with multiple repositories, so we repeat the index-state for Hackage. Cabal will ignore the first index-state stanza anyway.


lehins (approved):
For the record this duplicate index-state listing is needed due to this cabal issue: https://github.com/haskell/cabal/issues/8568

PR #361, opened 2023-01-19
Improve `cardano-binary` and account for migration to `cardano-ledger-binary`

lehins:
[`cardano-ledger-binary`](https://github.com/input-output-hk/cardano-ledger/blob/master/libs/cardano-ledger-binary) now implements versioned serialization that is suitable for `cardano-ledger` and possibly some users of thereof. For that reason `cardano-binary` will be mostly used by tools and libraries outside of `cardano-ledger` and can be simplified a bit. 

* Removed `Cardano.Binary.Annotated` and `Cardano.Binary.Drop` modules. They have been replaced by equivalent in [`cardano-ledger-binary`](https://github.com/input-output-hk/cardano-ledger/blob/master/libs/cardano-ledger-binary)
* Removed `Cardano.Binary.Raw`. It has moved into:  [`cardano-crypto-wrapper:Cardano.Crypto.Raw`](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/byron/crypto/src/Cardano/Crypto/Raw.hs)

This PR also includes various minor improvements as well, which are needed for `cardano-ledger-binary` as well, since it reuses some parts of `cardano-binary`:

* Generalized `cborError` to `MonadFail`
* Add `ToCBOR` instance for `Tokens -> Tokens`
* Add `To/FromCBOR` instances for `Term` and `ToCBOR` for `Encoding`
* Add `To/FromCBOR` instances for 6-tuples and 8-tuples
* Remove `FromCBOR` instance for `Ratio` in favor of `Rational`.
* Add `To/FromCBOR` instances for `Double`.
* Rename `toCBORMaybe` -> `encodeMaybe` with deprecation.
* Rename `decCBORMaybe` -> `decodeMaybe` with deprecation.
* Add `encodeNullMaybe` and `decodeNullMaybe`.
* Add `ToFromCBOR` instances for `Seq` and `StrictSeq`
* Deprecate `serializeEncoding` and `serializeEncoding'` in favor of `serialize` and
  `serialize'` respectfully, since `Encoding` now has the `ToCBOR` instance.
* Add `decodeFullDecoder'` that accepts strict `ByteString`.


binary/src/Cardano/Binary/FromCBOR.hs
```diff
@@ -577,9 +577,9 @@ instance FromCBOR UTCTime where
       (picosecondsToDiffTime timeOfDayPico)
 
 -- | Convert an 'Either'-encoded failure to a 'cborg' decoder failure
```
teodanciu, line 579: `Convert an 'Either'-encoded failure to a MonadFail`  ?

binary/src/Cardano/Binary/FromCBOR.hs
```diff
@@ -577,9 +577,9 @@ instance FromCBOR UTCTime where
...
-toCborError :: B.Buildable e => Either e a -> D.Decoder s a
+toCborError :: (MonadFail m, B.Buildable e) => Either e a -> m a
 toCborError = either cborError pure
 
 -- | Convert a @Buildable@ error into a 'cborg' decoder error
```
teodanciu, line 583: Same as above, the comment looks outdated

binary/CHANGELOG.md
```diff
@@ -17,3 +17,5 @@
 * Rename `decCBORMaybe` -> `decodeMaybe` with deprecation.
 * Add `encodeNullMaybe` and `decodeNullMaybe`.
 * Add `ToFromCBOR` instances for `Seq` and `StrictSeq`
+* Deprecate `serializeEncoding` and `serializeEncoding'` in favor of `serialize` and
+  `serialize'` respectfully, since `Encoding` now has the `ToCBOR` instance.
```
teodanciu, line 21: I think `respectively`  (but I love `respectfully` - they deserve it!)

binary/src/Cardano/Binary/Serialize.hs
```diff
@@ -60,17 +66,13 @@ serializeBuilder = CBOR.Write.toBuilder . toCBOR
...
 serializeEncoding :: Encoding -> BSL.ByteString
-serializeEncoding =
-  Builder.toLazyByteStringWith strategy mempty . CBOR.Write.toBuilder
-  where
-    -- 1024 is the size of the first buffer, 4096 is the size of subsequent
-    -- buffers. Chosen because they seem to give good performance. They are not
-    -- sacred.
-        strategy = Builder.safeStrategy 1024 4096
+serializeEncoding = serialize
+{-# DEPRECATED serializeEncoding "Use `serialize` instead, since `Encoding` has `EncCBOR` instance" #-}
```
teodanciu, line 70: `EncCBOR` or `ToCBOR` ?

binary/src/Cardano/Binary/ToCBOR.hs
```diff
@@ -648,11 +650,22 @@ variableListLenEncoding len contents =
...
   Just x  -> E.encodeListLen 1 <> encodeA x
 
+toCBORMaybe :: (a -> Encoding) -> Maybe a -> Encoding
+toCBORMaybe = encodeMaybe
+{-# DEPRECATED toCBORMaybe "In favor of `encodeMaybe`" #-}
+
+-- | Alternative way to encode a Maybe type.
+--
+-- /Note/ - this is not the default method for encoding `Maybe`, use `encodeMaybe` instead
+encodeNullMaybe :: (a -> Encoding) -> Maybe a -> Encoding
```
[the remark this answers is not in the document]
lehins: Keeping the name consistent with the one from `cardano-ledger-binary`: 
https://github.com/input-output-hk/cardano-ledger/blob/b533fd446ce210019c8d201ad90bb75c317625a1/libs/cardano-ledger-binary/src/Cardano/Ledger/Binary/Encoding/Encoder.hs#L316-L319

Regardless of this, I am not quite following how `encodeNullMaybe` is less discoverable than `encodeMaybeNull`?

binary/src/Cardano/Binary/Serialize.hs
```diff
@@ -60,17 +66,13 @@ serializeBuilder = CBOR.Write.toBuilder . toCBOR
...
 
 -- | A strict version of 'serializeEncoding'
 serializeEncoding' :: Encoding -> BS.ByteString
-serializeEncoding' = BSL.toStrict . serializeEncoding
+serializeEncoding' = serialize'
+{-# DEPRECATED serializeEncoding' "Use `serialize'` instead, since `Encoding` has `EncCBOR` instance" #-}
```
teodanciu, line 75: Same as above, seems like it should be `ToCBOR` in the deprecation message

teodanciu (approved):
Looks good to me, just some minor comments

binary/src/Cardano/Binary/FromCBOR.hs
lehins: Good catch. Fixed

binary/src/Cardano/Binary/Serialize.hs
lehins: Thank you. Another good catch. This snuck in due to some other refactoring/renaming that is underway.

binary/src/Cardano/Binary/Deserialize.hs
```diff
@@ -81,6 +82,17 @@ decodeFullDecoder lbl decoder bs0 = case deserialiseDecoder decoder bs0 of
     else Left $ DecoderErrorLeftover lbl leftover
   Left (e, _) -> Left $ DecoderErrorDeserialiseFailure lbl e
 
+decodeFullDecoder'
```
JaredCorduan, line 85: Does it pain you to have to propagate the ticks, or do you like it when the meaning is about strictness?

binary/src/Cardano/Binary/FromCBOR.hs
```diff
@@ -191,22 +197,25 @@ instance FromCBOR Word64 where
...
-instance (Integral a, FromCBOR a) => FromCBOR (Ratio a) where
+instance FromCBOR Float where
+  fromCBOR = D.decodeFloat
+
+instance FromCBOR Double where
+  fromCBOR = D.decodeDouble
+
+instance FromCBOR Rational where
   fromCBOR = do
     enforceSize "Ratio" 2
```
JaredCorduan, line 214:
```suggestion
    enforceSize "Rational" 2
```

binary/src/Cardano/Binary/FromCBOR.hs
```diff
@@ -476,10 +537,33 @@ decodeVector = decodeContainerSkelWithReplicate
...
 {-# INLINE decodeVector #-}
 
-instance (FromCBOR a) => FromCBOR (Vector.Vector a) where
+instance FromCBOR a => FromCBOR (Vector.Vector a) where
   fromCBOR = decodeVector
   {-# INLINE fromCBOR #-}
 
+instance FromCBOR a => FromCBOR (Seq.Seq a) where
+  fromCBOR = decodeSeq fromCBOR
+  {-# INLINE fromCBOR #-}
```
JaredCorduan, line 546: Sometime I would love to learn from you about how you decide when to add the inline-ing.

binary/src/Cardano/Binary/ToCBOR.hs
```diff
@@ -404,6 +409,15 @@ withWordSize x =
...
 
+instance ToCBOR Encoding where
+  toCBOR = id
+
+instance ToCBOR (Tokens -> Tokens) where
+  toCBOR = Encoding
+
+instance ToCBOR Term where
+  toCBOR = encodeTerm
+
```
JaredCorduan, line 420: oh fun! this looks really helpful

binary/src/Cardano/Binary/FromCBOR.hs
lehins: I add them, when I don't forget it. :smile: 

Small functions like that it is almost always safe to add an `INLINE`, larger functions however would be better with `INLINEABLE`. Occasionally it is better to use `NOINLINE` if you can see that inlining would result in duplicated work.

Most of the time it's a heuristic and therefore is hard to explain in concrete terms.

binary/src/Cardano/Binary/ToCBOR.hs
lehins: It actually is, I'll show later how it can be very nicely leveraged.

PR #376, opened 2023-03-16
Remove `development` flag. Add json instances

### LINKED ISSUE #372: Please don't impose Werror on downstream users
`Werror` is great for development and CI, but is not appropriate for users of your library. New compiler versions frequently bring out new warnings or refine their heuristics. When this breaks the build, all downstream can reasonably do is blindly disable Werror anyway.

Fresh project using `plutus-tx` (and thus, indirectly, `cardano-crypto-class`) fails with GHC 9.4.6 with:

```
[ 6 of 48] Compiling Cardano.Crypto.Hash.Class ( src/Cardano/Crypto/Hash/Class.hs, dist/build/Cardano/Crypto/Hash/Clas>

src/Cardano/Crypto/Hash/Class.hs:173:38: error: [-Wdeprecations, -Werror=deprecations]
    In the use of ‘serializeEncoding'’
    (imported from Cardano.Binary, but defined in cardano-binary-1.6.0.0-9zN4iiyMBUGJoTfCIR8U3A:Cardano.Binary.Seriali>
    Deprecated: "Use `serialize'` instead, since `Encoding` has `ToCBOR` instance"
    |
173 | hashWithSerialiser toEnc = hashWith (serializeEncoding' . toEnc)
    |                                      ^^^^^^^^^^^^^^^^^^
```

lehins:
Move a few JSON instances from `cardano-node`

Remove `development` cabal flag:  Fixes #372

### DIFF
--- a/base-deriving-via/CHANGELOG.md
+++ b/base-deriving-via/CHANGELOG.md
@@ -1,4 +1,8 @@
 ## 0.1.0.1
 
+* Remove `development` flag: #372
+
+## 0.1.0.1
+
 * Initial release
 
--- a/base-deriving-via/base-deriving-via.cabal
+++ b/base-deriving-via/base-deriving-via.cabal
@@ -1,7 +1,7 @@
 cabal-version:       >=1.10
 
 name:                base-deriving-via
-version:             0.1.0.1
+version:             0.1.0.2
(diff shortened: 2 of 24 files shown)


JaredCorduan (approved):
LGTM!

PR #377, opened 2023-03-20
DO NOT MERGE: make MLockedSizedBytes over a monad

tdammers:
This is an attempt at making it possible to mock out MLockedSizedBytes, and particularly the mlocking part, in `IOSim`; the idea is to make MLockedForeignPtr an associated type of MonadMLock, which would allow us to stub out

However, we cannot easily express `withMLSB :: (Ptr a -> m b) -> MLockedSizedBytes a -> m b` for `m ~ IOSim s`, because that would require lifting actions between `IOSim s` and `ST s` both ways, without losing the state associated with `IOSim`. This cannot be done using `withForeignPtr`, nor any other obvious pointer type, so instead we'll try a different route for the time being. In the meantime, we'll park the solution so far on this branch.


tdammers:
No longer relevant; mlocking can now happen in any `MonadST`, including `IOSim`.

PR #379, opened 2023-03-22
Newhoggy/ghc 9.2.x

newhoggy:


lehins:
@newhoggy Thank you for investigating this! I think we can close this PR now since the issue has been fixed in 3e6c0761350c66df95967779ff4525c504362d5f

PR #386, opened 2023-04-18
Nix based GHAs for easy reproducibility and cross compilation.

angerman:
This reuses our haskell.nix infrastructure through the input-output-hk/devx repository to build GHAs that effectively run
```
nix develop github:input-output-hk/devx#ghc<ver>-minimal-iog
```
(as well as JavaScript and Windows cross).

This is distinctly different from running the GHA using input-outpu-hk/actions/base + haskell, in that they use linux and macOS builder only (windows is cross compiled on linux). Re-using the same infrastructure we use to build full nix builds as well, however, building the application in a _shell_. Thus you effectively use nix to _provision_ the environment, while using usual tools (cabal, ghc, ...) to actually _build_ the software.

This also means, you can trivially debug the situation in CI, by simply sparing the same local shell on you linux (and macOS) machine. It also means we should soon be able to debug windows cross (and javascript cross) topics fairly easily.


angerman:
One thing, which the PR still lacks, is windows cross -> test artifacts on windows. Using upload/download artifacts.

But this should give a good first impression of the actions/devx workflow.

noelportillo (approved):
Keep your eyes open

angerman:
Closing — long stale draft.

PR #387, opened 2023-04-19
Haddocks on Github Pages

teodanciu:
This PR introduces a new github workflow, triggered on master, which:

* builds haddock for every package, aggregating them as a site (listed on the landing page)
* deploys this site to github pages, so the haddocks will be accessible here, after the PR is merged: https://input-output-hk.github.io/cardano-base

This is a copy of the site with this PR on my own fork, to demonstrate how the pages will look like: https://bienpulenta.github.io/cardano-base

### DIFF
--- a/scripts/haddocks.sh
+++ b/scripts/haddocks.sh
@@ -1,0 +1,110 @@
+#!/usr/bin/env bash
+
+# Build haddock documentation and an index for all projects in
+# `cardano-base` repository.
+#
+# usage:
+# ./haddocks.sh directory [true|false]
+#
+# $1 - where to put the generated pages, this directory contents will be wiped
+#      out (so don't pass `/` or `./` - the latter will delete your 'dist-newstyle')
+#      (the default is './haddocks')
+# $2 - whether to re-build haddocks with `cabal haddock` command or a component name
+#      (the default is true)
+#
+
+set -euo pipefail
+
+OUTPUT_DIR=${1:-"./haddocks"}
+REGENERATE=${2:-"true"}
+
+BUILD_DIR="dist-newstyle"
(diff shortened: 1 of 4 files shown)


scripts/haddocks.sh
```diff
@@ -0,0 +1,113 @@
...
+    --haddock-quickjump
+    --haddock-hyperlink-source
+    --haddock-option "--show-all"
+    --haddock-option "--use-unicode"
+    --haddock-option="--base-url=.."
+  )
+
+# build documentation of all modules
+if [ ${REGENERATE} == "true" ]; then
+  cabal haddock "${HADDOCK_OPTS[@]}" cardano-binary-test
```
lehins, line 41: Is `cardano-binary-test` special in some way? Shouldn't it be:
```suggestion
  cabal haddock "${HADDOCK_OPTS[@]}" all
```

lehins (approved):
Some nice bashscripting :slightly_smiling_face: 

Great work, thank you!

scripts/haddocks.sh
teodanciu: I left this by accident after testing something!! Thank you for spotting it

scripts/haddocks.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
```
coot, line 1: If the repo supports `ghc-9.4` you could replace this just with:
```
cabal haddock-project --hackage all
```
One needs `cabal-3.10.1.0` for that.

lehins: It might be worth switching in the future, but right now I don't want to rely on GHC-9.6, since it barely has been added to CI

PR #388, opened 2023-04-19
Refactor MonadSodium into MonadMLock etc.

tdammers:
This acknowledges the fact that by now, the MonadSodium typeclass is not really about libsodium anymore, and has become a kitchen sink of various concerns that we needed to abstract over in order to use mlocked memory and related functionality. Hence, we split up `MonadSodium` into a set of orthogonal typeclasses, each covering one of those concerns.

Specifically, the new typeclasses we introduce to replace `MonadSodium` are:

- `MonadMLock`, covering mlocked memory operations
- `MonadByteStringMemory`, covering access to raw memory in bytestrings (`Data.ByteString.useAsCString` etc.)
- `MonadPSB`, covering the `PinnedSizedBytes` API
- `MonadUnmanagedMemory`, covering memory allocations/deallocations outside of the Haskell heap (`allocaBytes` etc.)


tdammers:
Superseded by #404.

PR #389, opened 2023-04-19
SignKeyWithPeriodKES

tdammers:
This introduces a new type, `SignKeyWithPeriodKES`, which bundles a KES sign key with the KES period during which it is valid. This is largely a formality, but such a bundle is useful in that it allows us to keep KES keys associated with the correct period as they travel through the system, and have a KES evolution function (`updateKESWithPeriod`) that handles period and key together.

Depends on #388.


lehins (approved):
This is a very nice simplification. Love it!

PR #390, opened 2023-04-19
Try to depend on blst via pkg-config

michaelpj:


lehins:
I think header files need to be removed too: https://github.com/input-output-hk/cardano-base/blob/2c99deaf2bbec94379c028281235354830223d4e/cardano-crypto-class/cardano-crypto-class.cabal#L22-L23

And `blst` to be installed with pkg-config awareness on CI (no nix on GithubActions), whatever that would entail.

angerman:
> I think header files need to be removed too:
> 
> https://github.com/input-output-hk/cardano-base/blob/2c99deaf2bbec94379c028281235354830223d4e/cardano-crypto-class/cardano-crypto-class.cabal#L22-L23
> 
> And `blst` to be installed with pkg-config awareness on CI (no nix on GithubActions), whatever that would entail.

I'll try to clean all this up during the day.

.github/workflows/build-blst.bash
```diff
@@ -7,4 +7,20 @@ git clone https://github.com/supranational/blst
 cd blst
 git reset --hard $BLST_REF
 ./build.sh
+
+mkdir -p pkgconfig
```
michaelpj, line 11: separate question: should we try and upstream this to the blst repo? they might like it

flake.nix
```diff
@@ -78,13 +78,6 @@
               packages.ekg.components.library.enableSeparateDataOutput = true;
               packages.cardano-binary.configureFlags = [ "--ghc-option=-Werror" ];
               packages.cardano-crypto-class.configureFlags = [ "--ghc-option=-Werror" ];
-              # We need to override the pkgconfig libraries so that we can provide our fork of
-              # libsodium instead of the one from nixpkgs, which is what haskell.nix would
-              # otherwise choose. Unfortunately, this means that we also override any other
-              # pkgconfig libraries that haskell.nix would pick for us. So we also need to
-              # manually include those here.
-              packages.cardano-crypto-class.components.library.pkgconfig = pkgs.lib.mkForce [[ pkgs.libsodium-vrf pkgs.secp256k1 pkgs.blst ]];
-              packages.cardano-crypto-praos.components.library.pkgconfig = pkgs.lib.mkForce [[ pkgs.libsodium-vrf ]];
```
michaelpj, line 87: Surely you need the libsodium-vrf and secp256k1 bits still?

.github/workflows/build-blst.bash
michaelpj: ... are we going to have to put this in every single downstream project's GHA workflows? :(

angerman: Yea, no. This is going to suck too much. We need to have a better solution. But let's make this work first.

flake.nix
michaelpj: Because otherwise it will link against the wrong libsodium, surely? And then the tests should hopefully fail? IDK

angerman: iohk-nix will now map `libsodium` to `libsodium-vrf`; thusly, if you use nix (and iohk-nix to get our custom libraries), you should get the appropriate libsodium.

michaelpj: Oho, so we can finally bin all this stuff? That's nice.

angerman:
🤯 why did the blst-sources folder disappear in the test stage?

angerman:
dealing with this is stupid. Windows is ... fun.

We already have separate actions for secp, ... I guess we now also need one for blst. That's pretty insane, also building them from source all the time.

.github/workflows/build-blst.bash
lehins: We can later create a github action, similar to how secp256k1 is handled everywhere: https://github.com/input-output-hk/setup-secp256k1

angerman:
It's finally building after installing stuff into system global directories. 💀

flake.nix
angerman: Yes, that's the hope 🤞 Nuke all the hacks!

.github/workflows/build-blst.bash
lehins: It won't be necessary for libsodium, in fact I would discourage using custom libsodium on GA CI, because we promise our crypto libs to work with stock libsodium, so it is good that we test that version as well.

angerman:
And now it fails on linux 🤣

angerman:
@michaelpj @lehins ready for review.

michaelpj (commented):
Looks fine to me. I still think we should try and upstream the pkgconfig stuff, but whatever.

angerman:
> Looks fine to me. I still think we should try and upstream the pkgconfig stuff, but whatever.

Pretty much, pkgconfig + Makefile that supports prefix 🙄 but they seem to love their hand rolled build script 💀

angerman:
Two things we should do after this:
- [ ] one GHA to have all IOG custom libraries (would even argue that should live in iohk-nix, so you say `uses input-output-hk/iohk-nix` in GHA); and we keep it current with the libraries in iohk-nix.
- [ ] upstream the packaging improvements to blst.

angerman:
Unless anyone objects, @lehins I'll go and merge this into cardano-base for now.

lehins (approved):
I sincerely appreciate this! This whole previous approach of including a copy of header files with haskell package was a bit disgusting.

PR #392, opened 2023-04-21
Make it build with ghc-9.6.1

erikd:


.github/workflows/cabal.project.local.ci.Darwin
```diff
@@ -3,3 +3,13 @@ package cardano-crypto-praos
 
 package HsOpenSSL
   flags: +use-pkg-config
+
+allow-newer:
```
michaelpj, line 7: What we've been doing in some other places is to use the new cabal support for conditionals in project files to make it clear that these are only needed for 9.6+

Like this: https://github.com/haskell/haskell-language-server/blob/master/cabal.project#L118

Then I think you can sensibly put this in the main `cabal.project` rather than all these CI-specific ones.

erikd: Yes, that is better

newhoggy (approved):

cardano-crypto-class/src/Cardano/Crypto/KES/Simple.hs
```diff
@@ -12,6 +12,9 @@
 {-# LANGUAGE NoStarIsType #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 
+-- Needed for ghc-9.6.
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+
```
angerman, line 17: Why is this needed only for 9.6? If we remove the redundant constraints it will fail with 9.2? 8.10?

erikd: Nope. Only `ghc-9.6` needs it.

PR #393, opened 2023-04-22
add flake-compat

angerman:
This uses https://github.com/edolstra/flake-compat to provide `default.nix` and `shell.nix`. It also added some (hopefully) helpful guidance for people to switch to `nix develop` instead.


lehins (approved):
Awesome! Thank you.

PR #394, opened 2023-04-24
Improve comment on need for -Wno-redundant-constraints

erikd:


cardano-crypto-class/src/Cardano/Crypto/KES/Simple.hs
```diff
@@ -12,7 +12,9 @@
 {-# LANGUAGE NoStarIsType #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 
--- Needed for ghc-9.6.
+-- Needed for ghc-9.6 to avoid a redunant constraint warning on the
+-- `KESSignAlgorithm m (SimpleKES d t)` instance. Removing the constraint leaves another type
+-- error which is rather opaque.
 {-# OPTIONS_GHC -Wno-redundant-constraints #-}
```
angerman, line 18: @lehins maybe something you can take a look at when you are having a free minute?

.github/workflows/build-blst.bash
```diff
@@ -10,13 +10,13 @@ git reset --hard $BLST_REF
...
 mkdir -p ${PREFIX}/include/blst
-cp bindings/{blst.h,blst_aux.h} ${PREFIX}/include/blst/ 
+cp bindings/{blst.h,blst_aux.h} ${PREFIX}/include/blst/
 cp -f libblst.{a,dll,so,dylib} ${PREFIX}/lib/
 
 cat <<EOF > ${PREFIX}/lib/pkgconfig/libblst.pc
 prefix=${PREFIX}
 exec_prefix=\${prefix}
-libdir=\${prefix}
+libdir=\${prefix}/lib
```
angerman, line 19: oops! Thanks for spotting that!

angerman (approved):
LGTM!

cardano-crypto-class/src/Cardano/Crypto/KES/Simple.hs
lehins: Sounds good, I'll take a look at it.

I've seen similar behavior in one of my libraries with ghc-9.4, which complains about redundant constraint, but removing the constraint breaks ghc-8

So, we might have to add CPP for the GHC flag in order to accommodate older ghc versions

PR #395, opened 2023-04-24
Bump package versions

erikd:


lehins:
@erikd  This PR should not have been merged. Versions are wrong in this PR. Please do not release any to CHaPs

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      2.2
 name:               cardano-crypto-class
-version:            2.1.1.0
+version:            2.1.1.1
```
lehins, line 3: `2.1.1.0` have not been released yet, so this bump was unnecessary

PR #397, opened 2023-04-24
bls test vectors

curiecrypt:
Test vectors for `bls12-381`, still under construction!


lehins:
What's going on? Why are we getting PRs with Rust?

iquerejeta:
So this is for the test vectors. We were thinking the other day with @tdammers were should the script that creates test vectors for BLS12-381 bindings live. The conclusion was that the best place for that code to live was in the repo where the test-vectors lived, hence cardano-base. However, this is obviously open to discussion if you think it should be outside of the repo. 

The motivation to put it here is that we could include a test in the CI that checks that the file used for the test vectors is equal to what is generated by the rust script. However, the idea is that we don't put any dependencies on the haskell build with rust. i.e., the rust script only needs to be run by the CI or anyone wanting to re-generate the test vectors. 

WDYT?

iquerejeta:
I understand. The reason why we decided to do the test vectors in rust was to compare the library we use (`blst`) with zkcrypto's [library](https://github.com/zkcrypto/bls12_381) of the same curve, which is a library used in production. Creating the test vectors in C or Haskell would require bindings, and we'd depending on the rust library regardless, so IMO it doesn't really make sense. 

I understand wanting to leave rust outside of the build CI, in which case we can upload the rust code in another repo, link to that in the README and only have here the test vectors. 

If that's your preferred path, we'll go that way 👌

lehins:
> Creating the test vectors in C or Haskell would require bindings

I don't understand why would we need any sort of bindings if we wanted to write a C program that generates test vectors. I would expect it to be same as the rust program in this PR, which depends on Rust implementation of bls, except that program would be written in C and depend on then C implementation bls

> I understand wanting to leave rust outside of the build CI, in which case we can upload the rust code in another repo, link to that in the README and only have here the test vectors.

If you think what I am suggesting is too hard to do then we don't need to waste any time and I am not gonna stand in way of having this code in this repo. After all, it will only be used once to generate the test vectors. If we had to run it on CI and locally every time we had to run the test suite that would be a different story and I would be pushing back much harder against it.

lehins:
> my intention was to include a check in CI that the test-vectors file used for tests corresponds to the output of the rust binary

The output of the Rust program will be checked-in in this repo, no? Then why do we need to run Rust on CI?

> But we want to create these test vectors with the rust library, which is the 'other' reference implementation of bls12-381. We could do it with the underlying library, but the goal of these test vectors is that they are independently generated.

Which library are you trying to test? Are you trying to test the `libblst` library implementation or do you want to test the Haskell binding to `liblst` that we provide in `cardano-crypto-class` library? If it is the latter, then there is nothing wrong with using the same C library to generate test vectors. However, if it is the former and you want to test that Rust implementation matches the C implementation, then such test suite has no business being being in this repository as it is unrelated to our Haskell bindings in any way. A test like this would belong in either the Rust library or in the C library repository.

My understanding was that the reason for writing this script in Rust was only because it is easier to write Rust, rather than C.

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "test-vectors-bls12_381"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+bls12_381 = { version = "0.8.0", features = ["default", "experimental"] }
+rand_core   = "0.6.4"
+sha2 = "0.9"
```
iquerejeta, line 11: Not needed (`rand_core` imported from `rand_chacha`)

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+rand_chacha = "0.3.1"
+ff = "0.13.0"
+group = "0.13.0"
+hex = "0.4.3"
+blst = "0.3.10"
+rand = "0.7.3"
```
iquerejeta, line 17: Ditto

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "test-vectors-bls12_381"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
iquerejeta, line 7: No need to keep this

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/README.md
```diff
@@ -0,0 +1,58 @@
...
+Order of the values printed on `pairing_test_vectors`:
+- `[a]P`    
+- `[b]Q`
+- `[b]P`
+- `[a]Q`
+- `[a + b]P`
+- `[a * b]P`
+
+
+### 2- Test vectors for elliptic curve operations
```
iquerejeta, line 17: Missing introduction on what operations can be tested with these test vectors (addition, subtraction, scalar mult and negation)

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/README.md
```diff
@@ -0,0 +1,58 @@
...
+- `P + Q \in G_1`
+- `P - Q \in G_1`
+- `[scalar]Q \in G_1`
+- `-P \in G_1`
+- `P \in G_2`
+- `Q \in G_2`
+- `P + Q \in G_2`
+- `P - Q \in G_2`
+- `[scalar]Q \in G_2`
+- `-P \in G_2`
```
iquerejeta, line 30: Is the scalar stored somewhere?

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Order of the values printed on `serde_test_vectors`:
+- `G1_uncomp_curve`: G1 point uncompressed is on curve?
+- `G1_comp_curve`: G1 point compressed is on curve?
+- `G1_comp_group`: G1 point compressed is on group?
+- `G1_uncomp_group`: G1 point uncompressed is on group?
+- `G2_uncomp_curve`: G2 point uncompressed is on curve?
+- `G2_comp_curve`: G2 point compressed is on curve?
+- `G2_comp_group`: G2 point compressed is on group?
+- `G2_uncomp_group`: G2 point uncompressed is on group?
```
iquerejeta, line 47: I would change the name of these variables from `G1_uncom_curve` into `G1_uncom_not_on_curve`. Also, The question after the semicolon does not seem to provide any extra info.

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+### 4- BLS Signature
+Test vectors for BLS signature, using `blst` bindings.
```
iquerejeta, line 50: A little more context needed here. In the cardano-base bindings we don't allow the user to use `DST` and `aug` explicitly, and these need to be appended into the, by hand (following [hash-to-curve](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve#name-expand_message) spec), message before verifying.

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/bls_sig_aug_test_vectors
```diff
@@ -0,0 +1,2 @@
+a4c183123f42a817b3ffc8ade477827b4da8ed6a9ab909f66b3045e64861b95fd129fb466eef6376c956d7946d2276b5
```
iquerejeta, line 1: Include all test-vector files in a different folder.

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+use ff::Field;
+use group::Group;
+use rand_chacha::rand_core::{RngCore, SeedableRng};
+use rand_chacha::ChaCha20Rng;
+use std::fs::File;
+use std::io::prelude::*;
+use std::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub};
+
+fn pairing_properties<R: RngCore>(mut rng: R) -> std::io::Result<()> {
+    let P = G1Affine::from(G1Projective::random(&mut rng));
```
iquerejeta, line 18: You can operate in Projective, and do the conversion just at the end.

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+    file.write_all(b"\n")?;
+    file.write_all(hex_bQ.as_ref())?;
+    file.write_all(b"\n")?;
+    file.write_all(hex_bP.as_ref())?;
+    file.write_all(b"\n")?;
+    file.write_all(hex_aQ.as_ref())?;
+    file.write_all(b"\n")?;
+    file.write_all(hex_aplusbP.as_ref())?;
+    file.write_all(b"\n")?;
+    file.write_all(hex_atimesbP.as_ref())?;
```
iquerejeta, line 55: You could create a helper function to take as input a slice of hex strings. That would simplify the code of this script.

cardano-crypto-tests/bls12-381-test-vectors/test-vectors-bls12_381/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+                G1Affine::from_compressed(&G1_random_bytes)
+                    .is_none()
+                    .unwrap_u8(),
+                1
+            );
+            break;
+        }
+    }
+
+    let G1_comp_group_hex = hex::encode(G1_random_bytes);
```
iquerejeta, line 164: This should happen inside the loop, otherwise, it could happen that no point is found and we store a slice of zeroes.

iquerejeta (changes requested):
Left a few comments. Some of the are generic that can be applied in several spots of the script (e.g. the usage of projective point, or the usage of a helper function). Two final comments:
* I'd suggest we put the data files in a different folder
* We now have `bls12-381-test-vectors` that only contains `test-vectors-bls12_381`. I'd suggest we put the contents of `test-vectors-bls12_381` directly in the first folder.

curiecrypt (commented):
LGTM

lehins:
@iquerejeta This PR is up your alley. We've already talked about this work, so if you can oversee this work further that'll be great.

I suspect Haskell tests will be added in a separate PR, right?

iquerejeta:
Haskell tests will be added in a separate PR, yes. I will oversee the PR. If you could approve it to unblock your change request, so that once I approve we can merge.

cardano-crypto-tests/bls12-381-test-vectors/README.md
```diff
@@ -0,0 +1,107 @@
...
+- `G1_comp_not_on_curve`
+- `G1_comp_not_on_group`
+- `G1_uncomp_not_on_group`
+- `G2_uncomp_not_on_curve`
+- `G2_comp_not_on_curve`
+- `G2_comp_not_on_group`
+- `G2_uncomp_not_on_group`
+
+### 4- BLS Signature
+Test vectors for BLS signature, using `blst` bindings.
```
iquerejeta, line 86: We use zkcrypto to create the vectors, but guarantee that they are valid wrt to the `blst` bindings:
```suggestion
Test vectors for BLS signature, with successful validation using [blst](https://github.com/supranational/blst).
```

cardano-crypto-tests/bls12-381-test-vectors/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+The explicit usage of `aug` is not allowed in Cardano-base bindings. Therefore, before verification, they need to be appended into the message by following [hash-to-curve](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve#name-expand_message) spec.
+
+`DST` and `msg` values used to generate test vectors:
+
+```rust
+let dst = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_";
+let msg = b"blst is such a blast";
+let aug = b"Random value for test aug";
```
iquerejeta, line 95:
```suggestion
let aug = b"Random value for test aug. ";
```

iquerejeta (approved):
Looks good. Some minor suggestions before merging.

PR #401, opened 2023-04-26
Try to use input-output-hk/actions/base@latest

angerman:
The iohk-nix GHA should install the same libsodium, libsecp256k1 and libblst as we use in nix.


lehins (approved):
This is beautiful! Thank you!!!

PR #405, opened 2023-05-04
Simplify mlocked forgetting

lehins:
I got to the point where I needed to experiment before I could provide any feedback to #404

So, instead I just wrote a PR myself that applies the changes that I would try to recommend in a PR review


lehins:
target branches was manually rebased on these changes. Closing

PR #407, opened 2023-05-05
cardano-mempool/Setup.hs: get rid of #endif statement

locallycompact:


lehins (approved):
:+1: 

Thank you!

My new project template was messed up. I had this problem in one other place before.

PR #408, opened 2023-05-08
BLS test vectors

tdammers:
Integrate Rust test vectors with the Haskell BLS code and test suite.

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -19,7 +19,6 @@ category:           Currency
 build-type:         Simple
 extra-source-files: README.md
                     CHANGELOG.md
-                    cbits/blst_util.c
 
 flag secp256k1-support
     description: Enable support for functions from libsecp256k1. Requires
@@ -39,7 +38,6 @@ common project-config
 library
   import:            base, project-config
   hs-source-dirs:    src
-  c-sources: cbits/blst_util.c
   exposed-modules:
     Cardano.Crypto.DSIGN
     Cardano.Crypto.DSIGN.Class
--- a/flake.lock
+++ b/flake.lock
@@ -48,6 +48,23 @@
         "type": "github"
       }
     },
+    "blst": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1656163412,
+        "narHash": "sha256-xero1aTe2v4IhWIJaEDUsVDOfE77dOV5zKeHWntHogY=",
+        "owner": "supranational",
+        "repo": "blst",
+        "rev": "03b5124029979755c752eec45f3c29674b558446",
+        "type": "github"
+      },
+      "original": {
+        "owner": "supranational",
+        "repo": "blst",
+        "rev": "03b5124029979755c752eec45f3c29674b558446",
+        "type": "github"
+      }
+    },
     "cabal-32": {
       "flake": false,
       "locked": {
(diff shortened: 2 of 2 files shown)


lehins (changes requested):
Looks good. Just a few comments

lehins:
It's quite peculiar.
@tdammers and @iquerejeta do you guys know why are the tests failing on Windows?
In particular, is it a problem with tests or the actual implementation?

tdammers:
> @tdammers and @iquerejeta do you guys know why are the tests failing on Windows?
In particular, is it a problem with tests or the actual implementation?

I think I know.

The error originates in `Base16.decode`, specifically [here](https://hackage.haskell.org/package/base16-bytestring-1.0.2.0/docs/src/Data.ByteString.Base16.html#decode); it occurs when the length of the bytestring to be decoded is not even (which it has to be because each byte encodes to two hex digits).

And the reason it is uneven on Windows is because (as per review comment) we now use ByteString IO instead of String IO and `lines` from `Data.ByteString.Char8` instead of the `Prelude` one on `String`. My hypothesis is that it goes like this:

- git on the Windows CI machines is configured to automatically convert line endings on checkout (I think that's the default on Windows), so the hex files have `\r\n` line endings on Windows CI, but plain `\n` on Linux and OSX.
- We then open the file in binary mode (`Data.ByteString.readFile`; previously this was `System.IO.readFile`, which uses text mode), so the `\r` is retained.
- `Data.ByteString.Char8.lines` then splits on `\n`; it does not consider `\r` to be a newline character, and does not strip it away.
- The bytestring we feed to `Data.ByteString.Base16.decode` now has a trailing `\r` attached to it on Windows (but not on the other OSes, where git does not inject `\r`, because plain `\n` is the native line ending convention).

cardano-crypto-tests/cardano-crypto-tests.cabal
```diff
@@ -15,6 +15,7 @@ category:            Currency
 build-type:          Simple
 extra-source-files:  README.md
                      CHANGELOG.md
+data-files: bls12-381-test-vectors/test_vectors/*
```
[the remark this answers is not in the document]
tdammers: I've gone with the "list them all individually" solution, at least for now.

lehins:
@tdammers Oh yeah, Windows newline strikes again. I always forget about it. In that case we can go back to using `String` for reading and splitting on newlines: `map BS8.pack . Prelude.lines <$> Prelude.readFile f`. Hex decoding still needs to be done on ByteStrings.

>  line-split (still on Text), 

I thought about the same thing, but unfortunately `text` package also ignores the CR character, from haddock: "[lines](https://www.stackage.org/haddock/nightly-2023-05-16/text-2.0.2/Data-Text.html#v:lines) does not treat '\r' (CR, carriage return) as a newline character."

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -38,6 +39,7 @@ common project-config
 library
   import:            base, project-config
   hs-source-dirs:    src
+  c-sources: cbits/blst_util.c
```
lehins, line 42: This has already been added. See line 127 https://github.com/input-output-hk/cardano-base/blob/f855e4b571de47865341590196b712c7e20cf0d0/cardano-crypto-class/cardano-crypto-class.cabal#L127
```suggestion
```

lehins (approved):
LGTM

lehins:
> Bottom line, just stripping \r like I did in the last commit is probably the sanest solution here.

Yeah, sounds reasonable.

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -19,6 +19,7 @@ category:           Currency
 build-type:         Simple
 extra-source-files: README.md
                     CHANGELOG.md
+                    cbits/blst_util.c
```
angerman, line 22:
```suggestion
```
This seems suspicious? It's listed under c-sources already.

PR #411, opened 2023-05-11
Merge DSIGN and KES classes

tdammers:
This changes things as follows:

## DSIGN:

- The `DSIGNMAlgorithmBase` class disappears; its functionality is provided by the existing `DSIGNAlgorithm` class
- The `DSIGNMAlgorithm` class remains, but becomes a subclass of `DSIGNAlgorithm`
- Any methods and types from `DSIGNMAlgorithmBase` that do not involve sign keys or seeds will be shared between the pure (`DSIGNAlgorithm`) and mlocked (`DSIGNMAlgorithm`) APIs, since we only need to mlock seeds and sign keys. The `M` flavors of those methods and types have been deleted.
- The same phantom types are used for the pure and mlocked (if supported) flavors of DSIGN algorithms
- The distinction between pure and mlocked APIs is made by using the pure or mlocked sign key and seed types, and the pure or mlocked flavors of the corresponding methods (`genKeyDSIGN(M)`, `deriveVerKeyDSIGN(M)`, etc.). All DSIGN algorithms must support a pure API (`DSIGNAlgorithm`); those that also support mlocked DSIGN (currently only ED25519) additionally support `DSIGNMAlgorithm`.
- As a consequence, the signatures and verification keys produced and used by the pure and mlocked flavors of the same DSIGN algorithm are exactly the same.

## KES:

- `KESSignAlgorithm` has been merged into `KESAlgorithm`, which now offers the full range of KES operations. Since the mlocking monad (`m`) is no longer a typeclass parameter, the split is no longer necessary, and conceptually, it does not make sense to have a KES algorithm that can only verify, but not sign.


tdammers:
Depends on #404.

lehins (approved):
Love it!

PR #415, opened 2023-05-18
Reduce use of allow-newer in cabal.project file

erikd:


lehins (approved):
:+1:

cabal.project
```diff
@@ -12,7 +12,7 @@ repository cardano-haskell-packages
 -- The hackage index-state
 index-state: 2023-04-18T07:23:09Z
 -- The CHaP index-state
-index-state: cardano-haskell-packages 2023-04-24T15:32:00Z
+index-state: cardano-haskell-packages 2023-05-18T20:27:48Z
```
lehins, line 15: Updating `index-state` also requires an update to flakes. Run this:
```
$ nix flake lock --update-input CHaP
```

erikd: Done!

PR #418, opened 2023-06-21
Drop cicero from flake

angerman:
Only hydra left now.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -107,7 +107,7 @@
           # we also want cross compilation to windows.
           nixpkgs.lib.optionalAttrs (system == "x86_64-linux") {
           crossPlatforms = p: [p.mingwW64];
-        })
+        });
       in nixpkgs.lib.recursiveUpdate flake {
         # add a required job, that's basically all hydraJobs.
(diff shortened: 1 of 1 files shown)


lehins (approved):
I am out of the loop on the CI plans, so does it mean we switching back to hydra from cicero in all of the repos?

Also, it looks like hydra CI is failing. My suspicion is it shouldn't be :slightly_smiling_face:

jorisdral:
* Should https://github.com/input-output-hk/cardano-base/pull/418/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R13 also be removed?
* Some compilation errors could be solved by `nix flake lock --update-input CHaP`
* A benchmark build is failing in Hydra, and I think that's not showing up in the GitHub Actions CI because we don't build benchmarks there

lehins:
Thanks @jorisdral 
Opened a ticket about benchmarks: https://github.com/input-output-hk/cardano-base/issues/427

lehins:
Closing in favor of #428

PR #420, opened 2023-06-26
Prevent BLS Affine values from being coerced.

tdammers:
The same safeguard already existed for the Point type, but we also need it for Affine.


lehins (approved):
This is definitely a good idea. 
`Affine` seems to be an internal type, so I guess we can consider this as an internal safety measure and thus not a breaking or even a user visible change.

PR #422, opened 2023-07-05
Fix #241: Create `devcontainer.json` to enable GitHub CodeSpaces

yvan-sraka:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,13 @@
+{
+   "image":"ghcr.io/input-output-hk/devx-devcontainer:ghc8107-iog",
```
angerman, line 2: Should we use a more modern compiler here? 9.2, 9.6? For development I think we can use a more recent ones, and have 8.10 only in CI?

angerman (changes requested):
🎉

angerman (approved):

PR #433, opened 2023-08-04
WHNF tests for `strict-checked-vars`

jorisdral:
Adding these tests should safeguard against missing bang patterns and strictness annotations in the future.


strict-checked-vars/test/Test/Control/Concurrent/Class/MonadMVar/Strict/Checked/WHNF.hs
```diff
@@ -0,0 +1,312 @@
...
+                     whnfInvariant, (..:))
+
+{-------------------------------------------------------------------------------
+  Main test tree
+-------------------------------------------------------------------------------}
+
+tests :: TestTree
+tests = testGroup "WHNF" [
+      testGroup "IO" [
+          testIO    "No invariant"      noInvariant
```
coot, line 26: Why not provide a `QuickCheck` instance for `Invariant`?

jorisdral: Done

coot (approved):
LGTM, thanks  @jorisdral.

PR #438, opened 2023-08-29
Make it build with ghc-9.8

erikd:


cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -385,9 +385,12 @@ ptrPsbToSizedPtr = SizedPtr . castPtr
...
 pinnedByteArrayFromListN :: forall a. Prim.Prim a => Int -> [a] -> ByteArray
 pinnedByteArrayFromListN 0 _ =
-    die "pinnedByteArrayFromListN" "list length zero"
+    die "pinnedByteArrayFromListN" "list length zero #1"
 pinnedByteArrayFromListN n ys = runST $ do
-    marr <- newPinnedByteArray (n * Prim.sizeOf (head ys))
+    let headYs = case ys of
+                  [] -> die "pinnedByteArrayFromListN" "list length zero #2"
+                  (y:_) -> y
+    marr <- newPinnedByteArray (n * Prim.sizeOf headYs)
```
erikd, line 393: With `ghc-9.8` the use of `head` triggers a partial function warning.

tdammers: Not a fan of those `#1` / `#2` markers - they may be useful to track the error back to the source code, but they don't make sense unless you actually look at the source code, which is an unnecessary burden for a user of the library. I'd change them to something more descriptive, e.g.: `"specified size is zero"` when the first argument is `0`, and `"list length zero less than specified size"` when the first argument is nonzero but the list is empty - this gives the caller an immediate clue as to what they need to fix to make this function work (specify a nonzero list length, vs. making sure specified list length and actual list length match).

You can also avoid defining `headYs` by adding another top-level pattern matching, like so:
```haskell
pinnedByteArrayFromListN _ [] =
    die "pinnedByteArrayFromListN" "list length zero less than specified size"
pinnedByteArrayFromListN n ys@(y:_) =
    marr <- newPinnedByteArray (n * Prim.sizeOf y)
    ...
```

Then again, the only use site of this function is inside `psbFromBytes`, which itself is deprecated and not used anywhere in the library, so this might not actually be important enough to fix "properly".

cardano-binary/cardano-binary.cabal
```diff
@@ -16,7 +16,7 @@ build-type:          Simple
 extra-source-files:  README.md
                      CHANGELOG.md
 
-common base                         { build-depends: base >= 4.14 && < 4.19 }
+common base                         { build-depends: base >= 4.14 && < 5 }
```
tdammers, line 19: Is it a deliberate choice to move from a known-good upper bound on `base` to a "crystal ball" one? @lehins, opinion?

tdammers (changes requested):

cabal.project
```diff
@@ -37,3 +37,38 @@ benchmarks: true
...
+  tag: 2cb7304e91f17583fc12026762fc5ae6ba9aa1b6
+  subdir:
+    .
+    text-iso8601
+
+source-repository-package
+  type: git
+  -- Git head compiles with `ghc-9.8`, but Hackage version does not.
+  location: https://github.com/recursion-schemes/recursion-schemes
+  tag: cc2e88c3400a6548e975830c9addb12ab087545f
```
jorisdral, line 74: These need `--sha256` entries. See https://github.com/input-output-hk/ouroboros-consensus/blob/main/CONTRIBUTING.md#use-of-source-repository-packages

cabal.project
```diff
@@ -12,7 +12,7 @@ repository cardano-haskell-packages
 -- The hackage index-state
 index-state: 2023-07-31T10:10:32Z
 -- The CHaP index-state
-index-state: cardano-haskell-packages 2023-08-08T14:32:15Z
+index-state: cardano-haskell-packages 2023-08-28T22:23:10Z
```
jorisdral, line 15: Run `nix flake lock --update-input CHaP` when updating the `index-state` for CHaP

cabal.project
```diff
@@ -37,3 +37,38 @@ benchmarks: true
 
 program-options
   ghc-options: -Werror
+
+if impl(ghc >= 9.6)
```
jorisdral, line 41: Should it be this?
```suggestion
if impl(ghc >= 9.8)
```

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
erikd: Your suggestions results in an "Pattern match(es) are non-exhaustive" error.

Currently using at:
```haskell
--   then this throws an exception.
pinnedByteArrayFromListN :: forall a. Prim.Prim a => Int -> [a] -> ByteArray
pinnedByteArrayFromListN 0 _ =
    die "pinnedByteArrayFromListN" "list length zero"
pinnedByteArrayFromListN _ [] =
    die "pinnedByteArrayFromListN" "list length zero"
pinnedByteArrayFromListN n ys@(y:_) = runST $ do
    marr <- newPinnedByteArray (n * Prim.sizeOf y)
```

cardano-binary/cardano-binary.cabal
erikd: Yes, other packages in this repo (eg `cardano-crypto-class`) already have a `base` upper bound of `5`.

cabal.project
```diff
@@ -37,3 +37,38 @@ benchmarks: true
...
+
+-- `aeson` has a whole bunch of dependencies (and sometimes the dependency's dependency) which do
+-- not compile with `ghc-9.8`. Easier to just do this.
+source-repository-package
+  type: git
+  location: https://github.com/input-output-hk/aeson
+  tag: 2cb7304e91f17583fc12026762fc5ae6ba9aa1b6
+  subdir:
+    .
+    text-iso8601
```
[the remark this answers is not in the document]
erikd: This is intended as a placeholder until upstream `aeson` is fixed.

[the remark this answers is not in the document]
lehins: I really don't know, cause I have never tried this, but is it possible to indent this whole SRP to make it used only when in the `if impl(ghc >= 9.8)` branch?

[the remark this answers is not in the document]
erikd: Yes, it is. And done.

cabal.project
erikd: `aeson` SRP was removed and the other fixes to add a sha256 hash.

cabal.project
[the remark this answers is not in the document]
erikd: Actually removed it. No longer necessary.

cabal.project
lehins: This still has the old `8.6` value

erikd: Rebased and whater way too many times.

lehins (approved):

PR #443, opened 2023-10-29
DO NOT MERGE: cardano-crypto-class: Bump vector dependency

erikd:
Just making a PR for a point release.


lehins:
This should not have been a release, it could have been taken care of by a simple revision. 

Regardless of this fact the version bump was incorrect, it should have been a patch release: `2.1.3.1` and the upper bound could have been completely removed, rather than set to `< 0.14`

lehins:
Release was made closing

PR #445, opened 2023-11-08
strict-checked-vars: relaxed bounds io-sim-1.3.0.0

coot:


coot:
I will also add a revision in `CHaP`.

bolt12 (approved):

lehins (approved):
:+1:

coot:
Oh, it got merged despite nix eval errors. Actually none of the checks are `required`, this doesn't look right.

lehins:
>  this doesn't look right.

It is right. Hydra is unreliable in my opinion and it is impossible for a mere human to read those nix error messages. So, we do not block PRs just because nix fails.

That being said it would be nice if you could figure out how to fix it.

PR #446, opened 2023-11-08
Fixed nix build

coot:


lehins:
>there's an unrelated to this PR nothunk test failure on Windows.

Yeah, I've seen it before. It would be nice to get a reproducible example, but the failure doesn't provide us with this opportunity.

However, it looks like it happens on NoThunks test, so I am not gonna worry too much about it.

lehins (approved):

PR #449, opened 2023-12-11
Remove `Switch` modules, release `strict-checked-vars-0.2.0.0`

jorisdral:
This PR remove 'Switch' modules. From now on, instead of switching _imports_, this package switches the _representations_ of checked variables depending on the `checkmvarinvariants` and `checktvarinvariants` flags. This solves a problem where compiling projects that depend on `strict-checked-vars` might succeed with a flag turned on but fail when it is turned off (and vice versa). As an example, consider:
  
```haskell
module A where
import Control.Concurrent.Class.MonadSTM.Strict.TVar
someFunc :: TVar m a -> m ()
someFunc v = {- elided -}

module B where
import Control.Concurrent.Class.MonadSTM.Strict.TVar.Checked.Switch 
someFunc' :: TVar m a -> m ()
someFunc' = someFunc
```

With `-checktvarinvariants`, this code happily compiles, because `Switch` just re-exports `Control.Concurrent.Class.MonadSTM.Strict.TVar`. With `+checktvarinvariants`, this code now fails to compile, because `Switch` now re-exports the `StrictTVar` type from `Control.Concurrent.Class.MonadSTM.Strict.TVar.Checked`, and this type is distinct from the unchecked `StrictTVar`!

By changing the representation of checked `StrictTVar`s using CPP pragmas (instead of using the 'Switch' module), these types of compilation failures can no longer happen.


jorisdral:
> Javier and I chatted earlier. One of the problems is that `ouroboros-network` uses strict vars directly instead of checked-tvars.
> 
> So when `ouroboros-consensus` uses checked-tvars and wants to invoke `ouroboros-network`, they are different types when the flags are enabled. Before this PR, they're different types when the flag is on and equal types when they're off (which meant we didn't notice when this risk became true). After this PR, I think they're _always_ different types---is that right?

You're right, they would always be different types now, which means we would not get different compilation behaviour based on the package flags

> (Prior to seeing this PR, my idea was to add a `forget` function from the checked tvar type (regardless of flags) to the underlying type. This way, regardless of flag, Consensus could use that to pass checked tvars to the Diffusion Layer.)

Added in the form of `unsafeToUncheckedStrictMVar` and `unsafeToUncheckedStrictTVar`!

strict-checked-vars/test/Test/Control/Concurrent/Class/MonadMVar/Strict/Checked.hs
```diff
@@ -43,3 +42,10 @@ prop_invariantShouldFail :: MonadMVar m => PropertyM m ()
 prop_invariantShouldFail = run $ do
     v <- newMVarWithInvariant invPositiveInt 0
     modifyMVar_ v (\x -> pure $ x - 1)
+
+shouldFail :: Property -> Property
```
nfrisby, line 46: I'd suggest a less-biased name, since `shouldFail = id` is a bit non-sensical.

Perhaps `cppToggle` or so?

strict-checked-vars/test/Test/Control/Concurrent/Class/MonadSTM/Strict/TVar/Checked.hs
```diff
@@ -0,0 +1,52 @@
...
+prop_invariantShouldNotFail = run $ atomically $ do
+    v <- newTVarWithInvariant invPositiveInt 0
+    modifyTVar v (+ 1)
+
+prop_invariantShouldFail :: MonadSTM m => PropertyM m ()
+prop_invariantShouldFail = run $ atomically $ do
+    v <- newTVarWithInvariant invPositiveInt 0
+    modifyTVar v (subtract 1)
+
+shouldFail :: Property -> Property
```
nfrisby, line 47: Same as my other comment.

nfrisby (approved):
LGTM

(I'm assuming the `WHNF` tests were accidentally excluded before this PR? That part of the diff seems otherwise orthogonal to this PR.)

strict-checked-vars/strict-checked-vars.cabal
```diff
@@ -90,3 +91,9 @@ test-suite test
     -Wall -Wcompat -Wincomplete-uni-patterns
     -Wincomplete-record-updates -Wpartial-fields -Widentities
     -Wunused-packages -fno-ignore-asserts
+
+  if flag(checkmvarinvariants)
```
lehins, line 95: I don't know if it is too late for this, but I think cabal flags allow dashes? It could be worth adding them for improved readability. (same for `check-tvar-invariants`)
```suggestion
  if flag(check-mvar-invariants)
```

strict-checked-vars/strict-checked-vars.cabal
```diff
@@ -90,3 +91,9 @@ test-suite test
...
+    cpp-options: -DCHECK_MVAR_INVARIANTS
+
+  if flag(checktvarinvariants)
+    cpp-options: -DCHECK_TVAR_INVARIANTS
```
lehins, line 99:
```suggestion
    cpp-options: -DCHECK_TVAR_INVARIANTS

```

strict-checked-vars/strict-checked-vars.cabal
jorisdral: Good point, but I'd rather not change the flag now and break builds silently, since `cabal` won't complain if a flag doesn't exist

jasagredo: It is a pain in the neck that cabal doesn't warn about non-existent flags, and honestly I don't know why this is the case. But in this state, I'd rather not modify the flag name too.

lehins: I think retaining the old flag would allow for non-silent transition:
```
  if flag(checktvarinvariants)
    -- use `check-tvar-invariants` instead
    buildable: false
```

I only made this suggestion because difference between `m` vs `t` in these two flags is not very pronounced: checktvarinvariants vs checkmvarinvariants

PR #456, opened 2024-02-13
Move numeric instances of `EpochNo`and `EpochSize` to `testlib`

### LINKED ISSUE #455: Move the `Num` instance for `EpochNo` and `EpochSize`
`Word64` is the type that both of the `EpochNo` and `EpochSize` are based on. It especially prone to integer overflow when doing arithmetic. For this reason it would be safer to remove `Num` instance alltogether for those types from the main library

This will be even less of an inconvenience for `EpochNo`, because in ledger we have introduced `EpochInterval` type and `addEpochInterval :: EpochNo -> EpochInterval -> EpochNo` function that make the `Num` instance less useful.

The only downside to removing the `Num` instance is that construction from literals requires usage of a constructor, but that is only a tiny burden for testing.

Therefore, instead of completely removing the `Num` instances, I suggest we move them as standalone orphans into the new `testlib`, which would allow us to have the cake and eat it too.

As part of this task we should also move `EpochInterval` and `addEpochInterval` from `cardano-ledger` into the `cardano-slotting`

Lucsanszky:
Resolves #455

### DIFF
--- a/cardano-slotting/CHANGELOG.md
+++ b/cardano-slotting/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.2.0.0
 
+* Add `EpochInterval` and `addEpochInterval` from `cardano-ledger`.
 * Add `binOpEpochNo` helper function to facilitate binary operations on
   `EpochNo`.
 * Remove numeric instances (`Num`, `Integral`, `Real`) of `EpochNo` and
--- a/cardano-slotting/CHANGELOG.md
+++ b/cardano-slotting/CHANGELOG.md
@@ -1,9 +1,19 @@
 # Changelog for `cardano-slotting`
 
-## 0.1.2.0
+## 0.2.0.0
 
+* Add `binOpEpochNo` helper function to facilitate binary operations on
+  `EpochNo`.
+* Remove numeric instances (`Num`, `Integral`, `Real`) of `EpochNo` and
+  `EpochSize` for safety. 
+  They are still available for testing from the `testlib` as orphans.
 * New `Test.Cardano.Slotting.TreeDiff` module extracted from
   `cardano-ledger-binary`. It lives in a new public sublibrary `testlib`.
+
+### `testlib`
+
+* Add numeric instances (`Num`, `Integral`, `Real`) of `EpochNo` and
+  `EpochSize` as orphans.
 
 ## 0.1.1.1
 
--- a/cardano-slotting/cardano-slotting.cabal
+++ b/cardano-slotting/cardano-slotting.cabal
@@ -1,7 +1,7 @@
 cabal-version: 3.0
 
 name:                cardano-slotting
-version:             0.1.2.0
+version:             0.2.0.0
 synopsis:            Key slotting types for cardano libraries
 license:             Apache-2.0
 license-files:
@@ -52,6 +52,7 @@ library testlib
   visibility:           public
   hs-source-dirs:       testlib
   exposed-modules:      Test.Cardano.Slotting.Arbitrary
+                        Test.Cardano.Slotting.Numeric
                         Test.Cardano.Slotting.TreeDiff
   build-depends:        base
                       , cardano-slotting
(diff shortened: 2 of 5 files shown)


cardano-slotting/src/Cardano/Slotting/Slot.hs
```diff
@@ -114,9 +118,28 @@ withOriginFromMaybe (Just t) = At t
...
+  deriving newtype (Enum, Serialise, ToCBOR, FromCBOR, NoThunks, ToJSON, FromJSON, NFData)
 
 newtype EpochSize = EpochSize {unEpochSize :: Word64}
   deriving stock (Eq, Ord, Generic)
   deriving Show via Quiet EpochSize
-  deriving newtype (Enum, Num, Real, Integral, ToCBOR, FromCBOR, NoThunks, ToJSON, FromJSON, NFData)
+  deriving newtype (Enum, ToCBOR, FromCBOR, NoThunks, ToJSON, FromJSON, NFData)
+
+newtype EpochInterval where
+  EpochInterval :: {unEpochInterval :: Word32} -> EpochInterval
```
lehins, line 129: Out of curiosity, why the GADT syntax?

cardano-slotting/cardano-slotting.cabal
```diff
@@ -52,6 +52,7 @@ library testlib
   visibility:           public
   hs-source-dirs:       testlib
   exposed-modules:      Test.Cardano.Slotting.Arbitrary
+                        Test.Cardano.Slotting.Orphans
```
lehins, line 55: Let's call it `Numeric`, since `Arbitrary` and `TreeDiff` modules also provide orphan instances
```suggestion
                        Test.Cardano.Slotting.Numeric
```

lehins (commented):
Needs some changelog entries and a major version bump. Other than that looks good

cardano-slotting/src/Cardano/Slotting/Slot.hs
Lucsanszky: Ha, thanks for the catch! This makes me curious as well as I did not intend to do this. Here is what I think might have happened: when I copied `EpochInterval`, I realised it has `EncCBOR`, `DecCBOR` instances and immediately applied a code action via `hls` in order to attempt to import those. That did not happen obviously since those are defined in `cardano-ledger`, however the code action formatted the code and also transformed it into a `GADT` which I didn't realise. I'll remove it as it is unnecessary. 

By the way, regarding the `EncCBOR`, `DecCBOR` instances: I guess we would prefer to provide those instance definitions as orphans in `cardano-ledger` rather than moving the whole class declarations here and deriving them inplace, right?

lehins: > code action formatted the code and also transformed it into a GADT

That's why I stay away from too much smartness in my tooling :smile: 

> By the way, regarding the EncCBOR, DecCBOR instances: I guess we would prefer to provide those instance definitions as orphans in cardano-ledger rather than moving the whole class declarations here and deriving them inplace, right?

Those instances will need to be defined right next to `EncCBOR`/`DecBOR` type classes in `cardano-ledger-binary`, which will ensure that they are not orphans

cardano-slotting/CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
 # Changelog for `cardano-slotting`
 
-## 0.1.2.0
+## 0.2.2.0
```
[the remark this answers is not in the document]
Lucsanszky: > Interesting way to count 😉

Indeed it is. :sweat_smile:

cardano-slotting/CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
...
 
+* Add `EpochInterval` and `addEpochInterval` from `cardano-ledger`.
+* Add `binOpEpochNo` helper function to facilitate binary operations on
+  `EpochNo`.
+* Move numeric instances (`Num`, `Integral`, `Real`) of `EpochNo` and
+  `EpochSize` into a separate module as orphans.
```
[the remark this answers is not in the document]
Lucsanszky: Also, I realised that I guess we need to add a `testlib` entry to the `CHANGELOG` and indicate that we added some orphan instances there, so I did that. Or is that not required in this case?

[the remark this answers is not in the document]
lehins: >   I realised that I guess we need to add a testlib entry to the CHANGELOG and indicate that we added some orphan instances there, so I did that. 

Yeah, that is good decision

PR #458, opened 2024-02-15
cabal.project: Bump index-states

erikd:


lehins (changes requested):
As always, index-state update needs a flake update :slightly_smiling_face:

lehins (approved):
:+1: Thank you.

PR #460, opened 2024-02-15
Downgrade hlint and weeder.

lehins:
Since current versions are not compatible with ghc-9.2.8


lehins:
Duplicate implementation of #459

PR #464, opened 2024-02-27
CI: Actually use the cache dependencies

erikd:
Previously the GHAs would store the cache after building the dependencies but never restore them. Did a thorough investigation of the issue and added proper comments about how it works and why.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -66,9 +66,12 @@ jobs:
         echo "# cabal.project.local"
         cat cabal.project.local
 
+    # A dry run `build all` operation does *NOT* downlaod anything, it just looks at the package
+    # indices to generate an install plan.
     - name: Build dry run
       run: cabal build all --enable-tests --dry-run --minimize-conflict-set
 
+    # From the install plan we generate a dependency list.
     - name: Record dependencies
       id: record-deps
       run: |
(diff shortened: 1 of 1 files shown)


erikd:
@lehins `ledger` will be getting similar treatment in the near future.

.github/workflows/haskell.yml
```diff
@@ -77,17 +80,37 @@ jobs:
...
+    - name: Restore cached dependencies
+      uses: actions/cache/restore@v4
+      id: cache
+      with:
+        path: |
+          ${{ steps.setup-haskell.outputs.cabal-store }}
+          dist-newstyle
+        key: cache-${{ env.CABAL_CACHE_VERSION }}-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('dependencies.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-ghc-${{ env.GHC_VERSION }}-cabal-${{ env.CABAL_VERSION }}-
```
lehins, line 95: I can't seem to find where this key is being stored and I can't convince myself that it will ever match. In other words, why is this key needed?

lehins (commented):
I don't think this change is needed. 

Cache is being restored just fine on current CI builds. The `actions/cache` action does both saving and restoring, there is no requirement in splitting it into two separate steps. Here is an example of it cahce restoring working: https://github.com/IntersectMBO/cardano-base/actions/runs/8018029965/job/21903084223#step:8:18

So, my question is do we need this duplication, since it already works as expected.

.github/workflows/haskell.yml
erikd: Testing without that key now

erikd: After removing it the caching still works.

erikd:
> I don't think this change is needed.
> 
> Cache is being restored just fine on current CI builds. The `actions/cache` action does both saving and restoring, there is no requirement in splitting it into two separate steps. Here is an example of it cahce restoring working: https://github.com/IntersectMBO/cardano-base/actions/runs/8018029965/job/21903084223#step:8:18
> 
> So, my question is do we need this duplication, since it already works as expected.

Yes, your link shows the cache being stored, but where is the cached data actually being restored before the build begins?

In `cardano-api` the GHA was strong based on the GHA in this repo. After my change to add restoring of the cache before the build, CI build times dropped from about 25 minutes to under 3 minutes.

angerman:
Having consistent alignment across all repos is already a win in my book.

lehins (approved):
LGTM

PR #467, opened 2024-03-13
Fix gitRev TH splice

erikd:
Previously we had a TH quotation calling a TH splce, but that meant that the TH splice was, under certain circumstances, being cached in the object file in `~/.cabal/store`.

 Solution is to only use one level of TH, so the splice is only evaluated once.


erikd:
Test this as an SRP in `cardano-cli` and it works correctly.

erikd:
Not correct yet, so converting to a draft.

As an SRP in `cardano-cli`:

```
> cabal run -- cardano-cli --version
cardano-cli 8.20.3.0 - linux-x86_64 - ghc-9.8
git rev d0438e30ef79ec32649ba2399e900bc28f42bedd

> find . -type f -name cardano-cli
./dist-newstyle/build/x86_64-linux/ghc-9.8.2/cardano-cli-8.20.3.0/x/cardano-cli/build/cardano-cli/cardano-cli

> cd ..

> cardano-cli/./dist-newstyle/build/x86_64-linux/ghc-9.8.2/cardano-cli-8.20.3.0/x/cardano-cli/build/cardano-cli/cardano-cli --version
WARNING: fatal: Needed a single revision

cardano-cli 8.20.3.0 - linux-x86_64 - ghc-9.8
git rev 0000000000000000000000000000000000000000
```

erikd:
Just tested this in the `cardano-cli` repo and it works correctly.

cardano-git-rev/src/Cardano/Git/Rev.hs
```diff
@@ -9,44 +9,65 @@ module Cardano.Git.Rev
...
-         | gitRevEmbed /= zeroRev -> gitRevEmbed
-         | T.null fromGit         -> zeroRev
-         | otherwise              -> fromGit
-    |]
+gitRev
+  | gitRevEmbed /= zeroRev = textE gitRevEmbed
+  | otherwise              =
+#if defined(arm_HOST_ARCH)
+      -- cross compiling to arm fails; due to a linker bug
+      textE zeroRev
```
angerman, line 40: I hope we can get rid of this nonsense soon 🥶

angerman (approved):
Lgtm

PR #468, opened 2024-03-15
cardano-git-rev: Perform check for embedded git rev at run time 

amesgen:
Follow-up to #467, see the second commit message.

### DIFF
--- a/cardano-git-rev/src/Cardano/Git/Rev.hs
+++ b/cardano-git-rev/src/Cardano/Git/Rev.hs
@@ -1,8 +1,14 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE ForeignFunctionInterface #-}
+
+#if __GLASGOW_HASKELL__ >= 900
+{-# LANGUAGE TemplateHaskellQuotes #-}
+#else
+-- https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2288
+{-# LANGUAGE TemplateHaskell #-}
+#endif
 
 module Cardano.Git.Rev
   ( gitRev
--- a/cardano-git-rev/src/Cardano/Git/Rev.hs
+++ b/cardano-git-rev/src/Cardano/Git/Rev.hs
@@ -35,12 +35,7 @@ gitRev :: Q Exp
 gitRev
   | gitRevEmbed /= zeroRev = textE gitRevEmbed
   | otherwise              =
-#if defined(arm_HOST_ARCH)
-      -- cross compiling to arm fails; due to a linker bug
-      textE zeroRev
-#else
       textE =<< TH.runIO runGitRevParse
-#endif
 
 -- Git revision embedded after compilation using
 -- Data.FileEmbed.injectWith. If nothing has been injected,
@@ -48,8 +43,11 @@ gitRev
 gitRevEmbed :: Text
 gitRevEmbed = Text.pack $ drop 28 $ unsafeDupablePerformIO (peekCStringLen utf8 (c_gitrev, 68))
 
-#if !defined(arm_HOST_ARCH)
 runGitRevParse :: IO Text
+#if defined(arm_HOST_ARCH)
+-- cross compiling to arm fails; due to a linker bug
+runGitRevParse = pure zeroRev
+#else
 runGitRevParse = do
     (exitCode, output, errorMessage) <- readProcessWithExitCode_ "git" ["rev-parse", "--verify", "HEAD"] ""
     case exitCode of
--- a/cardano-git-rev/src/Cardano/Git/Rev.hs
+++ b/cardano-git-rev/src/Cardano/Git/Rev.hs
@@ -32,10 +32,11 @@ foreign import ccall "&_cardano_git_rev" c_gitrev :: CString
 -- This must be a TH splice to ensure the git commit is captured at build time.
 -- ie called as `$(gitRev)`.
 gitRev :: Q Exp
-gitRev
-  | gitRevEmbed /= zeroRev = textE gitRevEmbed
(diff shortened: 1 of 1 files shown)


cardano-git-rev/src/Cardano/Git/Rev.hs
```diff
@@ -32,10 +32,11 @@ foreign import ccall "&_cardano_git_rev" c_gitrev :: CString
...
 gitRev :: Q Exp
-gitRev
-  | gitRevEmbed /= zeroRev = textE gitRevEmbed
-  | otherwise              =
-      textE =<< TH.runIO runGitRevParse
+gitRev =
+    [| if
+         | gitRevEmbed /= zeroRev -> gitRevEmbed
+         | otherwise              -> $(textE =<< TH.runIO runGitRevParse)
+    |]
```
erikd, line 40: Not sure about this. It has two TH embedded inside TH. How does the staging work?

I wonder if the correct solution isn't something like:
```
gitRev :: Text
gitRev
  | gitRevEmbed /= zeroRev = textE gitRevEmbed
  | otherwise              = $(textE =<< TH.runIO runGitRevParse)
```

amesgen: > Not sure about this. It has two TH embedded inside TH. How does the staging work?

In the end, `gitrev` will just be a quoted expression, and when spliced at the top-level at usage site, everything will happen in "one go" (in particular, `TH.runIO runGitRevParse` will also be run when spliced, not earlier when eg `cardano-git-rev` gets built).

---

Your snippet would also correctly run the `gitRevEmbed /= zeroRev` check at runtime, but the `$(textE =<< TH.runIO runGitRevParse)` spliced at the top-level, so it (and its embedded `IO`) will be run when building `cardano-git-rev`, which, when another project depends on `cardano-git-rev`, will be cached across multiple runs in the cabal store. This is in contrast to what happens in this PR, where we splice into another quotation, not at the top-level.

Before #467, the same problem was present, as `gitRevFromGit` was spliced at the top-level: 
https://github.com/IntersectMBO/cardano-base/blob/113ff707b91e059d7f41ec18b7c1ade481ab7bba/cardano-git-rev/src/Cardano/Git/Rev.hs#L48
Hence, the splice and the `IO` contained therein was executed while building `cardano-git-rev`.

erikd: You have thoroughly convinced me I still have no idea how TH staging works.

erikd (approved):

cardano-git-rev/src/Cardano/Git/Rev.hs
amesgen: Maybe as another intuition: Here, in `cardano-git-rev`, we just want to provide expressions that users of this libraries can splice in their projects, but we don't actually need to splice anything at the top-level here ourselves. There actually is a subset of `TemplateHaskell` for that, [`TemplateHaskellQuotes`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/template_haskell.html#extension-TemplateHaskellQuotes):

> Template Haskell has the following new syntactic constructions. You need to use the extension [`TemplateHaskell`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/template_haskell.html#extension-TemplateHaskell) to switch these syntactic extensions on. Alternatively, the [`TemplateHaskellQuotes`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/template_haskell.html#extension-TemplateHaskellQuotes) extension can be used to enable the quotation subset of Template Haskell (i.e. without top-level splices). The [`TemplateHaskellQuotes`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/template_haskell.html#extension-TemplateHaskellQuotes) extension is considered safe under [Safe Haskell](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/safe_haskell.html#safe-haskell) while [`TemplateHaskell`](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/template_haskell.html#extension-TemplateHaskell) is not.

I added another commit that only enables `TemplateHaskellQuotes` here to emphasize that we don't (currently) want any top-level splices in `cardano-git-rev`.

cardano-git-rev/src/Cardano/Git/Rev.hs
```diff
@@ -1,7 +1,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TemplateHaskellQuotes #-}
```
erikd, line 4: Seems that is not supported by `ghc-8.10`.

amesgen: Oh right :weary:, `TemplateHaskellQuotes` was made more useful in GHC 9.0: https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2288

I made `TemplateHaskellQuotes` conditional via CPP, but also happy to just unconditionally use `TemplateHaskell` if that seems less noisy.

cardano-git-rev/src/Cardano/Git/Rev.hs
```diff
@@ -1,9 +1,15 @@
...
 {-# LANGUAGE ForeignFunctionInterface #-}
 
+#if __GLASGOW_HASKELL__ >= 900
+{-# LANGUAGE TemplateHaskellQuotes #-}
+#else
+-- https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2288
+{-# LANGUAGE TemplateHaskell #-}
+#endif
+
 module Cardano.Git.Rev
```
erikd, line 13: :+1:

erikd (approved):
LGTM

amesgen:
Can we have a release with this? The next node release will likely use `cardano-git-rev-2.*`, so it would be nice if it would contain this fix. Thanks!

PR #471, opened 2024-04-10
Fix Codespace (no longer call `post-create-command` too early)

yvan-sraka:


angerman (approved):
Lgtm. But the PR could use a description why we drop the dockerile now and why the HLS path setting isn't needed anymore?

PR #473, opened 2024-04-17
Update haskell workflow

angerman:
Adjust to node 20 instead of 16.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -5,6 +5,7 @@ on:
     branches: [ "**" ]
   push:
     branches: [ "master" ]
+  workflow_dispatch:
 
 jobs:
   whitespace:
@@ -18,7 +19,7 @@ jobs:
       fail-fast: false
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Trailing whitespace check
       if: matrix.os != 'windows-latest'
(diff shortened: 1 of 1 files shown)


.github/workflows/haskell.yml
```diff
@@ -36,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.7", "9.6.4", "9.8.1"]
+        ghc: ["8.10.7", "9.6.4", "9.8.1"]
```
lehins, line 40: We still use it as our main development ghc version. I'd prefer not to remove it from CI for now. Also `9.8.2` is now available
```suggestion
        ghc: ["8.10.7", "9.2.8", "9.6.4", "9.8.2"]
```

lehins (commented):
Other than list of GHC versions looks good

.github/workflows/haskell.yml
angerman: We didn't move to 9.6? 👀

erikd: @lehins I think we did move to `9.6`.

lehins: Nope. Our main ghc for deployment of cardano-node is still 8.10.7

lehins: But for development, it depends on the repo. For ledger it is still 9.2.8

angerman:
@lehins do you have admin to this repo? I don't seem to anymore.

lehins (approved):
LGTM. Thank you.

PR #474, opened 2024-05-20
Make it build with ghc 9.10

erikd:
Should not be merged until after the SRPs have been removed.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.8", "9.6.4", "9.8.2"]
+        ghc: ["8.10.7", "9.2.8", "9.6.4", "9.8.2", "9.10.1"]
         os: [ubuntu-latest, macos-13, macos-latest, windows-latest]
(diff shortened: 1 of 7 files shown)


erikd:
All SRPs have been removed.

.github/workflows/haskell.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.8", "9.6.4", "9.8.2"]
+        ghc: ["8.10.7", "9.2.8", "9.6.4", "9.8.2", "9.10.1"]
```
angerman, line 40: This has to have the longest set of compilers, 🤣

cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
```diff
@@ -54,7 +55,9 @@ module Cardano.Crypto.Hash.Class
   )
 where
 
-import Data.List (foldl')
+#if ! MIN_VERSION_base(4,20,0)
+import Data.Foldable (foldl')
+#endif
```
angerman, line 60: 🤦

angerman (approved):

PR #479, opened 2024-05-31
Add RIPEMD160 hashing function

paluh:
I'm really sorry @lehins but auto merge actually closed my PR without merging anything to `master` when I just resynced master with `origin` on github ;-) (I wanted to resync master and rebase locally and the push).

So this is really the same PR against the `master` as the [previous one](https://github.com/IntersectMBO/cardano-base/pull/477).

Please note that I don't bump cardano-crypto-class version in this PR.

--------

I have another question related to backporting - on my fork I created two relevant branches as suggested:

* release/cardano-crypto-class-2.1.4.0
* release/cardano-crypto-class-2.1.5.0

But I assume that these should be created on the `origin` side really. I have them so I can present the diff against these two on my fork:

https://github.com/paluh/cardano-base/compare/release/cardano-crypto-class-2.1.4.0...paluh:cardano-base:release/cardano-crypto-class-2.1.5.0?expand=1

Could you please tell me how to proceed?

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -52,6 +52,7 @@ library
     Cardano.Crypto.Hash.Class
     Cardano.Crypto.Hash.Keccak256
     Cardano.Crypto.Hash.NeverUsed
+    Cardano.Crypto.Hash.RIPEMD160
     Cardano.Crypto.Hash.SHA256
     Cardano.Crypto.Hash.SHA3_256
     Cardano.Crypto.Hash.Short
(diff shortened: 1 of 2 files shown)


lehins (approved):
> I assume that these should be created on the origin side really.

Yes, we need a PR created on this repository for backporting, because, first of all we need to have a record of it happening and most importantly we will be making a release from this repo, so the PR must happen into a branch in this repo. You should have access already to the repo, so you shouldn;t have any issues creating those the same branches in this repo.

> auto merge actually closed my PR without merging anything to master when I just resynced master with origin on github

you can always reopen a PR instead of creating a new one. But there is no harm in creating a new one either.

PR #480, opened 2024-05-31
Add RIPEMD160 hashing function

paluh:
This backport of RIPEMD-160 addition which includes fix for CI build as well.

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.2
 name:               cardano-crypto-class
-version:            2.1.4.0
+version:            2.1.5.0
 synopsis:
   Type classes abstracting over cryptography primitives for Cardano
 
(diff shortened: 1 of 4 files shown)


cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -52,6 +52,7 @@ library
     Cardano.Crypto.Hash.Class
     Cardano.Crypto.Hash.Keccak256
     Cardano.Crypto.Hash.NeverUsed
+    Cardano.Crypto.Hash.RIPEMD160
```
lehins, line 55: Can you also add a changelog entry about this. We'll also need to add this new version with the change on master's changelog, once we release `2.1.5.0` on CHaP

lehins (approved):
Just missing a section `2.1.5.0` in the changelog. Other than that it looks good.

cardano-crypto-class/cardano-crypto-class.cabal
paluh: Sure thing! Just pushed.

I have this tiny change here for `master` as well: https://github.com/IntersectMBO/cardano-base/compare/ripemd-1

Should I open up a PR now with this cosmetic change?

lehins: > Should I open up a PR now with this cosmetic change?

Yes please. (except on master it should also be in section `2.1.5.0` instead of `2.2.0.0`)

PR #485, opened 2024-06-13
Serialisation support for Genesis big ledger peer snapshot

crocodile-dentist:
Networking team needs to pin down CBOR encoding for CDDL spec of the big ledger peer snapshot format.

### DIFF
--- a/cardano-slotting/CHANGELOG.md
+++ b/cardano-slotting/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## 0.2.0.1
 
-*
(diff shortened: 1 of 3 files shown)


lehins (changes requested):
Not only overlapping instances is a bad idea, but also providing an orphan instance in `ouroboros-network-api` that overrides `WithOrigin SlotNo` is a horrible idea.

Cardano is the only user of `WithOrigin`, so we can easily provide monomorphic instances for `FromCBOR` and `ToCBOR` right here in this package

lehins:
Closing this PR as a wrong approach to the problem

PR #488, opened 2024-06-21
Prefix colliding libsodium symbols

jasagredo:
This ensures that the functions that we extracted from upstream libsodium cannot be silently overwritten in linking time.


lehins (commented):
I am all in favor of such rename.

However, why a few functions where excluded from the rename? Eg. `ge25519_p2_dbl`, `ge25519_p3_dbl` ...
The fact that `libsodium` does not currently define them in their header files, doesn't mean they could get exported at some future `libsodium` version. So, I'd err on the side of caution and prefix all functions with `cardano_` that are defined in header files from `cbits`, regardless if there is no name clash at this point. Also, it would be nice if we were consistent in our naming conventions.

cardano-crypto-praos/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog for `cardano-crypto-praos`
 
+# Next version
+
+* Prefixed private bundled c functions with `cardano_` to ensure they are not
+  silently overwritten.
+
 ## 2.1.2.1
 
 *
```
lehins, line 10: We don't have `Next version` in `cardano-ledger` and `cardano-base`. This allows us to cut a release from `master` at any time without any special preceding PRs

See this section about the [CHANGELOG](https://github.com/IntersectMBO/cardano-base/blob/master/RELEASING.md#changelogmd)

Also could you please bump up the version in the cabal file
```suggestion
## 2.2.0.0

* Prefixed private bundled c functions with `cardano_` to ensure they are not
  silently overwritten.
```

lehins (changes requested):
One last change to the changelog and cabal file is needed, other than that it is ready to be merged

lehins (approved):
:+1: Thank you

PR #490, opened 2024-07-03
Add the ByteSize measure

nfrisby:
The new `ByteSize` measure should be used for all byte size _summations_ relevant to the Cardano chain.

- It has checked arithmetic, with a pure representation of overflow.
- It still has a `Monoid` instance that will be useful for finger trees.

The other types defined in the module should be used for recording/transmitting the result of summations that do not overflow.


nfrisby:
TODO test suite

measures/src/Data/Measure/ByteSize.hs
```diff
@@ -0,0 +1,328 @@
...
+--   - An integer greater than the INVARIANT interval of 'ByteSize'.
+--   - Neither of those. For example, negative numbers, fractions, orange, etc.
+--
+-- Law: 'partialFrom' returns 'SNothing' only in the third case above.
+--
+-- Law: 'partialFrom' returns 'SJust' an overflow if and only if the second
+-- case above.
+class ByteSizePartialFrom a where
+    -- | See 'ByteSizePartialFrom'.
+    partialFrom :: a -> StrictMaybe ByteSize
```
lehins, line 224: I wouldn't really call it partial, since partial functions in Haskell usually are the ones that produce exceptions, eg. `error` is a partial function. I get it that math has a different view on the subject, but in order to stick to Haskell nomenclature I'd call it something like `toByteSizeMaybe`

`DefaultSignatures` can reduce some boiler plate
```suggestion
    partialFrom :: a -> StrictMaybe ByteSize
    default partialFrom :: ByteSizeFrom a => a -> StrictMaybe ByteSize
    partialFrom = partialFromDefault
```

measures/src/Data/Measure/ByteSize.hs
```diff
@@ -0,0 +1,328 @@
...
+
+partialFromDefault :: ByteSizeFrom a => a -> StrictMaybe ByteSize
+partialFromDefault = SJust . from
+
+-- | Law: @'partialFrom' = 'SJust' . 'from'@
+class ByteSizePartialFrom a => ByteSizeFrom a where
+    -- | See 'ByteSizeFrom'.
+    from :: a -> ByteSize
```
lehins, line 232: Not only `to` and `from` are very confusing and ambiguous names, but they are also highly likely to be used in user's codebase. Without qualification these will lead to clashes quite often (even `lens`/`microlens` has `to`)

I'd suggest better names that depict what is the actual conversion here. So in order to avoid ambiguity of what is being converted to what, I'd suggest something like this:

```haskell
-- | A type that can be safely converted to `ByteSize`
class ToByteSize a where
    toByteSize :: a -> ByteSize

-- | A type that can be safely constructed from `ByteSize`
class FromByteSize a where
    fromByteSize :: a -> ByteSize
```

measures/src/Data/Measure/ByteSize.hs
```diff
@@ -0,0 +1,328 @@
...
+import           Data.Maybe.Strict (StrictMaybe (..))
+import           Data.Measure (BoundedMeasure, Measure)
+import qualified Data.Measure as Measure
+import           Data.Type.Coercion (Coercion (Coercion))
+import           Data.Word (Word8, Word16, Word32, Word64)
+import           GHC.Generics (Generic)
+#if __GLASGOW_HASKELL__ < 900
+-- Use the GHC version here because this is compiler dependent, and only indirectly lib dependent.
+import           GHC.Natural (Natural)
+#endif
```
lehins, line 56: Why CPP here?
```suggestion
import           Numeric.Natural (Natural)
```

lehins (commented):
I took a quick glance at it. Looks interesting.
I'll do a proper review after the Chang HF, if that is ok?

measures/src/Data/Measure/ByteSize.hs
nfrisby: > Without qualification ...

So far, I'm intending for this module to be imported qualified.

measures/src/Data/Measure/ByteSize.hs
nfrisby: I don't know; I merely copied it from an existing module in the package.

measures/src/Data/Measure/ByteSize.hs
coot: Do we need to export `SJust . from` at all?

measures/src/Data/Measure/ByteSize.hs
```diff
@@ -0,0 +1,328 @@
...
+instance ByteSizePartialFrom Word32 where partialFrom = partialFromDefault
+instance ByteSizeFrom        Word32 where from        = fromSmallerUnsignedIntegral
+instance ByteSizeTo          Word32 where to          = toSmallerBoundedIntegral
+
+instance ByteSizePartialFrom Word64 where partialFrom = partialFromDefault
+instance ByteSizeFrom        Word64 where from        = coerce
+-- | NB the result will never be @'SJust' 'maxBound'@.
+instance ByteSizeTo          Word64 where to          = toBiggerIntegral
+
+instance ByteSizePartialFrom Natural where partialFrom = partialFromDefault
```
coot, line 284: This could overflow, isn't it? Shouldn't we do sth like this here:
```hs
 \a -> if a >= fromIntegral sentinel then SNothing else ...
```

measures/src/Data/Measure/ByteSize.hs
lehins: > So far, I'm intending for this module to be imported qualified.

I appreciate your intention, but that should not be forced on the user. Moreover, qualification does not solve the ambiguity of naming `from` and `to`.

So, to put it bluntly, I will not let functions with names like `from` and `to` to be merged.

noelportillo (commented):
I'm just saying

PR #491, opened 2024-08-28
Updated strict-checked-vars to io-classes-1.7

coot:


coot:
It turns out `io-classes-1.7` will require everybody to switch to `cabal-3.12.1.0` :shrug:

bolt12 (approved):

PR #492, opened 2024-08-30
Import cardano addresses repo

palas:
This PR integrates the repo https://github.com/IntersectMBO/cardano-addresses in https://github.com/IntersectMBO/cardano-base.

This PR:
- Moves the three haskell projects from `cardano-addresess` repo (the pattern is using the name of the cabal file/project as folder inside):
  *  `core` -> `cardano-addresses/cardano-addresses`
  * `command-line` -> `cardano-addresses/cardano-addresses-cli`
  *  `jsapi` -> `cardano-addresses/cardano-addresses-jsapi`
- Also the mainly JS project:
  *  `jsbits` -> `cardano-addresses/cardano-addresses-jsbits`
- Adds a nix output that generates a docker image with the `cardano-cli` executable (e.g: `.#docker.x86_64-linux.cardano-address`).
- Adds `ghcjs` cross-compilation
  * But I have discarded all packages from `ghcjs` cross-compilation except for those in `cardano-addresses`, because they weren't compiling.
- Migrates the three custom derivations from `cardano-addresses-js`:
  * `.#cardano-addresses-js`
  * `.#cardano-addresses-demo-js`
  * `.#cardano-addresses-js-shell`
- Adds convenience derivation with patched sources of `cardano-addresses-jsbits` called `.#jsbits`, because it seemed quite necessary for following the `README.md`s in `jsapi` and `jsbits`.
- Does some fixes to the `cardano-addresses` code to make it compatible with ghc above `9.8.2`, some required having some `CPP` directives.

Considerations:
- `jsapi` tests are disabled except for `ghcjs`, I think they are only meant to work on JS
- I've tried to adapt the `README.md`s, and I moved the one in the root to the `cardano-addresses` subdir.
- I haven't migrated the `musl` outputs, or any artifact generation. Happy to do them in another PR (or even partially in this one).
- There is an issue with docs, in `cardano-addresses` the haddock is not in the top level index because there are more things deployed: [demo app](https://intersectmbo.github.io/cardano-addresses/demo/) and [js doc](https://intersectmbo.github.io/cardano-addresses/typescript/). I haven't migrated either, it would require to move the `cardano-base` a level higher. Happy to do that in another PR.
- Dependency on `ghcjs-overlay`, needs to be updated periodically for `cabal build` to work without `nix`, this is done with an action currently. This is important for the CI action that caches the CodeSpaces. Don't know how the CI action is triggered, may need credentials.
- There is an [associated NPM package](https://www.npmjs.com/package/cardano-addresses), and it is still pointing to the `input-output-hk` repo for documentation. Presumably that requires credentials to manage too.
- In order to use the `cardano-addresses` libraries from this repo, I think it is necessary to cut a new version of the library and add it to CHaP. I have never done a CHaP release, but probably makes sense to do it in a separate PR.
- I have disabled `hjsonschema` library for versions of `ghc` greater than `9.8.2`, because it doesn't compile and is deprecated. It seems this only affects a test which is still being run on the other versions of ghc, so it is not a problem yet. Also, it was already disabled for `ghcjs`.


palas:
Branch before rebasing: https://github.com/IntersectMBO/cardano-base/tree/backup/import-cardano-addresses-repo

palas:
Comit: https://github.com/IntersectMBO/cardano-base/commit/aa66568d2c66feda7ffbf1dc7621e24b0e3690f3

palas:
Commit: https://github.com/IntersectMBO/cardano-base/commit/aa66568d2c66feda7ffbf1dc7621e24b0e3690f3

lehins (changes requested):
Oh my god! :open_mouth:  +42K lines of code. `cardano-addresses` is bigger than the whole of `cardano-base` repo right now.

We need to seriously discuss the future of `cardano-addresses`, because I am no longer sure that embedding it into any other repo is a good idea.
Looking at all of the changes in this PR and the massive size of `cardano-addresses` makes me realize that this is just not going to work.

lehins:
Closing, as it was in the decided to keep `cardfano-addresses` as a separate repo with its own maintainers.

PR #493, opened 2024-09-11
Revert "Updated strict-checked-vars to io-classes-1.7"

coot:
This reverts commit 8a5991475e4234f196a23aeb18cc4b2b42b16e22.

Haskell.nix doesn't support `io-classes ^>= 1.7`; we will stick with `io-classes ^>= 1.5`, which doesn't use public sublibraries at all.


lehins (approved):
:+1:

PR #498, opened 2024-09-12
Fix recent changes to haddock generation

lehins:
#496 broke the actual haddock generation. This time around I copied the full setup from ledger, so it should definitely work, since it does for `cardano-ledger` repo.


teodanciu (approved):
Looks good to me

PR #499, opened 2024-09-12
Disable `aarch64-darwin` build with flakes

lehins:
This should fix hydra CI.

After it is fixed it will also be marked as required


lehins:
Closing in favor of #503

PR #500, opened 2024-09-12
Add nightly ci and cancel conflicting runs

lehins:
This PR ensures we don't have duplicate CI runs for the same PR.
Also it adds nightly CI runs


neilmayhew (approved):
LGTM

There are probably other features we'll want to bring across from ledger, but we can leave that for another time.

PR #503, opened 2024-10-16
revert haskell.nix update

angerman:
This hopefully unbreaks the CI issues we see.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -39,3 +39,7 @@ benchmarks: true
 program-options
   ghc-options: -Werror
 
+-- Windows cross with bitvec/simd is broken.
+package bitvec
+  -- Workaround for windows cross-compilation
+  flags: -simd
--- a/flake.nix
+++ b/flake.nix
@@ -92,7 +92,7 @@
                 # tools that work only with default compiler
                 fourmolu = fourmoluVersion;
                 hlint = "3.8";
-                haskell-language-server = "2.9.0.0";
+                haskell-language-server = "2.2.0.0";
               };
 
             # and from nixpkgs or other inputs
--- a/flake.nix
+++ b/flake.nix
@@ -99,7 +99,7 @@
                 # tools that work only with default compiler
                 fourmolu = fourmoluVersion;
                 hlint = "3.8";
-                haskell-language-server = "2.2.0.0";
+                haskell-language-server = "2.9.0.0";
               };
 
             # and from nixpkgs or other inputs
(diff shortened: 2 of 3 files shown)


angerman:
@lehins this is as green as we can get it right now, I believe. There are only some 9.8.2 windows cross build failures, everything else is green.

flake.nix
```diff
@@ -48,7 +56,7 @@
         inherit (nixpkgs) lib;
 
         # see flake `variants` below for alternative compilers
-        defaultCompiler = "ghc966";
+        defaultCompiler = "ghc965";
```
lehins, line 59: Why is the downgrade?

flake.nix
```diff
@@ -84,7 +92,7 @@
             # tools we want in our shell, from hackage
             tools =
               {
-                cabal = "3.12.1.0";
+                cabal = "3.10.3.0";
```
lehins, line 95: Same here, how come we need to downgrade cabal?

cabal.project
```diff
@@ -39,3 +39,7 @@ benchmarks: true
 program-options
   ghc-options: -Werror
 
+-- Windows cross with bitvec/simd is broken.
+package bitvec
+  -- Workaround for windows cross-compilation
+  flags: -simd
```
lehins, line 45: Do you know if this is a valid syntax for `cabal.project`? Never tried it before, but if it works it would be much better to turn it off only on platforms that are problematic. The concern I have is that if we do use `simd` on linux and mac for the actual `cardano-node` then all of the tests in this repo are not tested with the same flags
```suggestion
  if os(windows)
    flags: -simd
```

lehins (approved):
Looking at the changes in the PR makes me believe that something is broken on haskellNix and we are waiting for it to be fixed. Only after it is fixed then we restore using newer ghc and cabal. Is that correct?
If my understanding is correct, could you please link to a github issue that tracks this, so we know when we can get rid of the git pin for haskellNix.

Also, could I ask you to cleanup the git history in this PR a little bit. If you won't get it to it, not the end of the world, I'll squash all them together when merging.

lehins (changes requested):
Looks like there are still failures on hydra. We do need to fix all of them

cabal.project
angerman: This sadly does not work.

flake.nix
angerman: This seems to work.

flake.nix
angerman: There is not later ghc in the haskell.nix pin :(

angerman:
As mentioned on slack, and forgot to commit here: we are stuck on the old haskell.nix pin until we can drop 8.10

lehins (approved):
Thank you!

PR #507, opened 2024-11-01
Update haskell.nix and disable GHC 8.10 Windows

hamishmack:
This PR will update to the latest haskell.nix, but disables GHC 8.10 windows cross compilation as the current combination of haskell.nix and nixpkgs is broken for GHC 8.10 and windows cross compilation.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -202,43 +202,6 @@
         "type": "github"
       }
     },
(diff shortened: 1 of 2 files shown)


lehins (approved):
Perfect. Thank you!

PR #509, opened 2024-11-07
expose Scalar, Field and the Point ptr

perturbing:
# Description

This PR exposes the constructors of the types: `PointPtr`, `Point` and `Scalar`. Additionally, it exports the type `Fr` type and its constructor. 

The goal here is so that we can use them in [this](https://github.com/cardano-scaling/haskell-accumulator/blob/86d7bb20ab529878212191b766774e972297d4a9/haskell-accumulator/lib/Bindings/Internal.hs#L9) package.

This change will cause some trouble for the plutus repository (I had to fork it when also depending on cardano-base with these changes).

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
@@ -42,7 +42,7 @@ module Cardano.Crypto.EllipticCurve.BLS12_381.Internal
   , BLSTError (..)
   , Point (..)
   , Point1
-  , Point2 
+  , Point2
   , PT
   , Scalar (..)
   , Fr (..)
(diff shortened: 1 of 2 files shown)


lehins (approved):
I don't see a problem of exporting unsafe constructors from an internal module. It is important to realize that this is really internal and unsafe, so regular libraries should not depend on it.

It is better to provide a safe interface for theses things if they are needed downstream

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -40,11 +40,12 @@ module Cardano.Crypto.EllipticCurve.BLS12_381.Internal
   , Affine1
   , Affine2
   , BLSTError (..)
-  , Point
+  , Point (..)
   , Point1
-  , Point2
+  , Point2
```
lehins, line 45:
```suggestion
  , Point2
```

PR #510, opened 2024-11-08
Support ghc-9.12

erikd:
# Description

# Checklist

- [ ] Self-reviewed the diff


orphans-deriving-via/src/Data/DerivingVia/NoThunks.hs
```diff
@@ -14,21 +14,21 @@ module Data.DerivingVia.NoThunks
...
+--   wNoThunks ctxt (InstantiatedAt x) =
+--       gwNoThunks (Proxy @'[]) ctxt fp
+--     where
+--       !fp = from x
+--
+--   showTypeOf _ = gShowTypeOf (from (undefined :: a))
+--
+-- -- Copied from the "NoThunks.Class"
+-- class GShowTypeOf f where gShowTypeOf :: f x -> String
+-- instance Datatype c => GShowTypeOf (D1 c f) where gShowTypeOf = datatypeName
```
erikd, line 34: Not sure about this. `GWNoThunks` seems to have been removed from the `nothunks` library.

erikd: Have asked @nfrisby in Slack about this.

cardano-crypto-tests/cardano-crypto-tests.cabal
```diff
@@ -95,6 +95,9 @@ library
                      Test.Crypto.Vector.StringConstants
                      Test.Crypto.Vector.Vectors
 
+  if impl(ghc >= 9.12)
+    ghc-options:     -Wno-deriving-typeable
```
lehins, line 99: It is better to remove all places where Typeable is being derived, instead of turning off the warning

erikd: I am assuming that all places that `Typeable` is derived, it is needed for compilers `< 9.12`.

lehins: > I am assuming that all places that Typeable is derived

FYI. All types by default in haskell starting with ghc-7.10 have `Typeable`, so it does not need to be derived anymore. That is exactly what the warning is about.

So, please do not disable warnings, they are actually trying to tell you something.

All you need to do is remove deriving Typeable. That's all.

lehins: Here is some documentation that you can read about it: https://hackage.haskell.org/package/base-4.20.0.1/docs/Data-Typeable.html#description

erikd: Have removed `-Wno-deriving-typeable`.

orphans-deriving-via/src/Data/DerivingVia/NoThunks.hs
erikd: Fixed.

cabal.project
```diff
@@ -39,3 +38,23 @@ program-options
...
+
+if impl(ghc >=9.12)
+  allow-newer:
+    -- Stuck on `cabal-3.14` issues and maintainer difficulties
+    -- https://github.com/haskell/aeson/issues/1124
+    -- Stuck on `cabal-3.14` and unresponsive maintainers
+    -- https://github.com/well-typed/cborg/pull/339
+    -- Fix blocked by aeson
+    -- https://github.com/haskellari/microstache/issues/43
+    -- Difficult maintainer
```
lehins, line 50: :smiling_face_with_tear:

erikd: Hoping this can still be merged with these `allow-newer` stanzas. No idea when any of these will be fixed.

lehins (approved):
We can definitely merge it with `allow-newer` stanzas, since the they are conditional on ghc-9.12

Good job on pushing the community to update their packages. Sorry you had to deal with difficult maintainers because of this!

PR #514, opened 2024-11-26
Perturbing/add msm bls

perturbing:
# Description

### Summary

This PR introduces multi-scalar multiplication (MSM) for BLS12-381 elliptic curve points:

---

#### ⚙️ Internal Module (`BLS12_381/Internal.hs`)
- Added types for projective/affine points and scalars, including array/block representations
- Added FFI bindings for 6 C functions:
  - `blst_{p1s,p2s}_mult_pippenger`, `blst_{p1s,p2s}_to_affine`, and `c_blst_{p1s,p2s}_mult_pippenger_scratch_sizeof`
- Added helper marshalling functions:
  - `withPointArray`, `withScalarArray`, `withAffineBlockArrayPtr`
- Implemented `blsMSM` with filtering of points-at-infinity and zero scalars for safety and efficiency

#### ✅ Tests (`Test/EllipticCurve.hs`)
- Added property tests comparing `blsMSM` against naive implementations
- Fixed `prop_randomFailsFinalVerify` to use group-based logic instead of point inequality
- Updated `Arbitrary` instance for `Point`
 - Increased probability of generating `blsZero` for better edge-case coverage

# Checklist

- [ ] Commits that only contain large amounts of formatting changes were added to `.git-blame-ignore-revs`
- [x] Self-reviewed the diff


lehins (commented):
Haven't looked into ffi calls, just at the haskell portion of it. 
I'll look into this in more depth once PR is out of draft

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +918,47 @@ scalarCanonical scalar =
...
+---- MSM operations
+
+-- | Multi-scalar multiplication using the Pippenger algorithm.
+-- The number of points must be equal or smaller than the number of scalars.
+-- For reference, see the usage of the rust bindings: https://github.com/perturbing/blst/blob/master/bindings/rust/src/pippenger.rs#L143C1-L161C11
+-- Note that we only implement the single thread version of the algorithm.
+blsMSM :: forall curve. BLS curve => [Point curve] -> [Scalar] -> Either BLSTError (Point curve)
+blsMSM ps ss
+  | null ps || null ss = Left BLST_UNKNOWN_ERROR
+  | length ps /= length ss = Left BLST_UNKNOWN_ERROR
```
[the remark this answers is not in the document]
perturbing: You are correct that the underlying blst implementation allows for it, but it just doesn't use the excess scalars. CIP-0133 [constrains](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0133#function-definition) it to be equal size.

perturbing:
Thank you for taking a first look @lehins, much appreciated!

The FFI is not working yet, I am getting some segmentation faults that I am trying to debug with valgrind.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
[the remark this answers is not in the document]
lehins: In that case the haddock is wrong

perturbing:
Hi, I fixed the bug I encountered with the memory layout (I overlooked how C code wanted the pointer).

I also added a property test, but got some weird behavior where my test's success depends on the running of other tests. See my comments [here](https://github.com/perturbing/cardano-base/blob/9bad7c5b075f6c80691bf5515257d54bd3dd991a/cardano-crypto-tests/test/Main.hs#L40) and [here](https://github.com/perturbing/cardano-base/blob/9bad7c5b075f6c80691bf5515257d54bd3dd991a/cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs#L49).

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
```diff
@@ -132,6 +138,12 @@ testBLSCurve name _ =
         BLS.blsMult (BLS.blsMult a b) c === BLS.blsMult (BLS.blsMult a c) b
     , testProperty "scalar mult distributive left" $ \(a :: BLS.Point curve) (BigInteger b) (BigInteger c) ->
         BLS.blsMult a (b + c) === BLS.blsAddOrDouble (BLS.blsMult a b) (BLS.blsMult a c)
+    , testProperty "MSM matches naive approach" $ \(psAndSs :: [(BLS.Point curve, BigInteger)]) ->
+        not (null psAndSs) ==>
```
lehins, line 142: QuickCheck has its own [`NonEmptyList`](https://hackage.haskell.org/package/QuickCheck-2.15.0.1/docs/Test-QuickCheck-Modifiers.html#t:NonEmptyList) that can be used for generating non empty lists:
```suggestion
    , testProperty "MSM matches naive approach" $ \(NonEmpty (psAndSs :: [(BLS.Point curve, BigInteger)])) ->
```

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
```diff
@@ -132,6 +138,12 @@ testBLSCurve name _ =
...
+        let points = map fst psAndSs
+            scalars = map ((\(BigInteger x) -> x) . snd) psAndSs
+            pairs = NonEmpty.fromList (zip points scalars)
```
lehins, line 145: This can be easily written with list comprehensions:
```suggestion
        let pairs = NonEmpty.fromList [(p, i) | (p, BigInteger i) <- psAndSs]
```

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
```diff
@@ -132,6 +138,12 @@ testBLSCurve name _ =
...
+        in BLS.blsMSM pairs === foldr (\(p, s) acc -> BLS.blsAddOrDouble acc (BLS.blsMult p s)) (BLS.blsZero @curve) (zip points scalars)
```
lehins, line 146: `NonEmpty` is also `Foldable`, so we can just use `pairs` here:
```suggestion
        in BLS.blsMSM pairs === foldr (\(p, s) acc -> BLS.blsAddOrDouble acc (BLS.blsMult p s)) (BLS.blsZero @curve) pairs
```

lehins (changes requested):
I have quickly looked through the PR, but haven't spotted exactly what is wrong with this PR, but there is definitely something seriously wrong with memory management in this PR. All the non-deterministic test failures, which should not be the case with functionality that pretends to be "pure" serve as a good indicator that the functionality has a bug. Moreover, the fact that unrelated tests are affected is another strong indicator that something is seriously wrong!

I'll try to dig deeper into this functionality some time this week. Maybe I can help you get to the bottom of this.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -428,6 +459,17 @@ withNewScalar_ = fmap fst . withNewScalar
...
 withNewScalar' = fmap snd . withNewScalar
 
+withScalarVector :: NonEmpty.NonEmpty Scalar -> (ScalarPtrVector -> IO a) -> IO a
+withScalarVector scalars go = do
+  let numScalars = NonEmpty.length scalars
+  let sizeReference = sizeOf (undefined :: Ptr ())
+  allocaBytes (numScalars * sizeReference) $ \ptr -> do
+    let copyPtrAtIx ix scalar =
+          withScalar scalar $ \(ScalarPtr sPtr) -> do
+          poke (ptr `advancePtr` ix) sPtr
```
tdammers, line 469: Wouldn't this leak the address in `sPtr` outside the `withScalar` block?

perturbing: Sharp! The same is true for the vector of pointers that is built in the `withAffineVector` function

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +951,51 @@ scalarCanonical scalar =
...
+            scratchSize = fromIntegral @CSize @Int $ c_blst_scratch_sizeof (Proxy @curve) numPoints'
+
+        allocaBytes scratchSize $ \scratchPtr -> do
+          withNewPoint' $ \resultPtr -> do
+            c_blst_mult_pippenger
+              resultPtr
+              affineVectorPtr
+              (fromIntegral numPoints)
+              scalarVectorPtr
+              255
```
kwxm, line 996: Could you add a comment saying what this is for, and why it's 255?  I think it corresponds to the `nbits` argument in the C code, but I'm having trouble working out what that's for.   I thought for a while that it might be the number of bits required by a scalar (in which case it would be too small), but I'm probably wrong about that.

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
```diff
@@ -132,6 +136,10 @@ testBLSCurve name _ =
         BLS.blsMult (BLS.blsMult a b) c === BLS.blsMult (BLS.blsMult a c) b
     , testProperty "scalar mult distributive left" $ \(a :: BLS.Point curve) (BigInteger b) (BigInteger c) ->
         BLS.blsMult a (b + c) === BLS.blsAddOrDouble (BLS.blsMult a b) (BLS.blsMult a c)
+    , testProperty "MSM matches naive approach" $ \(NonEmpty (psAndSs :: [(BLS.Point curve, BigInteger)])) ->
```
kwxm, line 139: You might want to bias it towards including some zero points in the inputs here.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -288,6 +307,22 @@ withNewAffine_ = fmap fst . withNewAffine
 withNewAffine' :: BLS curve => (AffinePtr curve -> IO a) -> IO (Affine curve)
 withNewAffine' = fmap snd . withNewAffine
 
+withAffineVector :: NonEmpty.NonEmpty (Affine curve) -> (AffinePtrVector curve -> IO a) -> IO a
+withAffineVector affines go = do
+  let numAffines = NonEmpty.length affines
+      sizeReference = sizeOf (undefined :: Ptr ())
```
kwxm, line 313: I know that it doesn't get evaluated, but that `undefined` still makes me very uncomfortable.  It looks like there's no alternative though: I tried `Proxy` and GHC didn't like it.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -288,6 +307,22 @@ withNewAffine_ = fmap fst . withNewAffine
...
+  allocaBytes (numAffines * sizeReference) $ \ptr ->
+    -- The accumulate function ensures that each `withAffine` call is properly nested.
```
kwxm, line 315: I know very little about the FFI and this problem certainly wouldn't have occurred to me.  I found myself worrying about laziness though, and whether there should be some `!`s somewhere.  Does the monadic code ensure that all of this happens in the right order?

kwxm (changes requested):
This looks OK, but there's a problem with inputs involving the zero point which needs to be fixed.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +951,51 @@ scalarCanonical scalar =
...
+
+-- | Multi-scalar multiplication using the Pippenger algorithm.
+-- The scalar will be brought into the range of modular arithmetic
+-- by means of a modulo operation over the 'scalarPeriod'.
+-- Negative number will also be brought to the range
+-- [0, 'scalarPeriod' - 1] via modular reduction.
+blsMSM :: forall curve. BLS curve => NonEmpty.NonEmpty (Point curve, Integer) -> Point curve
+blsMSM psAndSs =
+  unsafePerformIO $ do
+    let (points, scalarsAsInt) = unzipPointsAndScalars psAndSs
```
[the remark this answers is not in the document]
perturbing: This is indeed something that I did not expect to happen, and glad you caught it. 

I followed the rust binding implementation, and a little [test](https://github.com/perturbing/bench-msm/blob/perturbing/test-msm-inf-point/src/main.rs) shows that it has the same behavior!

I agree that this is not something we want, as adding zero to the list of inputs would not alter the answer. I propose we use 
```Haskell
-- | Infinity check on curve points.
blsIsInf :: BLS curve => Point curve -> Bool
blsIsInf p = unsafePerformIO $ withPoint p c_blst_p_is_inf
```
To filter out pairs of points and scalars for which the point is the zero element.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
perturbing: This is the size of the scalars (the order of the EC group is a 255 bit prime). I followed the rust `blstrs` implementation for these bindings, to match its use [here](https://docs.rs/blstrs/latest/src/blstrs/g1.rs.html#645) (which calls the blst rust bindings [here](https://github.com/supranational/blst/blob/165ec77634495175aefd045a48d3469af6950ea4/bindings/rust/src/lib.rs#L1615)).

kwxm: Ah, OK.  I was thinking about the size of the field elements, which definitely don't fit into 255 bits.  I wonder why the `blst` function takes this as a parameter instead of having it hard-coded or `#define`d somewhere: won't it always be 255?

cardano-crypto-tests/src/Test/Crypto/EllipticCurve.hs
perturbing: That is indeed a smart thing to do, I changed the arbitrary instance for `Point curve` in 8cd6ab6d35b51e9b7fa7b89d997b3db1b580f4bd. Note that I had to adjust the `prop_randomFailsFinalVerify` due to this change, as the former definition
```Haskell
prop_randomFailsFinalVerify :: BLS.Point1 -> BLS.Point1 -> BLS.Point2 -> BLS.Point2 -> Property
prop_randomFailsFinalVerify a b c d =
  a /= b && c /= d ==>
    BLS.ptFinalVerify (BLS.millerLoop a c) (BLS.millerLoop b d) === False
```
fails in the case where in each term, a value is `blsZero`. 

More explicitly, we have that this is true `e(a,blsZero) == e(c, blsZero)`, for which we have that `a/=b && c /= d`. In math, given `a=g^w`, `b=g^x`, `c=g^y` and `d=g^z`, for some scalars `w,x,y,z`, then by the bilinear property, we are basic testing in `e(a,c)==e(b,d)` that `w*y == x * z`. And in the case where both `c` and `d` are `blsZero = g^0`, we trivially have `w*0 == x* 0`. 

That I changed the above test to the more explicit form
```Haskell
prop_randomFailsFinalVerify :: BigInteger -> BigInteger -> BigInteger -> BigInteger -> Property
prop_randomFailsFinalVerify (BigInteger a) (BigInteger b) (BigInteger c) (BigInteger d) =
  (a * c `mod` BLS.scalarPeriod) /= (b * d `mod` BLS.scalarPeriod) ==>
    let a' = BLS.blsMult (BLS.blsGenerator @BLS.Curve1) a
        b' = BLS.blsMult (BLS.blsGenerator @BLS.Curve1) b
        c' = BLS.blsMult (BLS.blsGenerator @BLS.Curve2) c
        d' = BLS.blsMult (BLS.blsGenerator @BLS.Curve2) d
     in BLS.ptFinalVerify (BLS.millerLoop a' c') (BLS.millerLoop b' d') === False
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
perturbing: > I was thinking about the size of the field elements, which definitely don't fit into 255 bits.

The size of a field element is bound by the modulus
```Haskell
-- | The period of scalar modulo operations.
scalarPeriod :: Integer
scalarPeriod = 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001
```
Note that though this is of length 32 bytes in hex notation (and thus 256 bits), it does fit in 255 bits (see [here](https://www.rapidtables.com/convert/number/hex-to-decimal.html?x=73EDA753299D7D483339D80809A1D80553BDA402FFFE5BFEFFFFFFFF00000001)).

And yes, this is constant. I am unsure why it is not.

kwxm: > The size of a field element is bound by the modulus
I think the field elements are 381 bits long, hence the name.  The 255-bit scalar period is the order of G1,G2 and GT.  I keep confusing the two!

perturbing: There two fields at play here, the base field for the curve and the scalar field. The former has elements of size 381 bits and the latter has elements of size 255 bits. 

Now for the former, we construct an elliptical curve with `x` and `y` coordinates in this base field such that `y^2 = x^3 + 4` holds. This elliptical curve forms a group, and it has a subgroup `G1` of prime order `p`. This means that it has a generator `g` and that all elements `a` of this group can be written in the form `a = g^n`. Now since this is a cyclic group, the prime order introduces a new field, which it the latter field of 255 bits.

Now the G2 and GT groups arise by making the curve equation multi dimension, they behave the same. So indeed, the prime order (`scalarPeriod` above) is also the order of these two groups :smile:

perturbing (commented):
Hi, I am ready for another round of review (see also the PR description for an overview of the changes). 

Thanks in advance, much appreciated.

I think it's best to not yet merge this PR, best to first let @kwxm implement it in plutus via an SRP stanza.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -183,16 +193,36 @@ data Curve2
 
 ---- Unsafe PointPtr types
 
+-- A pointer to a (projective) point one of the two elliptical curves
```
tdammers, line 196: Why not make this a proper Haddock comment:
```suggestion
-- | A pointer to a (projective) point one of the two elliptical curves
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -288,6 +318,40 @@ withNewAffine_ = fmap fst . withNewAffine
 withNewAffine' :: BLS curve => (AffinePtr curve -> IO a) -> IO (Affine curve)
 withNewAffine' = fmap snd . withNewAffine
 
+withPointArray :: [Point curve] -> (PointArrayPtr curve -> IO a) -> IO a
```
tdammers, line 321: Relying on null-termination is dangerous. Right now, usages of this function are safe, but this is brittle: the second argument to `withPointArray` there closes over the size of the original list, so it gets passed in "out of band", and implicitly depends on `withPointArray` keeping the original list length unchanged (which right now it does, but that's an implicit assumption, not part of the explicit API).

`withPointArray` as it is invites incorrect usage, so I would suggest passing the size explicitly (cf. `CString` vs. `CStringLen`):

```suggestion
withPointArray :: [Point curve] -> (Int -> PointArrayPtr curve -> IO a) -> IO a
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +993,72 @@ scalarCanonical scalar =
...
+        foldr
+          (\(scalar, pt) acc -> blsAddOrDouble acc (blsMult pt (unsafePerformIO $ scalarToInteger scalar)))
+          blsZero
+          filteredPoints
+    _ -> do
+      let (scalars, points) = unzip filteredPoints
+          numPoints = length points
+
+      withNewPoint' @curve $ \resultPtr -> do
+        withPointArray points $ \pointArrayPtr -> do
```
tdammers, line 1040: See comment on `withPointArray`:
```suggestion
        withPointArray points $ \numPoints pointArrayPtr -> do
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -428,6 +505,26 @@ withNewScalar_ = fmap fst . withNewScalar
 withNewScalar' :: (ScalarPtr -> IO a) -> IO Scalar
 withNewScalar' = fmap snd . withNewScalar
 
+withScalarArray :: [Scalar] -> (ScalarArrayPtr -> IO a) -> IO a
```
tdammers, line 508: See `withPointArray`:
```suggestion
withScalarArray :: [Scalar] -> (Int -> ScalarArrayPtr -> IO a) -> IO a
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -428,6 +505,26 @@ withNewScalar_ = fmap fst . withNewScalar
...
+withScalarArray scalars go = do
+  let numScalars = length scalars
+      sizeReference = sizeOf (undefined :: Ptr ())
+  -- Allocate space for the scalars and a null terminator
```
tdammers, line 512: Why the null terminator? AFAICT, all code that consumes this API uses an explicit length to determine the end of the array; the null terminator is never used.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +993,72 @@ scalarCanonical scalar =
...
+blsMSM :: forall curve. BLS curve => Int -> [Integer] -> [Point curve] -> Point curve
+blsMSM threshold ss ps = unsafePerformIO $ do
+  zeroScalar <- scalarFromInteger 0
+  filteredPoints <-
+    foldM
+      ( \acc (s, pt) -> do
+          scalar <- scalarFromInteger s
+          -- here we filter out pairs that will not contribute to the result
+          -- This is also for safety, as the c_blst_to_affines C call
+          -- will fail if the input cointaints the point at infinity.
```
tdammers, line 1012:
```suggestion
          -- will fail if the input contains the point at infinity.
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +993,72 @@ scalarCanonical scalar =
...
+          -- We also filter out the zero scalar, as on windows builds,
+          -- the blst_mult_pippenger C call will fail for this case.
+          if not (blsIsInf pt) && scalar /= zeroScalar
+            then return ((scalar, pt) : acc)
+            else return acc
```
tdammers, line 1017: Reversing the logic will simplify this:
```suggestion
          if blsIsInf pt || scalar == zeroScalar
            then return acc
            else return ((scalar, pt) : acc)
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +993,72 @@ scalarCanonical scalar =
...
+      )
+      []
+      (zip ss ps)
+  case filteredPoints of
+    [] -> return blsZero
+    -- -- If there is only one point, we refert to blsMult function
+    -- -- The blst_mult_pippenger C call will also not work for
+    -- -- this case on windows builds.
```
tdammers, line 1025:
```suggestion
    -- If there is only one point, we refert to blsMult function
    -- The blst_mult_pippenger C call will also not work for
    -- this case on windows builds.
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -183,16 +193,36 @@ data Curve2
 
 ---- Unsafe PointPtr types
 
+-- A pointer to a (projective) point one of the two elliptical curves
 newtype PointPtr curve = PointPtr (Ptr Void)
 
+-- A pointer to a null-terminated array of pointers to points
```
tdammers, line 199:
```suggestion
-- | A pointer to a null-terminated array of pointers to points
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -183,16 +193,36 @@ data Curve2
...
+newtype PointArrayPtr curve = PointArrayPtr (Ptr Void)
+
 type Point1Ptr = PointPtr Curve1
 type Point2Ptr = PointPtr Curve2
 
+type Point1ArrayPtr = PointArrayPtr Curve1
+type Point2ArrayPtr = PointArrayPtr Curve2
+
+-- A pointer to an affine point on one of the two elliptical curves
```
tdammers, line 208:
```suggestion
-- | A pointer to an affine point on one of the two elliptical curves
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -183,16 +193,36 @@ data Curve2
...
 newtype AffinePtr curve = AffinePtr (Ptr Void)
 
+-- A pointer to a contiguous array of affine points
```
tdammers, line 211:
```suggestion
-- | A pointer to a contiguous array of affine points
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -183,16 +193,36 @@ data Curve2
...
+newtype AffineBlockPtr curve = AffineBlockPtr (Ptr Void)
+
+-- A pointer to a null-terminated array of pointers to affine points
```
tdammers, line 214:
```suggestion
-- | A pointer to a null-terminated array of pointers to affine points
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -288,6 +318,40 @@ withNewAffine_ = fmap fst . withNewAffine
...
+    let accumulate [] = do
+          poke (ptr `advancePtr` numPoints) nullPtr
+          go (PointArrayPtr (castPtr ptr))
+        accumulate ((ix, point) : rest) =
+          withPoint point $ \(PointPtr pPtr) -> do
+            poke (ptr `advancePtr` ix) pPtr
+            accumulate rest
+     in accumulate (zip [0 ..] points)
+
+-- | Given a block of affine points and a count, produce a null-terminated
```
tdammers, line 340: Like `withPointArray`, I don't think this should be null-terminated.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +993,72 @@ scalarCanonical scalar =
...
+-- Negative numbers will also be brought to the range
+-- [0, 'scalarPeriod' - 1] via modular reduction.
+blsMSM :: forall curve. BLS curve => Int -> [Integer] -> [Point curve] -> Point curve
+blsMSM threshold ss ps = unsafePerformIO $ do
+  zeroScalar <- scalarFromInteger 0
+  filteredPoints <-
+    foldM
+      ( \acc (s, pt) -> do
+          scalar <- scalarFromInteger s
+          -- here we filter out pairs that will not contribute to the result
```
tdammers, line 1010:
```suggestion
          -- Here we filter out pairs that will not contribute to the result.
```

tdammers (changes requested):
Looks mostly good to me.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
perturbing: Initially, I had it without, but the C code does use it (I discovered this by looking at the Rust implementation of the blst bindings [here](https://github.com/supranational/blst/blob/cbc7e166a10d7286b91a3a7bea341e708962db13/bindings/rust/src/pippenger.rs#L144)).

And to test it without, I ran [this](https://github.com/IntersectMBO/cardano-base/pull/514/commits/677279cc5f038ac548379d25f6fd722effe05df5) CI test, which [failed](https://github.com/IntersectMBO/cardano-base/actions/runs/14728349692/job/41336359225) on all windows machines.

perturbing: And to add to that, there are more things that are implicit in for the blst lib that I discovered. For example, for windows builds, given one pair for an unpiut (so one point + scalar), the msm call will also fail.

And to test it, see [this](https://github.com/IntersectMBO/cardano-base/pull/514/commits/de8c2558c9c143b7b52663a949d46dac1f0fb624) and [this](https://github.com/IntersectMBO/cardano-base/actions/runs/15020090010/job/42206982810) failure on all windows tests.

lehins (changes requested):
I had some questions and minor suggestions, but other than that it is looking good.

Needs a rebase on `master`. Also could you please add all visible changes to the Changelog and bump minor version in both cabal file and in [the latest section in the changlog file](https://github.com/IntersectMBO/cardano-base/blob/5d20803c1354d5f89cb3bc70c29f9981127bae2e/cardano-crypto-class/CHANGELOG.md?plain=1#L3) to `2.2.3.0`.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -875,6 +991,79 @@ scalarCanonical scalar =
...
+          --
+          --    pt ^ 0 = id
+          --
+          -- Which yields no contribution to summation, and
+          -- thus we can skip the point and scalar pair. This filter
+          -- saves us an extra input to the more expensive exponential
+          -- operation.
+          if blsIsInf pt || scalar == zeroScalar
+            then return acc
+            else return ((scalar, pt) : acc)
```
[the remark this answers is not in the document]
perturbing: Indeed, we need to modulo reduce the integer, as the mapping is not injective.

lehins (approved):
This PR looks ready to go. Thank you for the hard work @perturbing!

lehins:
@perturbing Could I ask you to cleanup the git commit history. Otherwise I am just gonna squash the whole 73 commits into one big one with very little info.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
perturbing: Thank you for digging into this! Ill clean up the commits after the weekend and bump the minor version + add to the changelog. Thanks for the help and the reviews!

PR #515, opened 2024-12-10
Move `cardano-mempool` to a separate repo.

lehins:
# Description

`cardano-mempool` was renamed to `memory-pool` in order to avoid confusion with an important concept of "mempool" used in consensus. Moreover it is a reusable package and it has no need to be tied to Cardano, so it got moved to a separate [reposity](https://github.com/input-output-hk/memory-pool) and [released to Hackage](https://hackage.haskell.org/package/memory-pool)

This PR is also a final thing that was needed for `cardano-crypto-class/tests` to have a fresh release to CHaP 

# Checklist

- [ ] Commits that only contain large amounts of formatting changes were added to `.git-blame-ignore-revs`
- [x] Self-reviewed the diff


Lucsanszky (approved):
:+1:

PR #519, opened 2025-02-27
Adding test vectors and golden tests from the standard for both versions of VRF

paweljakubas:
# Description

The PR aims to add test vectors (generated to used as etalon and also taken from two drafts version 03 (Praos) and version 13 (PraosCompatBatch)).

I added several exports and totally ortogonal and non-invasive impl in both impl of VRF - needed for testing. 

I used ReadP parser to avoid imports not that there are not better parser combinator libraries;-)

# Checklist

- [ ] Self-reviewed the diff


cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -59,8 +69,183 @@ tests =
...
+        , -- https://datatracker.ietf.org/doc/draft-irtf-cfrg-vrf/13/ - example 12
+          -- pi = 926e895d308f5e328e7aa159c06eddbe56d06846abf5d98c2512235eaa57fdce35b46edfc655bc828d44ad09d1150f31374e7ef73027e14760d42e77341fe05467bb286cc2c9d7fde29120a0b2320d04
+          testCase "generated golden test vector: vrf_ver13_standard_12" $
+            checkVer13TestVector "vrf_ver13_standard_12"
+        ]
     ]
 
+checkVer03TestVector :: FilePath -> Assertion
+checkVer03TestVector file = do
+  dir <- getCurrentDirectory
```
lehins, line 119: This is a totally incorrect way of getting to files that are included in a package.
What you need is the `Paths` interface provided by Cabal. See this section of documentation: https://cabal.readthedocs.io/en/3.4/cabal-package.html#accessing-data-files-from-package-code

For an example you can take a look at how we use it in ledger: 
* Add module to Cabal file: https://github.com/IntersectMBO/cardano-ledger/blob/f5b35692b4d92d664187781a98b7af3fab445bad/eras/conway/impl/cardano-ledger-conway.cabal#L148
* Import in [the test module](https://github.com/IntersectMBO/cardano-ledger/blob/f5b35692b4d92d664187781a98b7af3fab445bad/eras/conway/impl/huddle-cddl/Main.hs#L3-L8) and use `getDataFileName` to figure out the absolute path of the file.

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -59,8 +69,183 @@ tests =
...
+
+parserHex :: Maybe Int -> Parse.ReadP BS.ByteString
+parserHex lenM = do
+  str <- parseString
+  if str == "empty"
+    then
+      pure BS.empty
+    else case lenM of
+      Just len -> handleDecode str len
+      Nothing -> handleDecode str ((length str) `div` 2)
```
lehins, line 200: Redundant parenthesis
```suggestion
      Nothing -> handleDecode str (length str `div` 2)
```

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -59,8 +69,183 @@ tests =
...
+checkVer03TestVector :: FilePath -> Assertion
+checkVer03TestVector file = do
+  dir <- getCurrentDirectory
+  str <- readFile $ dir <> "/test_vectors/" <> file
+  let testVectorE = Read.readMaybe @VRFTestVector str
+  assertBool ("parsing test vector: " <> file <> " not successful") $ isJust testVectorE
+  let VRFTestVector {..} = fromJust testVectorE
+  signKey <- Ver03.skFromBytes signingKey
+  verKey <- Ver03.vkFromBytes verifyingKey
+  let bytesEq left right = case left of
```
lehins, line 126: "supplied" and "expected" names will give a better indication to what is going on.
Also, it will be useful to give this function a type signature with `HasCallStack`, so it will be clear which part of the test has failed.
```suggestion
  let bytesEq :: HasCallStack => Maybe ByteString -> ByteString -> Assertion
      bytesEq supplied expected = case left of
```

cardano-crypto-tests/test_vectors/vrf_ver03_generated_1
lehins, line 1: Make sure to list all of the test vector files in the cabal file [(see this example)]( https://github.com/IntersectMBO/cardano-ledger/blob/f5b35692b4d92d664187781a98b7af3fab445bad/eras/conway/impl/cardano-ledger-conway.cabal#L14-L17) 
It'll be easier if you give 'em some sort of extension, then you can use a wildcard, but it's up to you.

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -59,8 +69,183 @@ tests =
...
+        ]
     ]
 
+checkVer03TestVector :: FilePath -> Assertion
+checkVer03TestVector file = do
+  dir <- getCurrentDirectory
+  str <- readFile $ dir <> "/test_vectors/" <> file
+  let testVectorE = Read.readMaybe @VRFTestVector str
+  assertBool ("parsing test vector: " <> file <> " not successful") $ isJust testVectorE
+  let VRFTestVector {..} = fromJust testVectorE
```
lehins, line 123: This is a more idiomatic way of dealing with Maybe
```suggestion
  VRFTestVector {..} <- case testVectorE of
    Nothing -> assertFailure $ "Parsing test vector: " <> file <> " was not successful"
    Just testVector -> pure testVector
```

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -59,8 +69,183 @@ tests =
...
+
+data VRFTestVector = VRFTestVector
+  { name :: String
+  , version :: String
+  , ciphersuite :: String
+  , signingKey :: BS.ByteString
+  , verifyingKey :: BS.ByteString
+  , message :: BS.ByteString
+  , proof :: BS.ByteString
+  , hash :: BS.ByteString
```
lehins, line 183: I'd like to ask you to use camel case for `cipherSuite` and to use a common prefix for the fields:
```suggestion
  { testVectorName :: String
  , testVectorVersion :: String
  , testVectorCipherSuite :: String
  , testVectorSigningKey :: BS.ByteString
  , testVectorVerifyingKey :: BS.ByteString
  , testVectorMessage :: BS.ByteString
  , testVectorProof :: BS.ByteString
  , testVectorHash :: BS.ByteString
```

If `testVector` is too long of a prefix for your liking, I'd be ok with an abbreviation prefix of `tv`.

Without such prefix it is unclear where the binding is coming from (especially with RecordWildCards) and it can easily shadow binding with such a common name as `name` or `version` or `hash`. We want to avoid that

cardano-crypto-tests/src/Test/Crypto/VRF.hs
```diff
@@ -59,8 +69,183 @@ tests =
...
+  }
+
+data HexStringWithLength = HexStringWithLength
+  { payload :: String
+  , expectedLength :: Int
```
lehins, line 188: Ditto
```suggestion
  { hswlPayload :: String
  , hswlExpectedLength :: Int
```

lehins (changes requested):
Thank you for submitting this PR. I am always happy to have more tests.
Overall it looks good, but there are some issues I'd like to have resolved before merging.

I've never seen anyone use `ReadP` parser interface in such an anger before :grin:

cardano-crypto-praos/src/Cardano/Crypto/VRF/Praos.hs
```diff
@@ -376,6 +378,24 @@ vkFromBytes bs = do
...
+        ( "Invalid output length "
+            <> show @Int bsLen
+            <> ", expecting "
+            <> show @Int vrfKeySizeVRF
+        )
+    else do
+      output <- mkOutput
+      withForeignPtr (unOutput output) $ \ptr ->
+        copyFromByteString ptr bs vrfKeySizeVRF
+      return output
```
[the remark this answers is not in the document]
paweljakubas: sure! I was here "inspired" by already present implementations like `skFromBytes`, `vkFromBytes`,.. and follow the pattern. I agree with your remarks 100%

cardano-crypto-praos/src/Cardano/Crypto/VRF/PraosBatchCompat.hs
```diff
@@ -165,14 +173,17 @@ foreign import ccall "crypto_vrf_ietfdraft13_publickeybytes"
...
+  crypto_vrf_ietfdraft13_seedbytes :: CSize
 foreign import ccall "crypto_vrf_ietfdraft13_outputbytes"
   crypto_vrf_ietfdraft13_outputbytes :: CSize
 
 foreign import ccall "crypto_vrf_ietfdraft13_publickeybytes"
   io_crypto_vrf_ietfdraft13_publickeybytes :: IO CSize
 foreign import ccall "crypto_vrf_ietfdraft13_secretkeybytes"
   io_crypto_vrf_ietfdraft13_secretkeybytes :: IO CSize
+foreign import ccall "crypto_vrf_ietfdraft13_outputbytes"
+  io_crypto_vrf_ietfdraft13_outputbytes :: IO CSize
```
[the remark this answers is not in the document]
paweljakubas: 100% agree. Once again I was trying to be as minimally invasive as possible and followed "patterns" of `io_crypto_vrf_ietfdraft13_publickeybytes` and `io_crypto_vrf_ietfdraft13_secretkeybytes`.

cardano-crypto-praos/src/Cardano/Crypto/VRF/PraosBatchCompat.hs
```diff
@@ -402,6 +416,27 @@ mkOutput =
...
+            <> " or "
+            <> show @Int ioSize
+        )
+    else do
+      output <- mkOutput
+      withForeignPtr (unOutput output) $ \ptr ->
+        copyFromByteString ptr bs vrfKeySizeVRF
+      return output
+  where
+    bsLen = BS.length bs
```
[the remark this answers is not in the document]
paweljakubas: the same comment like for `Praos` impl. Of course applied your suggestion!

cardano-crypto-tests/src/Test/Crypto/VRF.hs
paweljakubas: good request.I did it and removed `directory` dependency. Thanks to that I have not added any dependency in this PR:-)

cardano-crypto-tests/src/Test/Crypto/VRF.hs
paweljakubas: done

cardano-crypto-tests/src/Test/Crypto/VRF.hs
paweljakubas: done, and extracted bytesEq and reused in both checks

cardano-crypto-tests/src/Test/Crypto/VRF.hs
paweljakubas: used `maybe` and followed the "idiomatic way"

cardano-crypto-tests/src/Test/Crypto/VRF.hs
paweljakubas: valid points. done!

paweljakubas:
Thanks @lehins for all the constructive remarks and `ReadP` nitpicking. I would use `attoparsec` if performance would be of issue. But as here it is only testing for being in line with standard test vectors and I don't want to clutter dependencies I opted for this simple, but good enough imho, parser combinator impl. Cheers

lehins (approved):
Beautiful. Thank you!

PR #525, opened 2025-04-09
Add SHA512 and SHA3_512 algorithms.

locallycompact:


locallycompact:
We are using these in an off-chain fork of plutus within the hydra-node. Would it be OK to include these in cardano-base?

locallycompact:
I'm not sure why the CI is failing

cardano-crypto-class/src/Cardano/Crypto/Hash/SHA3_512.hs
```diff
@@ -0,0 +1,23 @@
...
+where
+
+import Cardano.Crypto.Hash.Class
+import qualified Data.ByteArray as BA
+import qualified "crypton" Crypto.Hash as H
+
+data SHA3_512
+
+instance HashAlgorithm SHA3_512 where
+  type SizeHash SHA3_512 = 32
```
lehins, line 18: Size is not correct
```suggestion
  type SizeHash SHA3_512 = 64
```

cardano-crypto-class/src/Cardano/Crypto/Hash/SHA512.hs
```diff
@@ -0,0 +1,23 @@
...
+where
+
+import Cardano.Crypto.Hash.Class
+import qualified Data.ByteArray as BA
+import qualified "crypton" Crypto.Hash as H
+
+data SHA512
+
+instance HashAlgorithm SHA512 where
+  type SizeHash SHA512 = 32
```
lehins, line 18: Same here
```suggestion
  type SizeHash SHA512 = 64
```

lehins (changes requested):
Besides issues with sizes, there are a few other things that need to be done before we can merge this:
* Bump up version to `2.2.2.0`
* Add `2.2.2.0` section to the changelog and mention addition of `SHA512` and `SHA3_512` algorithms
* Add tests for those two new additions in here: https://github.com/IntersectMBO/cardano-base/blob/551356522e0c00a82f13b00ce30038cccb138b44/cardano-crypto-tests/src/Test/Crypto/Hash.hs#L38-L39
* Update the lower bound here to `^>= 2.2.2`: https://github.com/IntersectMBO/cardano-base/blob/551356522e0c00a82f13b00ce30038cccb138b44/cardano-crypto-tests/cardano-crypto-tests.cabal#L92 (due to new tests depending on those new algorithms
* Bump up minor version to `2.2.1.0` for `cardano-crypto-tests`
* Change this section of the changelog from `2.2.0.1` to `2.2.1.0` and a changelog entry about addition of the tests for those new algorithms

locallycompact:
@lehins re-review please

cardano-crypto-class/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for `cardano-crypto-class`
 
+## 2.2.2.0
+
+* Add `SHA512` and `SHA3_512` algorithms.
+
 ## 2.2.1.1
 
 *
```
lehins, line 9: Sorry, my bad. It should have replaced this section since `2.2.1.1` was never released
```suggestion
```

lehins (approved):
Looks great. Thank you.

locallycompact:
Ready

locallycompact:
Ready

lehins:
Thank you for this addition. Let me know of you need this change on CHaP, otherwise I'll wait until we be preparing next cardano-node release to release these two packages 

> I'm not sure why the CI is failing

Hydra has been broken for a while now. We are waiting until we can drop support for ghc-8.10, at which point we can unbreak hydra.

locallycompact:
Yes we do need it on CHaP please.

PR #528, opened 2025-04-11
Fix praos version

lehins:
# Description

Discovered some more issues from the #519 

This PR bumps up the version, removes unsafe exports and fixes the changelog

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-praos/src/Cardano/Crypto/VRF/PraosBatchCompat.hs
+++ b/cardano-crypto-praos/src/Cardano/Crypto/VRF/PraosBatchCompat.hs
@@ -40,7 +40,6 @@ module Cardano.Crypto.VRF.PraosBatchCompat (
   Seed,
   genSeed,
   keypairFromSeed,
-  seedFromBytes,
 
   -- * Conversions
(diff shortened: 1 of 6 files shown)


Lucsanszky (approved):
LGTM 👍

cardano-crypto-praos/src/Cardano/Crypto/VRF/PraosBatchCompat.hs
```diff
@@ -256,8 +255,8 @@ copyFromByteString ptr bs lenExpected =
 
 seedFromBytes :: ByteString -> Seed
```
teodanciu, line 256: No `HasCallStack`?  For any reason in particular?  out of curiosity

lehins: Cause there is only one use case in this module ;)
It was previously wrongfully exported, so I actually did add `HasCallStack` :smile: but then once I removed it from exports I removed the `HasCallStack` as well.

PR #534, opened 2025-05-16
Drop GHC support for <9.6.7

### LINKED ISSUE #530: Drop GHC-8.10 support
We had a successful release of cardano-node that uses ghc-9.6.7
It is time to drop ghc-8.10 support

aniketd:
# Description

Closes #530 

# Checklist

- [ ] ~~New tests are added if needed and existing tests are updated~~
- [ ] ~~All visible changes are prepended to the latest section of a `CHANGELOG.md` for the affected packages.
      **_New section is never added with the code changes._** (See [RELEASING.md](https://github.com/intersectmbo/cardano-base/blob/master/RELEASING.md#changelogmd))~~
- [x] Self-reviewed the diff

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,10 +3,17 @@
 
   inputs = {
     haskellNix.url = "github:input-output-hk/haskell.nix";
+    # allow us to independently update hackageNix
+    haskellNix.inputs.hackage.follows = "hackageNix";
 
     nixpkgs.follows = "haskellNix/nixpkgs-unstable";
     iohkNix.url = "github:input-output-hk/iohk-nix";
     flake-utils.url = "github:hamishmack/flake-utils/hkm/nested-hydraJobs";
+
+    hackageNix = {
+      url = "github:input-output-hk/hackage.nix";
+      flake = false;
+    };
 
     CHaP = {
       url = "github:intersectmbo/cardano-haskell-packages?ref=repo";
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "CHaP": {
       "flake": false,
       "locked": {
-        "lastModified": 1744135613,
-        "narHash": "sha256-ifKgooWGq45UYjweO0dOBE1rLBvf7tq/Co1Ob/8E8O0=",
+        "lastModified": 1749727790,
+        "narHash": "sha256-yCKoVpny9PSGTcSrCtXq77FuKFlq9Hf81ULVgCvkfmc=",
         "owner": "intersectmbo",
         "repo": "cardano-haskell-packages",
-        "rev": "143757ad1f38ddd9698696bd2e6c8ebb84eb2adf",
+        "rev": "b1b559392d6fc44626cef9d5b7821421cc7adcdc",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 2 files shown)


cabal.project
```diff
@@ -11,7 +11,7 @@ repository cardano-haskell-packages
 
 -- The hackage index-state
 index-state: 2025-05-14T16:20:57Z
-index-state: cardano-haskell-packages 2025-05-14T07:44:24Z
+index-state: cardano-haskell-packages 2025-04-08T17:22:48Z
```
aniketd, line 14: This change was necessary in order to be able to reuse the flake inputs that do not cause a problem on MacOS.

lehins: You should not be downgrading `index-state`

lehins (changes requested):
@aniketd I recommend you upgrade to latest `haskellNix` and if it gives you issues then reach out to someone on `#dx-pub`

Hacking around it by using an older state only delays the problem.

flake.nix
```diff
@@ -207,7 +201,7 @@
             };
           in
             mkDevShells cabalProject
-            # Additional shells for every GHC version supported by haskell.nix, eg. `nix develop .#ghc8107`
+            # Additional shells for every GHC version supported by haskell.nix, eg. `nix develop .#ghc8122`
```
neilmayhew, line 204:
```suggestion
            # Additional shells for every GHC version supported by haskell.nix, eg. `nix develop .#ghc9122`
```

cabal.project
aniketd: Thanks to @angerman 🙌 he found that removing the dependency on `ghcid` also removes the dependency on `hfsvnt` from `Cocoa` on MacOS. Since we do not use `ghcid` any more, this unblocks us. 👍

.github/workflows/haskell.yml
```diff
@@ -62,22 +62,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["8.10.7", "9.2.8", "9.6.6", "9.8.2", "9.10.1", "9.12.1"]
+        ghc: ["9.6", "9.8", "9.10", "9.12"]
```
lehins, line 65: We need to pin ghc versions exactly, it is not a good idea to allow CI to change versions from under you at an arbitrary point. We've always done it that way, so why did you decide to change it all of a sudden?
```suggestion
        ghc: ["9.6.7", "9.8.4", "9.10.2", "9.12.2"]
```

.github/workflows/haskell.yml
```diff
@@ -62,22 +62,18 @@ jobs:
...
         os: [ubuntu-latest, macos-13, macos-latest, windows-latest]
```
lehins, line 66: `macos-13` needs to be removed. It was only here for ghc-8.10
```suggestion
        os: [ubuntu-latest, macos-latest, windows-latest]
```

.github/workflows/haskell.yml
```diff
@@ -62,22 +62,18 @@ jobs:
...
-        # GHC versions older than ghc-9.2 are not supported on macos-latest
+        # GHC versions older than ghc-9.6 are not supported on macos-latest
```
lehins, line 67: Why do you think GHC versions older that ghc-9.6 all of a sudden stopped being supported by `macos-latest`?
If you read this comment carefully, you will realize that all of those `exclude`s can be removed, because they were here only because: "GHC versions older than ghc-9.2 are not supported on macos-latest"
So, please remove this whole section,

lehins (changes requested):
Recent changes totally messed up github actions.
Could you please fix them as soon as you get a chance

.github/workflows/haskell.yml
erikd: I actually do not agree.

GHC version `9.12.2` is only supposed to contain bug fixes from `ghc-9.12.1` . I, in the very UNLIKLEY event a but fix version does cause a regression, CI is the perfect place to find out about it.

.github/workflows/haskell.yml
aniketd: This is a grave oversight. We don't need to exclude anything now. :+1:

.github/workflows/haskell.yml
lehins: @erikd I appreciate your opinion, but CI suppose to be reproducible as much as possible.
That is why we use `index-state` and pin other dependencies.

> In the very UNLIKLEY event a bug fix version does cause a regression, CI is the perfect place to find out about it.

In this unlikely scenario we will still catch it, but only after manual update of ghc version on CI, which we usually update often.

I'd rather regressions like this be caught on a PR that updates the ghc version, then on a PR that changes the actual code in one of the packages in the repo.

I understand that every team has their own approach, but in cardano-ledger and in cardano-base this is the approach we take, so I'd appreciate if it was not changed because you are of a different opinion.

PR #543, opened 2025-09-01
Jj/strictmaybe instances

Soupstraw:
# Description

Adds `Read`, `Show1`, `Read1`,  `Eq1` and `Ord1` instances to `StrictMaybe`.

close https://github.com/IntersectMBO/cardano-base/issues/542

# Checklist

- [ ] Self-reviewed the diff


cardano-strict-containers/src/Data/Maybe/Strict.hs
```diff
@@ -138,3 +153,26 @@ instance Alternative StrictMaybe where
   empty = SNothing
   SNothing <|> r = r
   l <|> _ = l
+
+instance Eq1 StrictMaybe where
```
lehins, line 157: Could you please add a changelog for all new instances and bump up the minor version

lehins (approved):
Very nice! Thank you!

cardano-strict-containers/src/Data/Maybe/Strict.hs
Soupstraw: `1.5.0.0` does not seem to be released yet, so I didn't bump the minor version

PR #547, opened 2025-09-10
cardano-base: initialize with `CardanoFeatureFlag`

amesgen:
# Description

This PR adds a new library `cardano-base` that, for now, only contains a very simple module with a feature flag sum type. 

In the future, more functionality might be added/moved to `cardano-base` (potentially as a sublibrary), in contrast to the existing narrowly-scoped packages in this repo.

Addresses https://github.com/tweag/cardano-peras/issues/96

See eg https://github.com/IntersectMBO/ouroboros-network/commit/c330accfcd81e6d70d444e5e66de3205bb0f8838 for a potential real-world usage example

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-base/CHANGELOG.md
+++ b/cardano-base/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog for `cardano-base`
+
+## 0.1.0.0
+
+* Added `Cardano.Base.FeatureFlag`.
--- a/cardano-base/README.md
+++ b/cardano-base/README.md
@@ -1,0 +1,5 @@
+# `cardano-base`
+
+This package contains a collection of Cardano-related utilities.
+
+See the [Haddocks](https://intersectmbo.github.io/cardano-base/cardano-base/).
--- a/cardano-base/cardano-base.cabal
+++ b/cardano-base/cardano-base.cabal
@@ -1,0 +1,37 @@
+cabal-version: 3.0
+name: cardano-base
+version: 0.1.0.0
+synopsis: Various utilities for Cardano
+description: Various utilities for Cardano.
+category:
+  Cardano,
+
+copyright: 2025 Input Output Global Inc (IOG).
+author: IOHK
+maintainer: operations@iohk.io
+license: Apache-2.0
+license-files:
+  LICENSE
+  NOTICE
+
+extra-doc-files:
+  CHANGELOG.md
+  README.md
+
+common project-config
+  default-language: Haskell2010
+  build-depends: base >=4.18 && <5
+  ghc-options:
+    -Wall
+    -Wcompat
+    -Wpartial-fields
+    -Wredundant-constraints
+    -Wunused-packages
+
+library
+  import: project-config
+  hs-source-dirs: src
+  exposed-modules: Cardano.Base.FeatureFlags
+  build-depends:
+    aeson,
+    nothunks,
--- a/cardano-base/src/Cardano/Base/FeatureFlags.hs
+++ b/cardano-base/src/Cardano/Base/FeatureFlags.hs
@@ -1,0 +1,52 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | Mini-abstraction to aid parallel development of experimental Cardano
+-- features.
+--
+-- == Usage example
+--
+-- >>> import Data.Set (Set)
+-- >>> import qualified Data.Set as Set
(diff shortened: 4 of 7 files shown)


nfrisby:
This looks great to me; addressed all of the concerns I voiced during the call in which this idea was proposed.

Maybe add `Read` to the `deriving` clause, for easier QuickCheck debugging cycle.

cardano-base/src/Cardano/Base/FeatureFlags.hs
```diff
@@ -0,0 +1,35 @@
...
+-- | An experimental Cardano feature.
+data CardanoFeatureFlag
+  = -- | Feature flag for Ouroboros Peras (faster settlement).
+    PerasFlag
+  | -- | Feature flag for Ouroboros Leios (higher throughput).
+    LeiosFlag
+  | -- | Feature flag for Ouroboros Phalanx (anti-grinding).
+    PhalanxFlag
+  deriving stock (Show, Read, Eq, Ord, Generic)
+  deriving anyclass (NoThunks)
```
amesgen, line 35: Any further type classes to derive? Maybe `{From,To}JSON` (brings in an `aeson` dependency)?

lehins: Yeah, `aeson` is a fine dependency to add.

cardano-base/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog for `cardano-base`
+
+## Unreleased
```
lehins, line 3: We always have a version set in the changelog that will be the one that gets released
```suggestion
## 0.1.0.0
```

cardano-base/cardano-base.cabal
```diff
@@ -0,0 +1,36 @@
+cabal-version: 3.0
+name: cardano-base
+version: 0.0.0.0
```
lehins, line 3: Ditto
```suggestion
version: 0.1.0.0
```

cardano-base/cardano-base.cabal
```diff
@@ -0,0 +1,36 @@
...
+license-files:
+  LICENSE
+  NOTICE
+
+extra-doc-files:
+  CHANGELOG.md
+  README.md
+
+common project-config
+  default-language: GHC2021
```
lehins, line 22: This deviates from all other packages in cardano-base
```suggestion
```
I would not be opposed to go in this direction in the future, but for now consistency is more important

cardano-base/src/Cardano/Base/FeatureFlags.hs
```diff
@@ -0,0 +1,35 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DerivingStrategies #-}
+
+-- | Mini-abstraction to aid parallel development of experimental Cardano
+-- features.
+--
+-- == Usage example
+--
+-- >>> import Data.Set (Set)
```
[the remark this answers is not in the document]
lehins: Size of this feature set will be very small, I don't think it is worth adding micro optimizations like that

cardano-base/README.md
```diff
@@ -0,0 +1,7 @@
+# `cardano-base`
+
+This package contains a collection of Cardano-related utilities.
+
+## Feature flags
+
+See the [Haddocks](https://intersectmbo.github.io/cardano-base/cardano-base/Cardano-Base-FeatureFlags.html).
```
lehins, line 7: There wlll be a landing page, which is more future proof that pointing to a single module
```suggestion
See the [Haddocks](https://intersectmbo.github.io/cardano-base/cardano-base).
```

cardano-base/src/Cardano/Base/FeatureFlags.hs
```diff
@@ -0,0 +1,35 @@
...
+
+import GHC.Generics (Generic)
+import NoThunks.Class (NoThunks)
+
+-- | An experimental Cardano feature.
+data CardanoFeatureFlag
+  = -- | Feature flag for Ouroboros Peras (faster settlement).
+    PerasFlag
+  | -- | Feature flag for Ouroboros Leios (higher throughput).
+    LeiosFlag
```
lehins, line 31: This would be my preferred order, since it is alphabetical and also reflects the importance of the feature better. But, this is very subjective, so feel free to ignore this suggestion if you do not agree
```suggestion
  = -- | Feature flag for Ouroboros Leios (higher throughput).
    LeiosFlag
  | -- | Feature flag for Ouroboros Peras (faster settlement).
    PerasFlag
```

cardano-base/src/Cardano/Base/FeatureFlags.hs
lehins: > Enum/Bounded?

Yes, please!

lehins (approved):
Thank you!

PR #550, opened 2025-09-12
Upgrade contra-tracer

ana-pantilie:
# Description

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-tests/cardano-crypto-tests.cabal
+++ b/cardano-crypto-tests/cardano-crypto-tests.cabal
@@ -93,7 +93,7 @@ library
     cardano-crypto-praos >=2.2.1,
     cborg,
     containers,
-    contra-tracer ==0.1.0.2,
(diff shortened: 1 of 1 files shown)


ana-pantilie:
Closing in favor of https://github.com/IntersectMBO/cardano-haskell-packages/pull/1136

ana-pantilie:
> Closing in favor of https://github.com/IntersectMBO/cardano-haskell-packages/pull/1136

Nevermind, we should actually merge and adding it magically in a revision should be temporary, right?

lehins:
>  adding it magically in a revision should be temporary, right?

No, revision is permanent. We are changing the bounds in both places (on CHaP and here) in the same way, so the next time we make a release of cardano-crypto-test to CHaP with other changes, the change from this PR will already be on CHaP thanks to the revision.

PR #553, opened 2025-09-30
Release cardano-crypto-class

jasagredo:
Prepare release or `cardano-crypto-class` with Mempack 0.2

Also update CHaP after https://github.com/IntersectMBO/cardano-haskell-packages/commit/e0383257a045942d59ec4c301ec479bed7a14d63 to bring in revisions on `cardano-prelude`


cardano-crypto-class/CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 # Changelog for `cardano-crypto-class`
 
-## 2.2.3.1
+## 2.2.4.0
```
lehins, line 3: This is a patch version update. There are no semantic changes to the library itself

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -115,7 +115,7 @@ library
     io-classes >=1.4.1,
     memory,
     memory-pool,
-    mempack,
+    mempack >=0.2,
```
lehins, line 118: This constraint is not needed due to CPP that makes it compatible with mempack-0.1 as well. Such approach usually simplifies the transition

PR #587, opened 2025-12-08
Add BLS signature scheme to `DSIGN` interface

perturbing:
# Description

This extends and refactors the work done in #567 by @hjeljeli32.

This PR adds the two variants of the BLS12-381 signature scheme, `BLS12381MinVerKeyDSIGN` and `BLS12381MinSigDSIGN` to the already existing `DSIGN` class. Besides that, it also extends the class module to allow for aggregatable signature schemes, which both BLS variants are. It also includes tests and benchmark additions.

Though both BLS signature schemes work similarly, the difference is that the former signature scheme (`MinVerKey`) encodes the verification key in the smallest BLS group (`G1`), while having its signatures encoded in the bigger BLS group (`G2`). The latter (`MinSig`) does the reverse. Besides the size impact of the serialized objects, there is also a computational impact for signing/verification operations by this choice.

To play around with it, in a repl
```haskell
:set -XDataKinds -XTypeApplications -XOverloadedStrings -Wno-incomplete-uni-patterns
import qualified Data.ByteString as BS
import qualified Data.ByteString.Lazy as BL
import Data.Proxy (Proxy(..))
import Cardano.Binary
import Cardano.Crypto.DSIGN
import Cardano.Crypto.Seed
import Cardano.Crypto.Util (SignableRepresentation(..))

-- Setup
let seedLen = fromIntegral (seedSizeDSIGN (Proxy @BLS12381MinVerKeyDSIGN))
let seedBytes = BS.replicate seedLen 42
let seed = mkSeedFromBytes seedBytes
let skVer = genKeyDSIGN @BLS12381MinVerKeyDSIGN seed
let vkVer = deriveVerKeyDSIGN @BLS12381MinVerKeyDSIGN skVer
let simpleMsg = "Hello World" :: BS.ByteString

-- Basic use of new DSIGN interface
let emptyBLSContext = BLS12381SignContext Nothing Nothing
let sigVer = signDSIGN @BLS12381MinVerKeyDSIGN emptyBLSContext simpleMsg skVer
verifyDSIGN @BLS12381MinVerKeyDSIGN emptyBLSContext vkVer simpleMsg sigVer

-- More elaborate usage
let dstMainnet = Just "BLS_DST_CARDANO_V1" :: Maybe BS.ByteString
let augRandom = Just "BLS_AUG_MAINNET_V1" :: Maybe BS.ByteString
let ctxMainnet = BLS12381SignContext dstMainnet augRandom
let seedsBytes = map (BS.replicate seedLen) [0 .. 255]
let seeds = map mkSeedFromBytes seedsBytes
let sksVer = map (genKeyDSIGN @BLS12381MinVerKeyDSIGN) seeds
let vksVer = map (deriveVerKeyDSIGN @BLS12381MinVerKeyDSIGN) sksVer
let pops = map (createPossessionProofDSIGN @BLS12381MinVerKeyDSIGN ctxMainnet) sksVer
let msg = "some EB hash" :: BS.ByteString
let sigs = map (signDSIGN @BLS12381MinVerKeyDSIGN ctxMainnet msg) sksVer

-- Naive way of verifying without using aggregation property 
-- of signature scheme
-- 
-- 1) check that all vk's have valid PoPs
-- 2) check each signature without aggregation
sequence_ $ map (\(vk,pop) -> verifyPossessionProofDSIGN @BLS12381MinVerKeyDSIGN ctxMainnet vk pop) (zip vksVer pops)
sequence_ $ map (\(vk,sig) -> verifyDSIGN @BLS12381MinVerKeyDSIGN ctxMainnet vk msg sig) (zip vksVer sigs)

-- In the context of Leios, EB votes from each spo contain/map to a 
-- signature and verification key pair. Also, note that in this context, 
-- SPO register their bls keys in advance, at which time a PoP 
-- check is done. When constructing an EB certificate, 
-- it is the aggregation below that compresses all these signatures 
-- in one (together with who participated in signing)
let Right aggrSig = aggregateSigsDSIGN @BLS12381MinVerKeyDSIGN sigs
let Right aggrVer = uncheckedAggregateVerKeysDSIGN @BLS12381MinVerKeyDSIGN vksVer
-- Then any observer of that certificate can verifying the 
-- group signature against the many signers via
verifyDSIGN @BLS12381MinVerKeyDSIGN ctxMainnet aggrVer msg aggrSig
```
Note that this signature scheme introduces five things that a "normal" `DSIGN` interface does not have.

1. The `DSIGN` class is extended to optionally generate keys with extra key info (`KeyGenContext v`). This is to comply with the key generation as per [this](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#name-keygen) IETF draft. This change is backwards compatible with the other existing `DSIGN` signature schemes.
```ghci
ghci> genKeyDSIGN @BLS12381MinVerKeyDSIGN seed == genKeyDSIGNWithKeyInfo @BLS12381MinVerKeyDSIGN (Just "") seed
True
ghci> genKeyDSIGN @BLS12381MinSigDSIGN seed == genKeyDSIGNWithKeyInfo @BLS12381MinSigDSIGN (Just "") seed
True
ghci> genKeyDSIGNWithKeyInfo @Ed25519DSIGN () seed == genKeyDSIGN @Ed25519DSIGN seed
True
```
2. This signature scheme has a signing context different from `()`. More explicitly, for the domain separation tag and augmentation tag, we have that
```ghci
ghci> let sigVer2 = signDSIGN @BLS12381MinVerKeyDSIGN (BLS12381SignContext (Just "") (Just "")) simpleMsg skVer
ghci> sigVer2 == sigVer
True
```

3. The `rawSerialise{SigDSIGN, VerKeyDSIGN}` functions also compress their objects (as per [the](https://github.com/zcash/librustzcash/blob/6e0364cd42a2b3d2b958a54771ef51a8db79dd29/pairing/src/bls12_381/README.md#serialization) ZCash standard). That is, e.g.,
```ghci
ghci> vkVer 
VerKeyBLS12381 "5326f72f278230d9c89c1d2fc5f0027af4fd3b354c1c33429c729b1c41c02009e139b96c47e909d4cd160fc6218d3518aa6d098550dbec14b9383fa642dd58b4c046238ebef93ba5b8c4c3963a66f450f2dba5a19f5a98d0f51fc754b9a35315fdff02000000097602000cc40b00f4ebba58c7535798485f455752705358ce776dec56a2971a075c93e480fac35ef615"
ghci> BS.length $ rawSerialiseVerKeyDSIGN @BLS12381MinVerKeyDSIGN vkVer
48
```
4. The `rawDeserialise{SigDSIGN, SignKeyDSIGN VerKeyDSIGN}` also performs a check on whether the curve points (for a `VerKeyDSIGN` and `SigDSIGN`) and the raw scalar input (for the `SignKeyDSIGN`) are non-zero and reduces it to the field (for the safety of the other FFI calls that might follow on it). E.g.,
```ghci
ghci> let skBsZeros = BS.replicate 32 0
ghci> let skBsOnes = BS.replicate 32 255
ghci> rawDeserialiseSignKeyDSIGN @BLS12381MinVerKeyDSIGN skBsZeros
Nothing
ghci> rawDeserialiseSignKeyDSIGN @BLS12381MinVerKeyDSIGN skBsOnes 
Just (SignKeyBLS12381 "fdffffff0100000002480300fab78458f54fbcecef4f8c996f05c5ac59b12418")
```

5. This PR extends the `DSIGNAlgorithm v` interface with `DSIGNAggregatable v`, which allows for the utility functions that relate to aggregatable signature schemes like aggregated signature verification and the creation and verification of Proof of Possessions (PoP) of verification keys.

For others interested in this work and that want to run the benchmarks
```bash
 cabal run cardano-crypto-class:bench --  --match pattern BLS12381
 ```

# Integrator-facing documentation (proposal)
I couldn’t find a concise integrator-facing documentation for DSIGN + BLS. BLS has a few “extra moving parts” compared to other DSIGN schemes: two variants (min-sig vs min-verkey) with different artifact sizes/perf tradeoffs, PoP + aggregation workflows, and a signing context via (DST, AUG).

Proposal: add a short “BLS DSIGN Integrator Guide” section under cardano-crypto-class/README.md (or a more appropriate location), covering:
- Variant choice: BLS12381MinSigDSIGN vs BLS12381MinVerKeyDSIGN (sizes + typical tradeoff).
- ContextDSIGN: what DST/AUG mean, and recommended usage pattern.
- PoP/aggregation: expected flow for registrars/aggregators/validators.
- Minimal code snippet for major operations (sign/verify, aggregate/verify, etc) and key/sig/PoP compressed sizes + serialization/deserialization failure notes.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-class/bench/Bench/Crypto/DSIGN.hs
+++ b/cardano-crypto-class/bench/Bench/Crypto/DSIGN.hs
@@ -134,7 +134,7 @@ benchAggDSIGN :: forall v a
 benchAggDSIGN _ lbl =
   let msg = exampleSignable (Proxy @v)
       ctx = exampleContext (Proxy @v)
-      ns = 1 : [100, 200 .. 1000]
+      ns = 1 : [100, 200, 300]
   in bgroup (lbl <> "/Aggregatable") $
       [ bgroup ("n=" <> show n)
           [ env (pure (mkCase @v ctx msg n)) $ \c ->

### CODE REVIEW

File: cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs [hunk 1]
```diff
@@ -0,0 +1,358 @@
...
+    deriving anyclass (NoThunks)
+  algorithmNameDSIGN _ = "bls12-381-" ++ symbolVal (Proxy @(CurveVariant curve))
+  {-# NOINLINE deriveVerKeyDSIGN #-}
+  deriveVerKeyDSIGN (SignKeyBLS12381 skPsb) = do
+    VerKeyBLS12381 $ unsafeDupablePerformIO . psbUseAsCPtr skPsb $ \skp ->
+      psbCreate $ \vkPtp ->
+        c_blst_sk_to_pk @curve (PointPtr vkPtp) (ScalarPtr skp)
+  {-# NOINLINE signDSIGN #-}
+  signDSIGN ctx@(dst, aug) msg (SignKeyBLS12381 skPsb) =
+    warnIfNoDst ctx $ SigBLS12381 $ unsafeDupablePerformIO . psbCreate $ \sigPts -> do
```
perturbing: @lehins, wdyt of doing something like this `warnIfNoDst`? Is there a nice way of doing this? Similarly for `proveProofOfPossesion` down below.

As the warning mentions, it is really unsafe to not use a domain separation tag. That said, the IETF draft assumes no, so we do need to support it to be compatible. Of course we can also just add a comment about it, but I wonder if there is more to push this in the face of the developer.

Thanks!

hjeljeli32:
Dear @perturbing and @curiecrypt I added in the description a section about the integrator-facing documentation that I think it is worth to add. I proposed a structure of what we can include in such documentation. Please feel free to modify/add things to it.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
```diff
@@ -455,3 +469,136 @@ decodeSignKeyDSIGNM = do
...
+  -- Note that the signing context is passed since the PoP might depend on it.
+  aggregateVerKeysDSIGN ::
+    HasCallStack =>
+    ContextDSIGN v ->
+    [(VerKeyDSIGN v, ProofOfPossessionDSIGN v)] ->
+    Either String (VerKeyDSIGN v)
+  aggregateVerKeysDSIGN ctx verKeysAndPoPs = do
+    -- Verify every verKey and its PoP (fail-fast)
+    forM_ verKeysAndPoPs $ uncurry (verifyProofOfPossessionDSIGN ctx)
+    aggregateVerKeysDSIGNWithoutPoPs (map fst verKeysAndPoPs)
```
ch1bo: Alternatively, should we be more explicit about `aggregateVerKeysDSIGNWithoutPoPs` being unsafe (used alone) by renaming it `unsafeAggregateVerKeysDSIGN` and instead explain what makes it unsafe (no PoPs) in the haddock?

```suggestion
    unsafeAggregateVerKeysDSIGN (map fst verKeysAndPoPs)
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -753,6 +809,34 @@ foreign import ccall "blst_fp12_finalverify" c_blst_fp12_finalverify :: PTPtr ->
...
+    CSize ->
+    Ptr CChar ->
+    CSize ->
+    Ptr CChar ->
+    CSize ->
+    IO CInt
+
 ---- Raw BLST error constants
 
 foreign import ccall "cardano_blst_success" c_blst_success :: CInt
```
ch1bo: Likely out of scope of this PR: The [`capi` calling convention](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/ffi.html#the-capi-calling-convention) would allow to import constants (or even pragmas) directly.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+import GHC.TypeLits (KnownSymbol, Symbol, symbolVal)
+import GHC.TypeNats (KnownNat, type (+))
+import NoThunks.Class (NoThunks)
+import System.IO.Unsafe (unsafeDupablePerformIO)
+
+data BLS12381DSIGN curve
+
+-- Making sure different 'Signature schemes are not 'Coercible', which would ruin the
+-- intended type safety:
+type role BLS12381DSIGN nominal
```
ch1bo: This is interesting, I did not know about this!

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+-- we get a canonical representative of the equivalence class that might differ!
+-- That is, from compressed (x,y) we get back [x:y,0].
+instance
+  BLS (Dual curve) =>
+  Eq (SigDSIGN (BLS12381DSIGN curve))
+  where
+  SigBLS12381 s1 == SigBLS12381 s2 =
+    let p1 = Point @(Dual curve) s1
+        p2 = Point @(Dual curve) s2
+     in p1 == p2
```
ch1bo: Nitpick: These two instances would make slightly more sense further below (after we defined associated types `VerKeyDSIGN (BLS12381DSIGN curve)` and `SigDSIGN (BLS12381DSIGN curve)`. So when reading this code top down, I had to scroll down to read up on the definition of the keys first (unsurprisingly its just wrapped bytes, so not really a big deal).

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+  {-# NOINLINE rawDeserialiseSignKeyDSIGN #-}
+  rawDeserialiseSignKeyDSIGN bs =
+    -- A signing key is strictly a BE integer mod the curve order.
+    -- The `DSIGN` interface via PSB would ensure at the type level that
+    -- they are of size 32 bytes (256 bits). But we must even ensure
+    -- they are valid Scalars, i.e., less than the curve order (255 bits).
+    case scalarFromBS bs of
+      Left _ -> Nothing
+      Right (Scalar skPsb) -> Just (SignKeyBLS12381 skPsb)
+  rawDeserialiseSigDSIGN bs =
```
ch1bo: Should: `NOINLINE` this too?

I only spotted this after I introduced newlines between definitions.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+    SignKeyBLS12381 $
+      let (bs, _) = getBytesFromSeedT (seedSizeDSIGN (Proxy @(BLS12381DSIGN curve))) seed
+       in unsafeDupablePerformIO $ do
+            withMaybeCStringLen keyInfo $ \(infoPtr, infoLen) ->
+              unsafeUseAsCStringLen bs $ \(ikmPtr, ikmLen) ->
+                psbCreate $ \skp ->
+                  c_blst_keygen (ScalarPtr skp) ikmPtr (fromIntegral ikmLen) infoPtr (fromIntegral infoLen)
+
+  -- \| Note that this also compresses the signature according to the ZCash standard
+  rawSerialiseSigDSIGN (SigBLS12381 sigPSB) = blsCompress @(Dual curve) (Point sigPSB)
```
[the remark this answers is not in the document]
ch1bo: Or.. why are these methods all marked with `NOINLINE` in the first place? I can see why for the definitions that use `unsafeDupablePerformIO`, but not with these here below.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+            psbUseAsCPtr skPsb $ \skPtp -> do
+              withNewPoint_ @(Dual curve) $ \hashPtr -> do
+                c_blst_hash @(Dual curve)
+                  hashPtr
+                  msgPtr
+                  (fromIntegral msgLen)
+                  dstPtr
+                  (fromIntegral dstLen)
+                  augPtr
+                  (fromIntegral augLen)
```
ch1bo: Could: use `truncate` to be more explicit that this might truncate (it does not with `Int` and `CSize` afaik)

```suggestion
                  (truncate msgLen)
                  dstPtr
                  (truncate dstLen)
                  augPtr
                  (truncate augLen)
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+            withMaybeCStringLen keyInfo $ \(infoPtr, infoLen) ->
+              unsafeUseAsCStringLen bs $ \(ikmPtr, ikmLen) ->
+                psbCreate $ \skp ->
+                  c_blst_keygen (ScalarPtr skp) ikmPtr (fromIntegral ikmLen) infoPtr (fromIntegral infoLen)
+
+  -- \| Note that this also compresses the signature according to the ZCash standard
+  rawSerialiseSigDSIGN (SigBLS12381 sigPSB) = blsCompress @(Dual curve) (Point sigPSB)
+  {-# NOINLINE rawSerialiseVerKeyDSIGN #-}
+  -- \| Note that this also compresses the verification key according to the ZCash standard
+  rawSerialiseVerKeyDSIGN (VerKeyBLS12381 vkPSB) = blsCompress @curve (Point vkPSB)
```
ch1bo: What does "also" mean here? I see it is calling `blst_compress` on the verification key bytes. The documentation makes me think it there should maybe be more? Also, how is zcash related to all this? (is `libblst` coming from that project?)

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+      let VerKeyBLS12381 vkPsb =
+            deriveVerKeyDSIGN (SignKeyBLS12381 skPsb) ::
+              VerKeyDSIGN (BLS12381DSIGN curve)
+          vk = blsCompress @curve (Point vkPsb)
+          Point mu1Psb =
+            blsMult (blsHash @(Dual curve) vk dst aug) skAsInteger
+          Point mu2Psb =
+            blsMult (blsGenerator @(Dual curve)) skAsInteger
+      return $ ProofOfPossessionBLS12381 mu1Psb mu2Psb
+  verifyProofOfPossessionDSIGN (dst, aug) (VerKeyBLS12381 vkPsb) (ProofOfPossessionBLS12381 mu1Psb mu2Psb) =
```
ch1bo: Similarly here, we seemingly always need the points `mu1` and `mu2` of a PoP. So a function

```haskell
popToPoints :: ProofOfPossessionDSIGN (BLS12381DSIGN curve) -> (Point (Dual curve), Point (Dual curve))
```

or even changing the definition of `ProofOfPossessionDSIGN` to

```haskell
  data ProofOfPossessionDSIGN (BLS12381DSIGN curve)
    = ProofOfPossessionBLS12381
    { mu1 :: !(Point (Dual curve))
    , mu2 :: !(Point (Dual curve))
    }
```

would allow us an easy access to `ProofOfPossessionBLS12381{mu1, mu2}` while construction would need to just wrap it into `Point`.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+     in if check1 && check2
+          then Right ()
+          else Left "ProofOfPossessionDSIGN BLS12381DSIGN failed to verify."
+  rawSerialiseProofOfPossessionDSIGN (ProofOfPossessionBLS12381 mu1Psb mu2Psb) =
+    blsCompress @(Dual curve) (Point mu1Psb) <> blsCompress @(Dual curve) (Point mu2Psb)
+  rawDeserialiseProofOfPossessionDSIGN bs =
+    -- We use the compressed size of a point in Dual curve to split
+    let chunkSize = compressedSizePoint (Proxy @(Dual curve))
+     in case splitsAt [chunkSize] bs of
+          mu1Bs : mu2Bs : _ ->
```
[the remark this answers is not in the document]
ch1bo: Nevermind, there is a test (`"PoP deserialization (wrong length)"`) that proves this is already behaving this way.

The edit may still make sense to improve readability.

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
```diff
@@ -680,3 +738,114 @@ prop_dsignm_verify_neg_msg lock _ a a' =
...
+  -> Gen (ContextDSIGN v)
+  -> Gen Message
+  -> String
+  -> Spec
+testDSIGNAggregatableWithContext _ genContext genMsg name = testEnough . describe name $ do
+  describe "serialization" $ do
+    describe "raw" $ do
+      prop "PoP serialization" .
+        forAllShow (defaultProofOfPossessionGen @v genContext)
+                   ppShow $
```
ch1bo: Could: DRY this with a `forAllPoP = forAllShow (defaultProofOfPossessionGen @v genContext) ppShow` (9 times in this module)

cardano-crypto-class/testlib/Test/Crypto/Util.hs
```diff
@@ -213,6 +215,34 @@ instance Arbitrary Message where
...
+    | dst' <- shrinkMaybeByteString dst
+    ]
+      ++ [ BLS12381SignContext dst aug'
+         | aug' <- shrinkMaybeByteString aug
+         ]
+    where
+      shrinkMaybeByteString :: Maybe ByteString -> [Maybe ByteString]
+      shrinkMaybeByteString Nothing = []
+      shrinkMaybeByteString (Just bs) =
+        Nothing : map (Just . BS.pack) (shrink (BS.unpack bs))
```
ch1bo: Could: just define a `blsSigContextGen :: Gen (ContextDSIGN (BLS12381DSIGN curve))` (with `type ContextDSIGN (BLS12381DSIGN curve) = (Maybe ByteString, Maybe ByteString)`)

The shrinker does not seem to be used and the two call sites of `arbitrary :: Gen BLS12381SignContext` could just be the function above. The name data type is of no use otherwise. If you think differently, I'd like to challenge that with using the data type in the implementation then (where we are currently contempt with the type alias to a tuple)

ch1bo (approved):
This PR is very thorough and I could not find correctness issues (I'm not a cryptographer though!)

There are a few possible improvements, mostly about naming and readability. Most important _could_ be to keep `Point` values directly in the associated types for keys and pop. This could simplify the code without changing semantics.

The one thing I thought was wrong (a "Must" change) was in fact correct and already covered by a test - great!

The only thing that _should_ be considered is my question about `NOINLINE` of some functions.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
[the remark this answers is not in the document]
lehins: I concur with everything @ch1bo suggested here. I'd also recommend that `AggregatableVerKey` has its field be strict, instead of lazy, to avoid space leaks.

Moreover, is there a potential algorithm in the future that would have a different implementation? If no, then it should be extracted into a standalone function instead of making it a type class function with default implementation.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
lehins: `truncate` can't be used on an `Int`, since it doesn't have `RealFrac` instance. 

That being said, using `fromIntegral` like this is dangerous and explicit type signatures are required in order to future proof against type changes:
```suggestion
                  (fromIntegral @Int @CSize msgLen)
                  dstPtr
                  (fromIntegral @Int @CSize dstLen)
                  augPtr
                  (fromIntegral @Int @CSize augLen)
```

This suggestion need to be applied to **all** uses of `fromIntegral` in this repo and I would suggest to make it a habit in any codebase.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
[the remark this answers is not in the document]
perturbing: Perhaps `BLS12_381` is better, since it is more consistent with this module?

https://github.com/IntersectMBO/cardano-base/blob/6974b7053907594becec12ba49b2352c9b73b42f/cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs#L11

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
perturbing: Credits to the implementor of the other `BLS12_381` bindings implementor.

https://github.com/IntersectMBO/cardano-base/blob/6974b7053907594becec12ba49b2352c9b73b42f/cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs#L260

Similarly, the underlying bytes of the curve point are really a different representation under the hood. I also did not know this :wink:

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+--
+-- > Make sure that the either you switch off let-floating
+-- > (-fno-full-laziness), or that the call to unsafePerformIO cannot float
+-- > outside a lambda.
+--
+-- If we do not switch off let-floating, our calls to unsafeDupablePerformIO for
+-- FFI functions become nondeterministic in their behaviour when run with
+-- parallelism enabled (such as -with-rtsopts=-N), possibly yielding wrong
+-- answers on a range of tasks, including serialization.
+{-# OPTIONS_GHC -fno-full-laziness #-}
```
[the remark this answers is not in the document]
perturbing: Credit to the implementor of the other DSIGN interfaces, where I took this from

https://github.com/IntersectMBO/cardano-base/blob/6974b7053907594becec12ba49b2352c9b73b42f/cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs#L10

I think this is added because the node might do some optimization via parallelism for signatures (@lehins). Note that up until now, the BLS code was only used for plutus, which disallows such optimization, I believe. @kwxm can you confirm this?

If we now extend this, I think the BLS module then should indeed also have this. Nice one!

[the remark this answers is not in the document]
lehins: This flag `-fno-full-laziness` is not needed at all. In fact, there are no places in the repo where this is needed, because:
* that flag does not really save us from floating out, as the documentation promises. See the discussion on this GHC ticket: https://gitlab.haskell.org/ghc/ghc/-/issues/12656#comment:9
* We do not use any thaw like operations or any other IO operations that could be impacted by the full-laziness in an unsafe manner.

It is absolutely safe to use `unsafePerformIO` or `unsafeDupablePerformIO` in almost all cases that we have in `cardano-crypto-class`. The only places where we cannot use `unsafeDupablePerformIO` is with mlocked memory that uses `bracket`, which is only used by the kes-agent and not cardano-node itself.

We also do not need to mark those memory/crypto operations as `NOINLINE`, they are absolutely safe to `INLINE` and in fact it is desired, since it will have a very positive impact on performance.

The most important part why I am saying it is safe to use all of those unsafe functions, INLINE and not require a ghc flad is because all we do in crypto operations is allocate memory and never mutate it again. All of the calls to C over FFI are done while those buffers are mutable.

It would be unsafe to use with `unsafeInlineIO`/`accursedUnutterablePerformIO` in almost of those places, that is why there is only one place in this repo where we use it and it is indeed safe to use.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
lehins: Yes this is important, otherwise `coerce` will happily mess everything up.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
```diff
@@ -455,3 +469,138 @@ decodeSignKeyDSIGNM = do
...
+  ( DSIGNAlgorithm v
+  , Show (ProofOfPossessionDSIGN v)
+  , Eq (ProofOfPossessionDSIGN v)
+  , NoThunks (ProofOfPossessionDSIGN v)
+  , KnownNat (SizeProofOfPossessionDSIGN v)
+  ) =>
+  DSIGNAggregatable v
+  where
+  type SizeProofOfPossessionDSIGN v :: Nat
+  data ProofOfPossessionDSIGN v :: Type
```
lehins: Call we call it `PossessionProof`, instead of `ProofOfPossession`?
```suggestion
  type PossessionProofSizeDSIGN v :: Nat
  data PossessionProofDSIGN v :: Type
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -198,6 +210,11 @@ import System.IO.Unsafe (unsafePerformIO)
 data Curve1
 data Curve2
 
+-- | A type family mapping a curve to its dual curve (its an involution).
+type family Dual curve where
```
lehins: I would like to suggest a different name in order to prevent confusion with `Dual` type(s) from `base`
How about `DualCurve` or something along those lines?

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+                  (fromIntegral msgLen)
+                  dstPtr
+                  (fromIntegral dstLen)
+                  augPtr
+                  (fromIntegral augLen)
+                c_blst_sign @curve
+                  (Proxy @curve)
+                  (PointPtr sigPts)
+                  hashPtr
+                  (ScalarPtr skPtp)
```
lehins: Type application is redundant. That is what `Proxy` is for.
```suggestion
                c_blst_sign (Proxy @curve) (PointPtr sigPts) hashPtr (ScalarPtr skPtp)
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+
+  {-# NOINLINE verifyDSIGN #-}
```
lehins:
```suggestion
  {-# INLINE verifyDSIGN #-}
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -434,6 +456,20 @@ class (KnownNat (PointSize curve), KnownNat (AffineSize curve)) => BLS curve whe
   sizeAffine_ :: Proxy curve -> CSize
   sizeAffine_ _ = fromInteger (natVal (Proxy @(AffineSize curve)))
 
+  c_blst_sk_to_pk :: PointPtr curve -> ScalarPtr -> IO ()
+  c_blst_sign :: Proxy curve -> PointPtr (Dual curve) -> PointPtr (Dual curve) -> ScalarPtr -> IO ()
```
lehins: Not needed thanks to `TypeFamilyDependencies` suggestion
```suggestion
  c_blst_sign :: PointPtr (Dual curve) -> PointPtr (Dual curve) -> ScalarPtr -> IO ()
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+
+  algorithmNameDSIGN _ = "bls12-381-" ++ symbolVal (Proxy @(CurveVariant curve))
+
+  {-# NOINLINE deriveVerKeyDSIGN #-}
+  deriveVerKeyDSIGN (SignKeyBLS12381 skPsb) = do
+    VerKeyBLS12381 $ unsafeDupablePerformIO . psbUseAsCPtr skPsb $ \skp ->
+      psbCreate $ \vkPtp ->
+        c_blst_sk_to_pk @curve (PointPtr vkPtp) (ScalarPtr skp)
+
+  {-# NOINLINE signDSIGN #-}
```
lehins:
```suggestion
  {-# INLINE signDSIGN #-}
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+                  (Proxy @curve)
+                  (PointPtr sigPts)
+                  hashPtr
+                  (ScalarPtr skPtp)
+
+  {-# NOINLINE verifyDSIGN #-}
+  -- \| Context can hold domain seperation tag and/or augmentation data for signatures
+  verifyDSIGN (dst, aug) (VerKeyBLS12381 pbPsb) msg (SigBLS12381 sigPsb) =
+    unsafeDupablePerformIO $ do
+      withMaybeCStringLen dst $ \(pDst, dstLen) -> do
```
lehins: For consistency sake
```suggestion
      withMaybeCStringLen dst $ \(dstPtr, dstLen) -> do
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+        withAffine (toAffine @curve (Point pbPsb)) $ \pkAff ->
+          withAffine (toAffine @(Dual curve) (Point sigPsb)) $ \sigAff ->
+            withMaybeCStringLen aug $ \(pAug, augLen) ->
+              unsafeUseAsCStringLen (getSignableRepresentation msg) $ \(pMsg, msgLen) -> do
```
lehins:
```suggestion
              unsafeUseAsCStringLen (getSignableRepresentation msg) $ \(msgPtr, msgLen) -> do
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+                  c_blst_core_verify @curve
+                    pkAff
+                    sigAff
+                    True
+                    pMsg
+                    (fromIntegral msgLen)
+                    pDst
+                    (fromIntegral dstLen)
+                    pAug
+                    (fromIntegral augLen)
```
lehins: Ditto
```suggestion
                    msgPtr
                    (fromIntegral @Int @CSize msgLen)
                    dstPtr
                    (fromIntegral @Int @CSize dstLen)
                    augPtr
                    (fromIntegral @Int @CSize augLen)
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
```diff
@@ -162,6 +173,9 @@ class
   -- is at least 'seedSizeDSIGN' bytes long.
   genKeyDSIGN :: Seed -> SignKeyDSIGN v
 
+  genKeyDSIGNWithKeyInfo :: KeyGenContextDSIGN v -> Seed -> SignKeyDSIGN v
+  genKeyDSIGNWithKeyInfo _ = genKeyDSIGN
```
lehins: Is it useful to add this function and `KeyGenContextDSIGN` to `DSIGNAlgorithm` if none of the `DSIGN` algorithms except BLS can use it?
Should they go into `DSIGNAggregatable` instead?

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+  genKeyDSIGNWithKeyInfo keyInfo seed =
+    SignKeyBLS12381 $
+      let (bs, _) = getBytesFromSeedT (seedSizeDSIGN (Proxy @(BLS12381DSIGN curve))) seed
+       in unsafeDupablePerformIO $ do
+            withMaybeCStringLen keyInfo $ \(infoPtr, infoLen) ->
+              unsafeUseAsCStringLen bs $ \(ikmPtr, ikmLen) ->
+                psbCreate $ \skp ->
+                  c_blst_keygen (ScalarPtr skp) ikmPtr (fromIntegral ikmLen) infoPtr (fromIntegral infoLen)
+
+  -- \| Note that this also compresses the signature according to the ZCash standard
```
lehins:
```suggestion
  -- Note that this also compresses the signature according to the ZCash standard
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+    -- We use the compressed size of a point in Dual curve to split
+    let chunkSize = compressedSizePoint (Proxy @(Dual curve))
+     in case splitsAt [chunkSize] bs of
+          mu1Bs : mu2Bs : _ ->
+            -- Note that these also performs group membership checks
+            case ( blsUncompress @(Dual curve) mu1Bs
+                 , blsUncompress @(Dual curve) mu2Bs
+                 ) of
+              (Right (Point mu1Psb), Right (Point mu2Psb)) ->
+                Just $ ProofOfPossessionBLS12381 mu1Psb mu2Psb
```
lehins: Instead of using an ugly case statement on a tuple there is a neat trick that allows using `Maybe`'s `MonadFail` instance to short circuit on a wrong pattern match.
```suggestion
            -- Note that these also perform group membership and size checks.
            -- It will also ensure that all of the supplied `ByteString` is consumed
            -- through the size checks.
            Right (Point mu1Psb) <- pure $ blsUncompress @(Dual curve) mu1Bs
            Right (Point mu2Psb) <- pure $ blsUncompress @(Dual curve) mu2Bs
            Just $ ProofOfPossessionBLS12381 mu1Psb mu2Psb
```

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -253,6 +270,11 @@ type family PointSize curve where
   PointSize Curve1 = CARDANO_BLST_P1_SIZE
   PointSize Curve2 = CARDANO_BLST_P2_SIZE
 
+-- | Type family to get the size of a compressed point on a given curve
+type family CompressedPointSize curve where
+  CompressedPointSize Curve1 = 48
+  CompressedPointSize Curve2 = 96
```
lehins: I think it would be a good idea to avoid duplication of these hard coded values and extract `compressedSizePoint_` function out of the `BLS` type class and implement it in terms of this type family.

cardano-crypto-class/testlib/Test/Crypto/Util.hs
```diff
@@ -213,6 +215,34 @@ instance Arbitrary Message where
...
 
+--------------------------------------------------------------------------------
+-- BLS12381 signature context
+--------------------------------------------------------------------------------
+
+data BLS12381SignContext = BLS12381SignContext
+  { blsSignContextDst :: Maybe ByteString
+  , blsSignContextDAug :: Maybe ByteString
+  }
+  deriving (Eq, Show)
```
lehins: We should make this the actual context, instead of a tuple of maybes

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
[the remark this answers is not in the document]
ch1bo: Sure. As this is a "nitpick", not changing would also be fine by me.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
ch1bo: Good point. Maybe that warrants a `intToCSize :: Int -> CSize` with `intToCSize = fromIntegral` somewhere in this module (or package)?

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
```diff
@@ -455,3 +469,138 @@ decodeSignKeyDSIGNM = do
...
+-- Examples of aggregatable signatures schemes are the BLS signature scheme and
+-- the Pixel scheme.
+class
+  ( DSIGNAlgorithm v
+  , Show (ProofOfPossessionDSIGN v)
+  , Eq (ProofOfPossessionDSIGN v)
+  , NoThunks (ProofOfPossessionDSIGN v)
+  , KnownNat (SizeProofOfPossessionDSIGN v)
+  ) =>
+  DSIGNAggregatable v
```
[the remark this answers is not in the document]
perturbing: Fixed by 2ab0c54

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
```diff
@@ -455,3 +469,136 @@ decodeSignKeyDSIGNM = do
...
+  --  This function is unsafe and should only be used when the caller can
+  --  guarantee that all verification keys are valid (i.e., their PoPs have been
+  --  verified through other means).
+  aggregateVerKeysDSIGNWithoutPoPs ::
+    HasCallStack =>
+    [VerKeyDSIGN v] ->
+    Either String (VerKeyDSIGN v)
+
+  -- | Aggregate multiple signatures into a single signature
+  aggregateSigDSIGN ::
```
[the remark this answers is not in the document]
perturbing: Fixed by 2ab0c54

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
```diff
@@ -680,3 +738,114 @@ prop_dsignm_verify_neg_msg lock _ a a' =
...
+    prop "aggregate verify positive" $
+      forAllShow (genAggregateCase genContext genMsg) ppShow $
+          \(ctx, msg, vksPops, sigs) ->
+            case aggregateSigDSIGN @v sigs of
+              Left _  -> counterexample "aggregateSigDSIGN failed" False
+              Right s -> verifyAggregateDSIGN @v ctx vksPops msg s === Right ()
+    prop "aggregate verify negative (wrong message)" $
+      forAllShow (genAggregateCase2Msgs genContext genMsg) ppShow $
+        \(ctx, msg, msg', vksPops, sigs) ->
+          msg /= msg' ==> case aggregateSigDSIGN @v sigs of
```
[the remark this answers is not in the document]
perturbing: Fixed in f10130b

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+  type SeedSizeDSIGN (BLS12381DSIGN curve) = CARDANO_BLST_SCALAR_SIZE
+  type SizeSignKeyDSIGN (BLS12381DSIGN curve) = CARDANO_BLST_SCALAR_SIZE
+
+  -- \| These *Sizes* are used in the serialization/deserialization
+  -- so these use the compressed sizes of the BLS12-381 `Point curve`
+  type SizeVerKeyDSIGN (BLS12381DSIGN curve) = CompressedPointSize curve
+  type SizeSigDSIGN (BLS12381DSIGN curve) = CompressedPointSize (Dual curve)
+  type Signable (BLS12381DSIGN curve) = SignableRepresentation
+
+  -- \| Context can hold domain seperation tag and/or augmentation data for signatures
```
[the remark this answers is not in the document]
perturbing: Fixed by 4dac9169

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+    deriving newtype (NFData)
+    deriving stock (Show, Generic)
+    deriving anyclass (NoThunks)
+
+  newtype SignKeyDSIGN (BLS12381DSIGN curve)
+    = SignKeyBLS12381 (PinnedSizedBytes (SizeSignKeyDSIGN (BLS12381DSIGN curve)))
+    -- Us Eq from PinnedSizedBytes is needed here, as we need constant time
+    -- comparison for signing keys
+    deriving newtype (Eq, NFData)
+    deriving stock (Show, Generic)
```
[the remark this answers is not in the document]
perturbing: We require it here

https://github.com/IntersectMBO/cardano-base/blob/6974b7053907594becec12ba49b2352c9b73b42f/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs#L90

Shall we do the above show instance for all implementations of this class? If so, I'll happily open an issue for it and do it in another PR

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,507 @@
...
+  type SizeSigDSIGN (BLS12381DSIGN curve) = CompressedPointSize (Dual curve)
+  type Signable (BLS12381DSIGN curve) = SignableRepresentation
+
+  -- \| Context can hold domain seperation tag and/or augmentation data for signatures
+  type ContextDSIGN (BLS12381DSIGN curve) = (Maybe ByteString, Maybe ByteString)
+  type KeyGenContextDSIGN (BLS12381DSIGN curve) = Maybe ByteString
+
+  newtype VerKeyDSIGN (BLS12381DSIGN curve)
+    = -- Note that the internal representation is the uncompressed point size
+      VerKeyBLS12381 (PinnedSizedBytes (PointSize curve))
```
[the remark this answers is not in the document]
perturbing: Fixed by 4dac91690

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
perturbing: I made an issue for this. Thanks for the suggestion :) 

https://github.com/IntersectMBO/cardano-base/issues/604

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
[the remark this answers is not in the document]
lehins: Yes please. It is totally ok to require a Show instance, it is just not ok to actually show any secrets. I didn't even realize there was a derived Show instance for `SignKeyDSIGN Ed25519DSIGN`. For cardano-node it is not really an issue, since it never deals with `SignKey`s and memory locked version for kes-agent actually does the correct thing of not showing the contents (despite the wrong haddock).

Conceptually even `VerKeyDSIGN Ed25519DSIGN` should not even have a Show instance, since there are real world scenarios when verification key is also kept secret, we just don't use it that way in Cardano.

We could potentially show a hash of VerKey and SignKey, instead of the actual key. But I have no preference.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,489 @@
...
+                  _ -> return $ Left "SigDSIGN BLS12381DSIGN failed to verify"
+
+  {-# NOINLINE genKeyDSIGN #-}
+  genKeyDSIGN = genKeyDSIGNWithKeyInfo Nothing
+  {-# NOINLINE genKeyDSIGNWithKeyInfo #-}
+  -- \| Generate a signing key from a seed and optional key info
+  -- as per the IETF bls signature draft 05
+  genKeyDSIGNWithKeyInfo keyInfo seed =
+    SignKeyBLS12381 $
+      let (bs, _) = getBytesFromSeedT (seedSizeDSIGN (Proxy @(BLS12381DSIGN curve))) seed
```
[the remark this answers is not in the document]
lehins: Yes, I 100% agree, I never liked this ugliness! Please, a ticket is very much welcomed.
DSIGN has statically known size, so all we need to do is add a type wrapper around `Seed` that reflects its size at a type level as well, which would allow us to make this interface total.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
perturbing: Rereading this conversation, I do think that the name of the function is too opinionated towards BLS signatures. 

In this PR, the extra key generation context is often called extra key info. But in the context of Pixel, the context is some public parameters.

So I'm suggesting renaming it to
```haskell
genKeyDSIGNWithContext :: KeyGenContextDSIGN v -> Seed -> SignKeyDSIGN v
genKeyDSIGNWithContext _ = genKeyDSIGN
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
```diff
@@ -162,6 +174,9 @@ class
   -- is at least 'seedSizeDSIGN' bytes long.
   genKeyDSIGN :: Seed -> SignKeyDSIGN v
 
+  genKeyDSIGNWithContext :: KeyGenContextDSIGN v -> Seed -> SignKeyDSIGN v
```
perturbing: This function is not tested. I will add it

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -0,0 +1,482 @@
...
+--
+-- > Make sure that the either you switch off let-floating
+-- > (-fno-full-laziness), or that the call to unsafePerformIO cannot float
+-- > outside a lambda.
+--
+-- If we do not switch off let-floating, our calls to unsafeDupablePerformIO for
+-- FFI functions become nondeterministic in their behaviour when run with
+-- parallelism enabled (such as -with-rtsopts=-N), possibly yielding wrong
+-- answers on a range of tasks, including serialization.
+{-# OPTIONS_GHC -fno-full-laziness #-}
```
lehins: I think we forgot to remove this
```suggestion
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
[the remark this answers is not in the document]
lehins: Sorry @ch1bo  I was sure I responded to your questions, but maybe I forgot to submit my comment or it is in some other conversation in this PR. In any case:

> Should we just drop all NOINLINE pragmas in the DSIGN API (and let the compiler decide whether to inline or not)? I understand that the INLINE pragmas are useful for the plutus use case though.

We should add `INLINE` pragmas for all of the crypto functions, since they are usually very small, because the heavy lifting is done over FFI and inlining in Haskell can have a huge impact on performance.

Also we need to get rid of this flag together with the incorrect comment:
```suggestion
```

cardano-crypto-class/bench/Bench/Crypto/DSIGN.hs
```diff
@@ -79,10 +86,115 @@ class ExampleSignable v a | v -> a where
...
+    , NFData (SigDSIGN v)
+    , NFData (PossessionProofDSIGN v)
+    )
+  => Proxy v
+  -> String
+  -> Benchmark
+benchAggDSIGN _ lbl =
+  let msg = exampleSignable (Proxy @v)
+      ctx = exampleContext (Proxy @v)
+      ns = 1 : [100, 200 .. 1000]
```
lehins: That's a little too many benchmarks, IMHO. This will be a little saner
```suggestion
      ns = 1 : [250, 500 .. 1000]
```

lehins (approved):
This is a serious piece of work and I'd like to thank everyone that was involved!

PR #597, opened 2025-12-23
move header file from c includes to install-includes

### LINKED ISSUE #596: Prefix all `blst_` C definitions with `cardano_blst_`
We shouldn't be adding C definitions that could potentially conflict with definitions in the original library or some other C code in a Haskell package that depends on `cardano-crypto-class`

`cbits/blst_utils.*` are the offenders here.

Same comment applies to all of the `#define`s, I believe.

perturbing:
# Description

closes #596 

# Checklist

- [x] Self-reviewed the diff


lehins (approved):
@perturbing Thank you for adding those prefixes.

PR #598, opened 2025-12-23
Move `blst_util.h` from `install-includes` to `extra-source-files`

neilmayhew:
# Description

Move `blst_util.h` from `install-includes` to `extra-source-files`
    
This avoids using the deprecated `includes` field and is more consistent with what we do in other packages. The only reason to put it in `install-includes` is for client code to be able to use it, but we want client code to use the Haskell bindings and not the C bindings.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -21,6 +21,9 @@ extra-doc-files:
   CHANGELOG.md
   README.md
 
+extra-source-files:
+  cbits/blst_util.h
+
 data-files:
   bls12-381-test-vectors/test_vectors/bls_sig_aug_test_vectors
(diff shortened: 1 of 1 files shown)


lehins (commented):
@perturbing As it turns out:
* in order for `.h` files to be used in a Haskell project they need to be included in the `extra-source-files`
* in order for `.h` files to be usable outside of the project by some dependency then instead it needs to be included in `install-includes`
* `includes` is a long deprecated cabal field that was only need for ghc < 6.12

So, this PR fixes all of those issues.
We'll wait for your review before, merging this PR, but @neilmayhew  and I felt like there is no need to make those custom size definitions from the header file available to the outside world, since there are more usable Haskell alternatives that are provided by BLS interface at the type level.

PR #599, opened 2025-12-25
A couple of memory optimizations

lehins:
# Description

This PR switches `OutputVRF` to use `ByteArray` in order to not contribute to pinned memory fragmentation.
Also it switches decoding of `PackedBytes` to use `ByteString` for temporary buffer to avoid unnecessary copy.

Couple of useful `FromCBOR`/`ToCBOR` instances were added to accommodate above changes

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.3.0.0
 
+* Remove `Serialise` instance for `PackedBytes` as unused
 * Switch `OutputVRF` to use `ByteArray` instead of `ByteString`. Change field accessor name to `getOutputVRFByteArray`
 * Add `byteArrayToNatural`, `naturalToByteArray` and `byteArrayToInteger`.
 * Expose `bytesToInteger`
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -130,7 +130,6 @@ library
     mtl,
     nothunks,
     primitive >=0.8,
-    serialise,
     template-haskell,
     text,
     th-compat,
--- a/cardano-crypto-class/src/Cardano/Crypto/PackedBytes.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/PackedBytes.hs
@@ -28,7 +28,6 @@ module Cardano.Crypto.PackedBytes
 
 import Codec.CBOR.Decoding as D (decodeBytes)
 import Cardano.Binary (FromCBOR (..), ToCBOR (..), Size)
-import Codec.Serialise (Serialise(..))
 import Control.DeepSeq (NFData(..))
 import Control.Monad (unless, when)
 import Control.Monad.Primitive (primitive_)
@@ -140,10 +139,6 @@ instance KnownNat n => MemPack (PackedBytes n) where
       -- related to the `IO` we are escaping.
       (\addr# -> accursedUnutterablePerformIO $ packPinnedPtr (Ptr (addr# `plusAddr#` curPos#)))
   {-# INLINE unpackM #-}
-
-instance KnownNat n => Serialise (PackedBytes n) where
-  encode = toCBOR
-  decode = fromCBOR
 
 instance KnownNat n => ToCBOR (PackedBytes n) where
   toCBOR = toCBOR . unpackBytes


cardano-crypto-class/src/Cardano/Crypto/Util.hs
```diff
@@ -137,6 +155,12 @@ bytesToInteger (BS.PS fp (GHC.I# off#) (GHC.I# len#)) =
     importIntegerFromAddr :: Addr# -> Word# -> Int# -> IO Integer
     importIntegerFromAddr addr sz = integerFromAddr sz addr
 
+byteArrayToInteger :: ByteArray -> Integer
```
neilmayhew: Maybe this should have a Haddock comment similar to the one for `bytesToInteger`

cardano-crypto-praos/src/Cardano/Crypto/VRF/Praos.hs
```diff
@@ -59,23 +59,26 @@ module Cardano.Crypto.VRF.Praos (
...
 import Control.DeepSeq (NFData (..))
-import Control.Monad (void)
+import Control.Monad (void, (<$!>))
 import Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Unsafe as BS
 import Data.Coerce (coerce)
-import Data.Maybe (fromMaybe)
+import Data.Primitive.ByteArray (
+  ByteArray,
```
neilmayhew: Although this is just a re-export of the one from `base`, it might be confusing to readers to see it coming from a different place in this module. Although it would require an extra import to get it from `base`, the greater consistency might be helpful.

cardano-crypto-praos/testlib/Test/Crypto/VRF.hs
```diff
@@ -490,5 +492,6 @@ instance
 
 instance VRFAlgorithm v => Arbitrary (OutputVRF v) where
   arbitrary = do
-    bytes <- BS.pack <$> vectorOf (fromIntegral (sizeOutputVRF (Proxy :: Proxy v))) arbitrary
-    return $ OutputVRF bytes
+    sbs <- SBS.pack <$> vectorOf (fromIntegral (sizeOutputVRF (Proxy :: Proxy v))) arbitrary
+    case sbs of
+      SBS.SBS ba -> return $ OutputVRF $ ByteArray ba
```
neilmayhew: Use `SBS.SBS ba <- ...`?

cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
```diff
@@ -224,10 +225,16 @@ hashFromOffsetBytesShort ::
   Maybe (Hash h a)
 hashFromOffsetBytesShort bytes offset = UnsafeHashRep <$> packShortByteStringWithOffset bytes offset
 
--- | The representation of the hash as bytes, as a 'ShortByteString'.
+-- | /O(n)/ - The representation of the hash a 'ShortByteString'.
```
neilmayhew: Grammar

neilmayhew:
```suggestion
-- | /O(n)/ - The representation of the hash as a 'ShortByteString'.
```

cardano-crypto-praos/testlib/Test/Crypto/VRF.hs
neilmayhew:
```suggestion
  SBS.SBS ba <- SBS.pack <$> vectorOf (fromIntegral (sizeOutputVRF (Proxy :: Proxy v))) arbitrary
  return $ OutputVRF $ ByteArray ba
```

lehins: :smile: 
That is how I originally implemented it. However, `bytestring` maintainers forgot to add `COMPLETE` pragma to the `SBS` pattern synonym in the newer `bytestring` version, which wrongfully tells ghc that this pattern match can fail, therefore it asks for `MonadFail` instance, which `Gen` monad naturally lacks.
Would you be willing to submit a PR into `bytestring` to add a `COMPLETE` pragma for that pattern synonym? :grin:

cardano-crypto-class/src/Cardano/Crypto/VRF/Class.hs
```diff
@@ -210,22 +212,31 @@ instance
...
   deriving newtype (NFData)
 
+-- ByteArray is already in NF
+deriving via
+  OnlyCheckWhnfNamed "OutputVRF" (OutputVRF v)
+  instance
+    NoThunks (OutputVRF v)
+
+getOutputVRFBytes :: OutputVRF v -> ByteString
+getOutputVRFBytes = SBS.fromShort . byteArrayToShortByteString . getOutputVRFByteArray
```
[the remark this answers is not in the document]
lehins: There is no reason to deprecate helper functions like this, since it will only introduce breakage for no reason. It is the underlying representation that matters most, intermediate not so much.
In fact, in production we only ever convert `OutputVRF` to `Natural`.
Conversion to `ByteString` only happens in tests and JSON instances, so nothing terribly important that has to be as efficient as possible

PR #608, opened 2026-01-15
Add newtype wrappers around IPv4 and IPv6

### LINKED ISSUE #607: Move `IPv4` and `IPv6` from ledger
https://github.com/IntersectMBO/cardano-ledger/pull/5520 introduced newtype wrappers around `iproute` types in order to avoid orphan instances and fix problems with original implementation.
Implementation of those types need to be moved into `cardano-base`, instead of being defined in `cardano-ledegr-binary`

f-f:
# Description

Fix #607

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-base/src/Cardano/Base/IP.hs
+++ b/cardano-base/src/Cardano/Base/IP.hs
@@ -1,0 +1,143 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+-- | This module provides newtype wrappers for 'IPv4' and 'IPv6' addresses
+-- from the @iproute@ package. These wrappers exist to have a correct 'Show'
+-- instance and eliminate the need for orphan instances.
+module Cardano.Base.IP (
+  IPv4,
+  IPv6,
+  mkIPv4,
+  unIPv4,
+  mkIPv6,
+  unIPv6,
+  toIPv4,
+  toIPv4w,
+  fromIPv4,
+  fromIPv4w,
+  toIPv6,
+  toIPv6w,
+  fromIPv6,
+  fromIPv6w,
+  fromHostAddress6,
+)
+where
+
+import Control.DeepSeq (NFData (..), rwhnf)
+import Data.Aeson (FromJSON (..), ToJSON (..))
+import qualified Data.Aeson as Aeson
+import qualified Data.IP as IP
+import qualified Data.Text as Text
+import Data.Word (Word32)
+import GHC.Generics (Generic)
+import NoThunks.Class (NoThunks (..))
+import Text.Read (readMaybe, readPrec)
+
+newtype IPv4 = IPv4 {unIPv4 :: IP.IPv4}
+  deriving newtype (Eq, Ord)
+  deriving stock (Generic)
+
+instance NFData IPv4 where
+  rnf = rwhnf
+
+instance NoThunks IPv4 where
+  wNoThunks _ (IPv4 _) = return Nothing
+  showTypeOf _ = "IPv4"
+
+-- >>> show (toIPv4 [192, 168, 1, 1])
+-- "\"192.168.1.1\""
+instance Show IPv4 where
+  show (IPv4 ip) = show (show ip)
+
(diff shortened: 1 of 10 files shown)


cardano-base/src/Cardano/Base/IP.hs
```diff
@@ -0,0 +1,127 @@
...
+toIPv6 = toIPv6w . IP.toHostAddress6 . IP.toIPv6
+
+toIPv6w :: (Word32, Word32, Word32, Word32) -> IPv6
+toIPv6w (!w1, !w2, !w3, !w4) = IPv6 (IP.toIPv6w (w1, w2, w3, w4))
+
+fromIPv6 :: IPv6 -> [Int]
+fromIPv6 (IPv6 ip) = force (IP.fromIPv6 ip)
+
+fromIPv6w :: IPv6 -> (Word32, Word32, Word32, Word32)
+fromIPv6w (IPv6 ip) = case IP.toHostAddress6 ip of
```
lehins: I think it would make sense to actually add `fromHostAddress6` that forces the contents, instead of forcing it here.
If we can guarantee the invariant that `IPv6` newtype is always strict by construction then we never need to force it when we unwrap it.

lehins (changes requested):
It would be useful to export conversion functions that allows to wrap with forcing and unwrap the underlying types:
```haskell
wrapIPv4 :: IP.IPv4 -> IPv4
wrapIPv4 = IPv4

unwrapIPv4 :: IPv4 -> IP.IPv4
unwrapIPv4 = unIPv4

wrapIPv6 :: IP.IPv6 -> IPv6
wrapIPv6 ipv6 = 
  case IP.toIPv6w ipv6 of
    (!w1, !w2, !w3, !w4) -> IPv6 ipv6 

unwrapIPv6 :: IPv6 -> IP.IPv6
unwrapIPv6 = unIPv6
```

If we did that, then this would be the new implementation of `toIPv6w`:
```haskell
toIPv6w :: (Word32, Word32, Word32, Word32) -> IPv6
toIPv6w = wrapIPv6 . IP.toIPv6w
```

Instead of `wrap` prefix we could use `mk` or something else entirely, I don't have any strong preference about it.

We also need to add `FromCBOR` and `ToCBOR` instances for those two new types to the `cardano-binary`.

cardano-base/src/Cardano/Base/IP.hs
```diff
@@ -0,0 +1,143 @@
...
+  readPrec = do
+    s <- readPrec
+    case readMaybe s of
+      Just ip -> pure (IPv4 ip)
+      Nothing -> fail "invalid IPv4"
+
+instance FromJSON IPv4 where
+  parseJSON =
+    Aeson.withText "IPv4" $ \txt -> case readMaybe (Text.unpack txt) of
+      Just ip -> pure (IPv4 ip)
```
neilmayhew: Although it's a noop, it would be nice to see `mkIPv4` used here (and in the `Read` instance). That would be consistent with the v6 implementation, and also allow for changes to the `IPv4` type in the future (even though that's unlikely).

cardano-base/src/Cardano/Base/IP.hs
```diff
@@ -0,0 +1,143 @@
...
+      Nothing -> fail "invalid IPv6"
+
+instance FromJSON IPv6 where
+  parseJSON =
+    Aeson.withText "IPv6" $ \txt -> case readMaybe (Text.unpack txt) of
+      Just ip -> pure (mkIPv6 ip)
+      Nothing -> fail $ "failed to read as IPv6 " ++ show txt
+
+instance ToJSON IPv6 where
+  toJSON (IPv6 ip) = Aeson.toJSON (show ip)
```
neilmayhew: It would be nice to see some tests for these instances. It would be a bit cumbersome to have these as doctests, so perhaps you could consider adding a small test suite. It wouldn't need to have property tests (since those would end up being a bit tautological), just a few examples.

If you do end up doing that, adding the doctest examples for `Show` and `Read` might be a good idea, since we may not always be running doctests in CI. We had them disabled for a long time because things broke awkwardly and we didn't have time to fix them.

I've found the `aeson-qq` library to be very helpful when writing example tests involving JSON.

neilmayhew (commented):
Looks great!

Just a few minor suggestions.

neilmayhew (approved):
Looking good!

PR #612, opened 2026-01-24
Move `Arbitrary` instances from `cardano-ledger`

neilmayhew:
# Description

Move these instances closer to their types. This enables us to write better tests for `cardano-base` libraries.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.3.0.0
 
+* Add `Arbitrary` instances for `SigDSIGN`, `SignKeyDSIGN`, `SignedDSIGN`, `VerKeyDSIGN`
 * Add `Arbitrary` instances for `CertVRF`, `CertifiedVRF`, `OutputVRF`, `SignKeyVRF`, `VerKeyVRF`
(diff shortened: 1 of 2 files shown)


cardano-base/testlib/Test/Cardano/Base/Bytes.hs
```diff
@@ -0,0 +1,49 @@
...
+  uniformByteStringM,
+  uniformShortByteStringM,
+ )
+#else
+import System.Random.Stateful (
+  StatefulGen (..),
+  runStateGen_,
+  uniformByteStringM,
+ )
+#endif
```
neilmayhew: I'm doing it this clumsy way rather than the original way (in `cardano-ledger-binary`),

```haskell
import System.Random.Stateful hiding (genByteString, genShortByteString)
```

because `genByteString`, `genShortByteString` will eventually disappear and there will be a compilation error that we have to come back and fix.

lehins: invalid `hiding` does not result in compilation error, it is a warning, that is why the way that it is implemented in `cardano-ledger-binary` is fine.
That being said, using explicit import the way you did it here is also fine.

PR #629, opened 2026-02-13
remove `mu2` from PoP, following IETF draft

### LINKED ISSUE #625: optimize BLS PoP to follow IETF draft
In #587 we added BLS signatures to `cardano-base`, this because Leios will need them. This issue suggest dropping $` \mu_2 `$ from the proof of possession.

In the work that has been done, the proof of possession protocol followed the Leios paper. Where we roughly define (see the code [here](https://github.com/IntersectMBO/cardano-base/blob/master/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs#L424:L427))

- $` \mu_1 = sign("PoP" | vk\_bytes, sk) \in G_{1/2}`$
- $` \mu_2 = g_{\{1/2\}}^{sk} \in G_{\{1/2\}} `$ (note that this is just the public key in the other group)

For a $` vk \in G_{\{2/1\}} `$. The latter, $` \mu_2 `$ has been added to follow the paper, breaking what the IETF does (which just returns $` \mu_1 `$ as the proof). In normal BLS signatures, what the IETF standardizes, the latter has no additional use. In academics, this is often added to the model of BLS signatures to aid the security proofs. This is since there is an isomorphism between $` G_1 `$ and $` G_2 `$, and the math is easier if you have the representation of the public key without this mapping, by just checking that $` e(pk,g_1) = e(g_2, \mu_2) = e(g_2, pk') `$ (for $` pk \in G_1 `$ and $` pk' \in G_2 `$).

Now note that in Leios we also use the BLS key as a VRF, which does not warrant the use of a proof of possession, but we should still be careful with changes that do not follow the paper. 

If we do make this change, this will save bytes in the PoP, but also the verification time of such a PoP will become equivalent to just a signature check (we currently do roughly the work of two such checks). Additionally, this will make our implementation IETF compliant, making it easier for other languages to grab off-the-shelf BLS signature libs instead of having to implement their own crypto.

- [x] Check with research that dropping $` \mu_2 `$ poses no security risk (it is only of utility in the security proofs)
- [x] Implement the changes

The below image shows the adjusted BLS protocol for Leios (BLS signatures + VRF over the same keys)

Image

perturbing:
# Description

Closes #625

Just for safety, this PR is split into three commits to ensure that in these intermediate steps, all test still pass. As an overview, in commit:

1. We change the type of the PoP
2. We simplify how the PoP is created by switching to using `signDIGN`
3. We simplify the verification of the PoP by using `verifyDSIGN`

Commit 2 and 3 reduce the FFI surface we use. Also note that throughout the whole test, we have a full round of PoP creation and verification under the hood [here](https://github.com/IntersectMBO/cardano-base/blob/ef58f660a20823e6cd4a6d27623bf9e3eef46139/cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs#L788:L790). The passing of each committee implies that functionally we did not change anything.

### DIFF
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.5.0.0
 
+* Drop mu2 from `PossessionProofDSIGN`
 * Add `Ord` superclass constraint for the `CertVRF` associated type in `VRFAlgorithm`
 * Add `Ord` instances for `CertVRF {SimpleVRF,PraosVRF,PraosBatchCompatVRF}`
 * Add `Ord` instances for `Cardano.Crypto.VRF.Praos{,BatchCompat}.Proof`
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
@@ -88,14 +88,12 @@ import Cardano.Crypto.EllipticCurve.BLS12_381.Internal (
   blsGenerator,
   blsHash,
   blsIsInf,
-  blsMult,
   blsUncompress,
   c_blst_keygen,
   finalVerifyPairs,
   mkBLSTError,
   scalarFromBS,
   scalarToBS,
-  scalarToInteger,
   toAffine,
   withAffine,
   withMaybeCStringLen,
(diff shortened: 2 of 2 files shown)


lehins (approved):
Sorry for taking so long to getting to review this PR.
I don't know why mu2 was used to begin with. So, ignoring my understanding of cryptographic reasoning for this change, these PR look good.

perturbing:
Rebased :+1:

PR #630, opened 2026-02-16
add test vectors of IETF draft bls signature scheme

### LINKED ISSUE #626: add external bls ietf draft 6 test-vectors to tests
Given this issue https://github.com/IntersectMBO/cardano-base/issues/625, it would be good to check that the implementation is compliant with the IETF draft [here](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-06.html).

Being compliant is important, as it prevents other implementations of the node or tools from having to run their own crypto.

Note that the audit (see this issue) has the scope of whether we are compliant.

Since the IETF is still in draft, the spec there has no test vectors. That said, previous work from @iquerejeta [here](https://github.com/input-output-hk/bls-e2e-testvectors/tree/main/src) does provide these. 

It does not provide test vectors for the PoP or key gen (this is compliant by construction of blast; see [here](https://github.com/supranational/blst/blob/806515251de5cfb9761ad12eb2f2f151b1ad0277/src/keygen.c#L145) against the draft-6 required [string](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-06.html#name-choosing-a-salt-value-for-k))

- [ ] Create/find test vectors for PoP
- [ ] Add test vectors for signatures from above link

perturbing:
Closes #626

This PR adds test vectors for our BLS signature implementation.

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -181,7 +181,9 @@ library testlib
     Test.Crypto.Regressions
     Test.Crypto.RunIO
     Test.Crypto.Util
+    Test.Crypto.Vector.BLS12381DSIGN
     Test.Crypto.Vector.SerializationUtils
+    Test.Crypto.Vector.Vectors
 
   other-modules: Paths_cardano_crypto_class
   autogen-modules: Paths_cardano_crypto_class
@@ -215,7 +217,6 @@ library testlib
     exposed-modules:
       Test.Crypto.Vector.Secp256k1DSIGN
       Test.Crypto.Vector.StringConstants
-      Test.Crypto.Vector.Vectors
 
 library benchlib
   import: base, project-config
--- a/cardano-crypto-class/testlib/Test/Crypto/Vector/BLS12381DSIGN.hs
+++ b/cardano-crypto-class/testlib/Test/Crypto/Vector/BLS12381DSIGN.hs
@@ -1,0 +1,114 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+module Test.Crypto.Vector.BLS12381DSIGN (
+  tests,
+) where
+
+import Cardano.Crypto.DSIGN (
+  DSIGNAggregatable (..),
+  DSIGNAlgorithm (ContextDSIGN, SigDSIGN, Signable, VerKeyDSIGN),
+  algorithmNameDSIGN,
+  verifyDSIGN,
+ )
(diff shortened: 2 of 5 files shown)


perturbing:
This can be merged after https://github.com/IntersectMBO/cardano-base/pull/629

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -215,7 +217,6 @@ library testlib
     exposed-modules:
       Test.Crypto.Vector.Secp256k1DSIGN
       Test.Crypto.Vector.StringConstants
-      Test.Crypto.Vector.Vectors
```
lehins, line 218: Moving this module out of this cabal flag will cause it to fail compilation if `DSECP256K1_ENABLED` is turned off.

I am not quite sure why it was developed behind a cabal flag, but we don't want to break it all of a sudden. I suggest we investigate whether  we can remove this flag and make secp256k a permanent feature without ability to opt-out.

Although, that will have to be done in a separate PR.

cardano-crypto-class/testlib/Test/Crypto/Vector/BLS12381DSIGN.hs
```diff
@@ -0,0 +1,114 @@
...
+  ) =>
+  ContextDSIGN v ->
+  ([VerKeyDSIGN v], a, SigDSIGN v) ->
+  Spec
+verifyAggregatedTest ctx (vKeys, msg, sig) =
+  it "Aggregated verification should be successful." $ do
+    let result =
+          case uncheckedAggregateVerKeysDSIGN vKeys of
+            Left err -> Left err
+            Right avk -> verifyDSIGN ctx avk msg sig
```
lehins, line 111:
```suggestion
    let result = do
          avk <- uncheckedAggregateVerKeysDSIGN vKeys
          verifyDSIGN ctx avk msg sig
```

cardano-crypto-class/testlib/Test/Crypto/Vector/Vectors.hs
```diff
@@ -18,20 +18,31 @@ module Test.Crypto.Vector.Vectors (
...
 )
 where
 
 import Cardano.Binary (FromCBOR)
 import Cardano.Crypto.DSIGN (
+  BLS12381MinSigDSIGN,
+  BLS12381MinVerKeyDSIGN,
+  DSIGNAggregatable (PossessionProofDSIGN),
   DSIGNAlgorithm (SigDSIGN, SignKeyDSIGN, VerKeyDSIGN),
   EcdsaSecp256k1DSIGN,
```
lehins, line 36: If we want this module to be always visible, then we need to hide all secp256k functionality behind `SECP256K1_ENABLED` CPP flag

If we are to preserve this flag we should really have a separate CI build with it disabled

lehins (changes requested):
Aside from issue with `SECP256K1_ENABLED` flag this PR looks good

PR #634, opened 2026-03-07
cardano-crypto-class: constraint crypton library

coot:
The newest version on Hackage breaks the build process.


lehins (changes requested):
We should avoid bounds that are too strict for no good reason

lehins (approved):
Thank you!

PR #635, opened 2026-03-10
implement audit findings

### LINKED ISSUE #710: Audit cryptography primitives
## Why

Changes to `cardano-base` are critical for the security of the whole `cardano-node`. It is common practice that code going into this repository is independently audited before used in production. The changes to `cardano-base` made in course of #547 are going to also be covered by the next audit.

## What

- Audit of recently added BLS key aggregation primitives to `cardano-crypto-class` (pending version `2.3.0.0`) undertaken by BCryptic
- Addition of `blsMSM` of [`Cardano.Crypto.EllipticCurve.BLS12_381.Internal`](https://github.com/IntersectMBO/cardano-base/blob/4b27937b34b957a1afb0014f9d6e9038163e0763/cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs#L1028) audited
- Addition of BLS12381.hs to the DSIGN interface audited

perturbing:
# Description

This PR fixes seven things,

1. It removes the unused termination null pointer from `PointArrayPtr` and in its helper function `withPointArray`. It also removes these null pointers for `AffineArrayPtr` and removes the previously used function `withAffineBlockArrayPtr` as in the usage is no longer needed. This is because `blst_{p1s,p2s}_to_affines` already returns a continuous block of points, and `blst` is flexible in what form it receives its data (the previous method of conversion was not needed). Lastly, this commit also removes the unused termination null pointer from `ScalarArrayPtr` and its helper function `withScalarArray`.

2. It fixed the `memcmp` FFI import, which was incorrectly using `CSize` (unsigned int). Note that the return value of `memcmp` is a signed integer [here](https://www.tutorialspoint.com/c_standard_library/c_function_memcmp.htm). Since we used this function only in
```haskell
eqAffinePtr :: forall curve. BLS curve => AffinePtr curve -> AffinePtr curve -> IO Bool
eqAffinePtr (AffinePtr a) (AffinePtr b) =
  (== 0) <$> c_memcmp (castPtr a) (castPtr b) (sizeAffine_ (Proxy @curve))
```
Where the sign did not matter, it did not cause problems. **(Report finding 3)**

3. Use `CBool` instead of `Bool` in the FFI calls **(Report finding 6)**

4. Serialization of PoP bytes should reject the zero point in the group (point at infinity)  **(Report finding 2)**

5. Add the PoP Cipher suite as per the IETF draft (plus a comment on usage). **(Report finding 4)**

6. Add an extra `blsIsInf` check on inputs that are aggregated for signatures and verification keys. Note that before this change we roughly had the timing
```bash
nix run github:IntersectMBO/cardano-base/c9a6c770c018ccc7195156e70efe6355b1bab76b#cardano-crypto-class:bench:bench -- --match pattern 'DSIGN/BLS12381MinVerKey/Aggregatable/n=300/aggregateVerKeys (no PoPs)'
benchmarking DSIGN/BLS12381MinVerKey/Aggregatable/n=300/aggregateVerKeys (no PoPs)
time                 187.8 μs   (186.8 μs .. 188.7 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 186.7 μs   (186.1 μs .. 187.4 μs)
std dev              2.169 μs   (1.758 μs .. 2.980 μs)
```
Now this becomes
```bash
nix run github:IntersectMBO/cardano-base/1cb297dcf7b75b3a44b5df847161cb7e9394e971#cardano-crypto-class:bench:bench -- --match pattern 'DSIGN/BLS12381MinVerKey/Aggregatable/n=300/aggregateVerKeys (no PoPs)'
benchmarking DSIGN/BLS12381MinVerKey/Aggregatable/n=300/aggregateVerKeys (no PoPs)
time                 197.6 μs   (196.3 μs .. 199.6 μs)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 207.3 μs   (204.6 μs .. 209.7 μs)
std dev              8.305 μs   (7.332 μs .. 9.105 μs)
```
which is not significant, given that we gain safety. **(Report finding 5)**

7. add non-identity check to `rawDeserialiseVerKeyDSIGN` and `rawDeserialiseSignKeyDSIGN`. **(Report finding 8)**

Note that the report has 8 findings, where the above fixes 7. This is because **(report finding 7)** was on [this](https://github.com/IntersectMBO/cardano-base/pull/538) issue, which is already resolved (the audit confirmed it was correctly addressed)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.4.0.1
 
-*
+* Remove constructors of `BLS12381SignContext` from export; use `minSigPoPDST` or `minVerKeyPoPDST` for the standard PoP ciphersuites.
 
 ## 2.4.0.0
 
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog for `cardano-crypto-class`
 
-## 2.4.0.1
+## 2.5.0.0
 
 * Remove constructors of `BLS12381SignContext` from export; use `minSigPoPDST` or `minVerKeyPoPDST` for the standard PoP ciphersuites.
 
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
@@ -90,7 +90,6 @@ import Cardano.Crypto.EllipticCurve.BLS12_381.Internal (
   blsIsInf,
   blsMult,
   blsUncompress,
-  blsZero,
   c_blst_keygen,
   compressedSizePoint,
   finalVerifyPairs,
@@ -493,27 +492,35 @@ instance
     deriving anyclass (NFData)
 
   {-# INLINE uncheckedAggregateVerKeysDSIGN #-}
-  uncheckedAggregateVerKeysDSIGN verKeys = do
-    -- Sum the verification keys as curve points
-    let aggrPoint =
-          F.foldl' blsAddOrDouble (blsZero @curve) (map verKeyToPoint verKeys)
-    -- Unlikely case, but best to reject infinity as an aggregate verification
-    -- key. This happens if, for every secret/verification key pair, the inverse
-    -- of each secret key (and thus also the verification key) is also present
-    -- in the list.
-    if blsIsInf @curve aggrPoint
-      then Left "aggregateVerKeysDSIGN: aggregated verification key is infinity"
-      else Right $ VerKeyBLS12381 aggrPoint
+  uncheckedAggregateVerKeysDSIGN verKeys =
+    -- Reject any input verification key that is the infinity point
+    if any (blsIsInf @curve . verKeyToPoint) verKeys
+      then Left "uncheckedAggregateVerKeysDSIGN: input verification key is infinity"
+      else case map verKeyToPoint verKeys of
+        [] -> Left "uncheckedAggregateVerKeysDSIGN: empty list of verification keys"
+        (p : ps) ->
+          let aggrPoint = F.foldl' blsAddOrDouble p ps
+           in -- Unlikely case, but best to reject infinity as an aggregate verification
+              -- key. This happens if, for every secret/verification key pair, the inverse
+              -- of each secret key (and thus also the verification key) is also present
+              -- in the list.
+              if blsIsInf @curve aggrPoint
+                then Left "uncheckedAggregateVerKeysDSIGN: aggregated verification key is infinity"
+                else Right $ VerKeyBLS12381 aggrPoint
 
   {-# INLINE aggregateSigsDSIGN #-}
-  aggregateSigsDSIGN sigs = do
-    -- Sum the signatures as curve points
-    let aggrPoint =
-          F.foldl' blsAddOrDouble (blsZero @(DualCurve curve)) (map sigToPoint sigs)
-    -- Unlikely case, but best to reject infinity as an aggregate signature
-    if blsIsInf @(DualCurve curve) aggrPoint
-      then Left "aggregateSigsDSIGN: aggregated signature is infinity"
-      else Right $ SigBLS12381 aggrPoint
+  aggregateSigsDSIGN sigs =
+    -- Reject any input signature that is the infinity point
+    if any (blsIsInf @(DualCurve curve) . sigToPoint) sigs
+      then Left "aggregateSigsDSIGN: input signature is infinity"
+      else case map sigToPoint sigs of
+        [] -> Left "aggregateSigsDSIGN: empty list of signatures"
+        (p : ps) ->
+          let aggrPoint = F.foldl' blsAddOrDouble p ps
+           in -- Unlikely case, but best to reject infinity as an aggregate signature
+              if blsIsInf @(DualCurve curve) aggrPoint
+                then Left "aggregateSigsDSIGN: aggregated signature is infinity"
+                else Right $ SigBLS12381 aggrPoint
 
   {-# INLINE createPossessionProofDSIGN #-}
   createPossessionProofDSIGN BLS12381SignContext {blsSignContextDst = dst, blsSignContextAug = aug} (SignKeyBLS12381 skScalar) =
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
@@ -1,0 +1,8 @@
+module Cardano.Crypto.DSIGN.BLS12381 (
+  module X,
+  BLS12381SignContext,
+)
+where
+
+import Cardano.Crypto.DSIGN.BLS12381.Internal (BLS12381SignContext)
+import Cardano.Crypto.DSIGN.BLS12381.Internal as X hiding (BLS12381SignContext)
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
@@ -382,7 +382,11 @@ instance
     -- That is, the deserialised point is in the subgroup of Curve1/Curve2.
     case blsUncompress @curve bs of
       Left _ -> Nothing
-      Right vkPsb -> Just (VerKeyBLS12381 vkPsb)
+      Right vkPsb ->
+        -- Reject the identity (point at infinity) as a verification key
+        if blsIsInf @curve vkPsb
+          then Nothing
+          else Just (VerKeyBLS12381 vkPsb)
 
   {-# INLINE rawDeserialiseSignKeyDSIGN #-}
   rawDeserialiseSignKeyDSIGN bs =
@@ -392,7 +396,11 @@ instance
     -- they are valid Scalars, i.e., less than the curve order (255 bits).
     case scalarFromBS bs of
       Left _ -> Nothing
-      Right skScalar -> Just (SignKeyBLS12381 skScalar)
+      Right skScalar ->
+        -- Reject the zero scalar as a signing key
+        if BS.all (== 0) (scalarToBS skScalar)
+          then Nothing
+          else Just (SignKeyBLS12381 skScalar)
 
   {-# INLINE rawDeserialiseSigDSIGN #-}
   rawDeserialiseSigDSIGN bs =
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
@@ -500,11 +500,12 @@ instance
     deriving anyclass (NFData)
 
   {-# INLINE uncheckedAggregateVerKeysDSIGN #-}
-  uncheckedAggregateVerKeysDSIGN verKeys =
+  uncheckedAggregateVerKeysDSIGN verKeys = do
+    let verKeyPoints = map verKeyToPoint verKeys
     -- Reject any input verification key that is the infinity point
-    if any (blsIsInf @curve . verKeyToPoint) verKeys
+    if any (blsIsInf @curve) verKeyPoints
       then Left "uncheckedAggregateVerKeysDSIGN: input verification key is infinity"
-      else case map verKeyToPoint verKeys of
+      else case verKeyPoints of
         [] -> Left "uncheckedAggregateVerKeysDSIGN: empty list of verification keys"
         (p : ps) ->
           let aggrPoint = F.foldl' blsAddOrDouble p ps
@@ -517,11 +518,12 @@ instance
                 else Right $ VerKeyBLS12381 aggrPoint
 
   {-# INLINE aggregateSigsDSIGN #-}
-  aggregateSigsDSIGN sigs =
+  aggregateSigsDSIGN sigs = do
+    let sigPoints = map sigToPoint sigs
     -- Reject any input signature that is the infinity point
-    if any (blsIsInf @(DualCurve curve) . sigToPoint) sigs
+    if any (blsIsInf @(DualCurve curve)) sigPoints
       then Left "aggregateSigsDSIGN: input signature is infinity"
-      else case map sigToPoint sigs of
+      else case sigPoints of
         [] -> Left "aggregateSigsDSIGN: empty list of signatures"
         (p : ps) ->
           let aggrPoint = F.foldl' blsAddOrDouble p ps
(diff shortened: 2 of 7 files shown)


cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -391,27 +511,35 @@ instance
...
-    -- of each secret key (and thus also the verification key) is also present
-    -- in the list.
-    if blsIsInf @curve aggrPoint
-      then Left "aggregateVerKeysDSIGN: aggregated verification key is infinity"
-      else Right $ VerKeyBLS12381 aggrPoint
+  uncheckedAggregateVerKeysDSIGN verKeys =
+    -- Reject any input verification key that is the infinity point
+    if any (blsIsInf @curve . verKeyToPoint) verKeys
+      then Left "uncheckedAggregateVerKeysDSIGN: input verification key is infinity"
+      else case map verKeyToPoint verKeys of
```
lehins: Let's avoid duplicate work
```suggestion
  uncheckedAggregateVerKeysDSIGN verKeys = do
    let verKeyPoints = map verKeyToPoint verKeys
    -- Reject any input verification key that is the infinity point
    if any (blsIsInf @curve) verKeyToPoint
      then Left "uncheckedAggregateVerKeysDSIGN: input verification key is infinity"
      else case verKeyPoints of
```

lehins (commented):
Looks good. Just some comments with minor improvements.

I'll take one final look after comments are addressed.

cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs
```diff
@@ -409,11 +409,14 @@ withAffineBlockArrayPtr ::
   (AffineArrayPtr curve -> IO a) ->
   IO a
 withAffineBlockArrayPtr affinesBlockPtr numPoints go = do
-  allocaBytes (numPoints * sizeOf (nullPtr :: Ptr ())) $ \affineVectorPtr -> do
+  -- Allocate space for the points and a null terminator
```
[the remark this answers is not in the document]
perturbing: Note that the C functions `blst_p1s_to_affine` and `blst_p2s_to_affine` also do not require null-termination as per [this](https://github.com/supranational/blst/blob/f62244ef50ad1a603decdb8f215e982d2a467bb6/bindings/blst.h#L227) and [this](https://github.com/supranational/blst/blob/f62244ef50ad1a603decdb8f215e982d2a467bb6/src/multi_scalar.c#L19).

I think this got added because the [go](https://github.com/supranational/blst/blob/f62244ef50ad1a603decdb8f215e982d2a467bb6/bindings/go/blst.tgo#L48) bindings and [rust](https://github.com/supranational/blst/blob/f62244ef50ad1a603decdb8f215e982d2a467bb6/bindings/rust/src/pippenger.rs#L85) do add it.

So then the conclusion is, we should remove the null terminators everywhere.

lehins (changes requested):
I am sorry, but I will not allow unjustified null pointer termination in this repo. This is just bad practice all around! 🤮

perturbing:
I agree with your sentiment. I think this difference perhaps means that this is wrong
> that we are using the libblst API as it was intended.

The addition of the null pointer in the Go and Rust bindings is not to indicate null termination; it's to switch between continuous vs. scattered memory layout mode. Note that in libblst we have [here](https://github.com/supranational/blst/blob/f62244ef50ad1a603decdb8f215e982d2a467bb6/src/multi_scalar.c#L23) the logic
```C
point = *points ? *points++ : point+1;
```
Which is basically a check to test if the block needs to be used in a continuous layout or as a scattered one. That is, given the Rust/Go case `[&block_start, null]`
```C
const ptype *point = NULL;
...

// Iteration 1: *points = &block_start (non-null)
point = *points ? *points++ : point+1;
// → uses &block_start, advances `points` to index 1 (the null)

// Iteration 2: *points = null
point = *points ? *points++ : point+1;
// → uses point+1 (= block_start+1), does NOT advance `points` (stays at null)

// Iteration 3: *points = null
point = *points ? *points++ : point+1;
// → uses point+1 (= block_start+2), stays at null

// ... and so on for all npoints
```
So, the null at index 1 is read every subsequent iteration and each time causes point+1 (contiguous walk). Now in the scattered case, we then pass N individual pointers:
```C
points = [ ptr→Point_0 | ptr→Point_1 | ptr→Point_2 | ptr→Point_3 | ptr→Point_4 ]                                                                                                                                                          
                                                                                                                                                                                      
// Iteration 1: *points = ptr→Point_0 (non-null)                                                                                                                                    
point = *points ? *points++ : point+1;
// → uses ptr→Point_0, advances points to index 1
                                                                                                                                                                                      
// Iteration 2: *points = ptr→Point_1 (non-null)
point = *points ? *points++ : point+1;                                                                                                                                              
// → uses ptr→Point_1, advances points to index 2
                                                                                                                                                                                      
// Iteration 3: *points = ptr→Point_2 (non-null)
// → uses ptr→Point_2, advances points to index 3                                                                                                                                   
                 
// ... and so on
```
Now note that the C function `blst_{p1s,p2s}_to_affine` returns a continuous block of points! So I think that the current code does not need `withAffineBlockArrayPtr` in the `blsMSM` function (this saves `n-1 writes)` . That is,
```haskell
allocaBytes (numPoints * sizeAffine (Proxy @curve)) $ \affinesBlockPtr -> do
              c_blst_to_affines (AffineBlockPtr affinesBlockPtr) pointArrayPtr numPoints'
              -- Mode 2: contiguous block — [ptr_to_block, null] is sufficient
              allocaBytes (2 * sizeOf (nullPtr :: Ptr ())) $ \affineVectorPtr -> do
                let ptrArray = castPtr affineVectorPtr :: Ptr (Ptr ())
                pokeElemOff ptrArray 0 (castPtr affinesBlockPtr)
                pokeElemOff ptrArray 1 nullPtr
                let affineArrayPtr = AffineArrayPtr affineVectorPtr
                allocaBytes scratchSize $ \scratchPtr -> do
                  c_blst_mult_pippenger
                    resultPtr
                    affineArrayPtr
                    numPoints'
                    scalarArrayPtr
                    (255 :: CSize)
                    (ScratchPtr scratchPtr)
```
I'll push some changes and inform the auditor of this nuance! Similarly, `withPointArray` and `withScalarArray` allocated one extra pointer slot and write a null that is never read.

And thank you for the good review and finding the inconsistency!

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -26,6 +26,8 @@ module Cardano.Crypto.DSIGN.BLS12381 (
   SigDSIGN (..),
   PossessionProofDSIGN (..),
   BLS12381SignContext (..),
```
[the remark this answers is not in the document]
lehins: I have concidered tests when making the suggestion and they do not conflict with those tests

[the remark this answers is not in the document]
lehins: My point is we advice to only use `minVerKeyPoPDST` and `minSigPoPDST`, but we only test arbitrary stuff that we don't even recommend using.

If something needs to be exported for testing only, it should be re-exported from an internal module, but we need to make sure we test the actual API that we provide for users.

We can't do both: pretend like we provide a safe API and export things that allow to easily bypass this safety, in other words, statement like this becomes a lie:

> Why only the "PoP" ciphersuite is exported
>
> The main reason is API clarity and safety.

So, my suggestion is
* to make sure we have tests that use both of `minVerKeyPoPDST` and 
`minSigPoPDST`
* If you want to continue testing arbitrary contexts, that is fine as well, but then in order to do that we need to move: `src/Cardano/Crypto/DSIGN/BLS12381.hs‎` -> `src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs‎` and re-export everything, but `BLS12381SignContext` constructor and `blsSignContextDst` field accessor


Alternatively, we can:
* remove a lie from haddock promising "safe api"
* and still add tests for  `minVerKeyPoPDST` and `minSigPoPDST`

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -26,6 +26,8 @@ module Cardano.Crypto.DSIGN.BLS12381 (
...
+  minSigPoPDST,
+  minVerKeyPoPDST,
```
lehins: These need a mention in the changelog.
Also we are prepping a release right now, so this PR needs a version bump: https://github.com/IntersectMBO/cardano-haskell-packages/pull/1343
If you'd like to include these changes in this release we can wait.

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -130,16 +131,116 @@ data BLS12381DSIGN curve
...
+--
+-- * 'minVerKeyPoPDST' selects the __minimal-pubkey-size__ variant:
+--   public keys live in G1 (48 bytes compressed), signatures in G2
+--   (96 bytes compressed).
+--
+-- The draft recommends the minimal-pubkey-size variant for aggregation,
+-- because the size of @(PK_1, ..., PK_n, signature)@ is usually dominated by
+-- the public keys. Other protocols, like Leios, might favor minimal-signature-size.
+--
+-- == Example
```
lehins: I wonder if it would be hard to convert these examples to actual doctest?
It would nice to ensure they don't go out of sync with implementation

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
perturbing: Lets add them, will do it now

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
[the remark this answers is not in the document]
perturbing: What about a dedicated helper function? See eb874776f1a5802494ec91f8abcef3431e7da212

> but we only test arbitrary stuff that we don't even recommend using.

True, but I want to be assured that the C function is not faulty over its domain (though we only select on value in it).

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
```diff
@@ -25,7 +25,10 @@ module Cardano.Crypto.DSIGN.BLS12381 (
   SignKeyDSIGN (..),
   SigDSIGN (..),
   PossessionProofDSIGN (..),
-  BLS12381SignContext (..),
+  BLS12381SignContext,
+  mkBLS12381SignContext,
```
lehins: We want to prevent arbitrary construction, but still allow access to the fields. While this change does the exact opposite
```suggestion
```

perturbing: ill switch to the internal module structure :+1:

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
lehins, line 1: Could I ask you to to split this commit into two:
1. move the `Cardano/Crypto/DSIGN/BLS12381.hs` -> `Cardano/Crypto/DSIGN/BLS12381/Internal.hs`
2. Creates new ``Cardano/Crypto/DSIGN/BLS12381.hs` that does the re-exports

With this approach we will not loose the git history of the file

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
```diff
@@ -0,0 +1,593 @@
...
+
+-- | The BLS12381 signing context for the "PoP" based ciphersuite for the minimal verification key size variant of bls signatures
+minVerKeyPoPDST :: BLS12381SignContext
+minVerKeyPoPDST = BLS12381SignContext (Just "BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_") Nothing
+
+type family CurveVariant (c :: Type) :: Symbol where
+  CurveVariant Curve1 = "BLS-Signature-Mininimal-Verification-Key-Size"
+  CurveVariant Curve2 = "BLS-Signature-Mininimal-Signature-Size"
+
+-- | This module advicese to only the proof-of-possession (PoP) ciphersuite
```
lehins: I am not a native English speaker, but `advicese` doesn't sound like a real word to me?

Is this what was intended? 
```suggestion
-- | This module advises to only the proof-of-possession (PoP) ciphersuite
```

Maybe this would be more clear?
```suggestion
-- | This module provides support only for proof-of-possession (PoP) ciphersuite
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
lehins: We don't need to do that now. I'll create a ticket for this task, so we can see if it is not too hard to change those into doctests

lehins (changes requested):

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
perturbing: I tried to do this, but my Nix setup is not working, so I am currently unable to get this done :(

lehins: No worries, this isn't urgent or necessary. It is more of a nice to have kind of thing

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
perturbing: Hehe, I am also not a native speaker, but you are correct; it is advises!

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381.hs
lehins: FTR: #647

cardano-crypto-class/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ## 2.4.0.0
 
+* Remove constructors of `BLS12381SignContext` from export; use `minSigPoPDST` or `minVerKeyPoPDST` for the standard PoP ciphersuites.
```
lehins: This entry now needs to go in the section above, which needs to be bumped to `2.5.0.0`

lehins (approved):
Awesome! Thank you!

PR #638, opened 2026-04-01
Add TreeDiff utilities to cardano-binary testlib

MavenRain:
Toward #5360 (Part 1)


MavenRain:
> IntersectMBO/cardano-ledger#5360

Yes I did.  My apologies.  I'll correct the description.

cardano-binary/testlib/Test/Cardano/Binary/TreeDiff.hs
```diff
@@ -1,27 +1,127 @@
...
+  ansiWlPretty,
+
+  -- * Re-exports from ImpSpec
+  ansiDocToString,
+  assertColorFailure,
+  callStackToLocation,
+  srcLocToLocation,
+) where
 
 import qualified Cardano.Binary as Plain
```
lehins, line 52: There is no versioned decoder in cardano-base, so it makes sense to remove sense to remove qualified import as `Plain`
```suggestion
import Cardano.Binary
```

cardano-binary/testlib/Test/Cardano/Binary/TreeDiff.hs
```diff
@@ -1,27 +1,127 @@
...
+  Pretty (..),
+  Doc,
+  AnsiStyle,
+  ansiWlPretty,
+
+  -- * Re-exports from ImpSpec
+  ansiDocToString,
+  assertColorFailure,
+  callStackToLocation,
+  srcLocToLocation,
```
lehins, line 49: We should not be re-exporting stuff from other libraries.

```suggestion
```

cardano-binary/testlib/Test/Cardano/Binary/TreeDiff.hs
```diff
@@ -1,27 +1,110 @@
+{-# LANGUAGE ImplicitParams #-}
```
Soupstraw, line 1:
```suggestion
```
This seems to be unused

PR #652, opened 2026-05-08
QuickCheck compatibility shim

erikd:
# Description

* Fix doctest for ghc-9.14 and add it to CI
* Update cabal to 3.16 on CI and in nix
* Update index-state for CHaP and Hackage
* Introduce compatibility with `QuickCheck-2.18`

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -62,7 +62,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["9.6.7", "9.8.4", "9.10.2", "9.12.2"]
+        ghc: ["9.6.7", "9.8.4", "9.10.2", "9.12.2", "9.14.1"]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     env:
@@ -75,7 +75,7 @@ jobs:
       id: setup-haskell
       with:
         ghc-version: ${{ matrix.ghc }}
-        cabal-version: 3.14.1.0
+        cabal-version: 3.16.1.0
 
     - name: Install system dependencies
       uses: input-output-hk/actions/base@latest
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -192,7 +192,7 @@ library testlib
     base,
     base16-bytestring,
     bytestring >=0.10.12.0,
-    cardano-base:testlib,
+    cardano-base:testlib >=0.1.4,
     cardano-binary,
     cardano-crypto-class,
     cborg,
--- a/cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE FlexibleContexts #-}
@@ -152,6 +153,25 @@ deriving instance HashAlgorithm h => Storable (Hash h a)
 -- >>> let hash = $$("0123456789abcdef") :: Hash ShortHash ()
 -- >>> print hash
 -- "0123456789abcdef"
+#if __GLASGOW_HASKELL__ >= 914
+-- >>> let hash = $$("deadbeef") :: Hash ShortHash ()
+-- ...
+--     • <Hash blake2b_prefix_8>: Expected in decoded form to be: 8 bytes, but got: 4
+--     • In the typed Template Haskell splice: $$("deadbeef")
+-- ...
+--       In the expression: $$("deadbeef") :: Hash ShortHash ()
+--       In an equation for ‘hash’:
+--           hash = $$("deadbeef") :: Hash ShortHash ()
+-- ...
+-- >>> let hash = $$("123") :: Hash ShortHash ()
+-- ...
+--     • <Hash blake2b_prefix_8>: Malformed hex: invalid bytestring size
+--     • In the typed Template Haskell splice: $$("123")
+-- ...
+--       In the expression: $$("123") :: Hash ShortHash ()
+--       In an equation for ‘hash’: hash = $$("123") :: Hash ShortHash ()
+-- ...
+#else
 -- >>> let hash = $$("deadbeef") :: Hash ShortHash ()
 -- ...
 --     • <Hash blake2b_prefix_8>: Expected in decoded form to be: 8 bytes, but got: 4
@@ -167,6 +187,7 @@ deriving instance HashAlgorithm h => Storable (Hash h a)
 --       In the expression: $$("123") :: Hash ShortHash ()
 --       In an equation for ‘hash’: hash = $$("123") :: Hash ShortHash ()
 -- ...
+#endif
 instance HashAlgorithm h => IsString (Q (TExp (Hash h a))) where
   fromString hexStr = do
     let n = fromInteger $ natVal (Proxy @(HashSize h))
--- a/cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
+++ b/cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
@@ -28,7 +28,6 @@ import Test.QuickCheck (
   forAllShow,
   ioProperty,
   counterexample,
-  withMaxSuccess,
   )
 import Test.Hspec (Spec, describe, it, shouldBe)
 import Test.Hspec.QuickCheck (prop, modifyMaxSuccess)
@@ -89,6 +88,7 @@ import Cardano.Crypto.EllipticCurve.BLS12_381 (Curve1, Curve2, blsCompress, blsG
 import Cardano.Crypto.EllipticCurve.BLS12_381.Internal (blsZero)
 import Cardano.Crypto.PinnedSizedBytes (PinnedSizedBytes)
 import Cardano.Crypto.DirectSerialise
+import Test.Cardano.Base.QuickCheck (withNumTests)
 import Test.Crypto.Util (
   BadInputFor,
   genBadInputFor,
@@ -827,14 +827,14 @@ testDSIGNAggregatableWithContext _ genContext genKeyCtx genMsg name = testEnough
         forAllPoP $ prop_cbor_direct_vs_class encodePossessionProofDSIGN
   describe "aggregate" $ do
     prop "aggregate verify positive" $
-      withMaxSuccess 1000 .
+      withNumTests 1000 .
       forAllShow (genAggregateCase genContext genMsg) ppShow $
           \(ctx, msg, vksPops, sigs) -> (=== Right ()) $ do
             sig <- aggregateSigsDSIGN @v sigs
             aggVk <- aggregateVerKeysDSIGN ctx vksPops
             verifyDSIGN @v ctx aggVk msg sig
     prop "aggregate verify negative (wrong message)" $
-      withMaxSuccess 1000 .
+      withNumTests 1000 .
       forAllShow (genAggregateCase genContext genMsg) ppShow $ \(ctx, msg, vksPops, sigs) ->
           forAllShow arbitrary ppShow $ \msg' ->
             msg /= msg' ==> (=/= Right ()) $ do
--- a/cabal.project
+++ b/cabal.project
@@ -11,8 +11,8 @@ repository cardano-haskell-packages
 
 -- The index-states
 index-state:
-  , hackage.haskell.org 2026-03-07T07:19:41Z
-  , cardano-haskell-packages 2026-01-26T18:50:24Z
+  , hackage.haskell.org 2026-05-07T19:55:27Z
+  , cardano-haskell-packages 2026-05-05T04:28:05Z
 
 packages:
   base-deriving-via
--- a/cabal.project
+++ b/cabal.project
@@ -41,30 +41,18 @@ package bitvec
   -- Workaround for windows cross-compilation
   flags: -simd
 
--- cabal-allow-newer
+-- cabal-allow-newer begin
 if impl(ghc >=9.14)
   allow-newer:
-    , aeson:base
-    , aeson:bytestring
-    , aeson:containers
-    , aeson:deepseq
-    , aeson:ghc-prim
-    , aeson:template-haskell
-    , aeson:time
-    , binary-orphans:base
     , boring:base
+    , canonical-json:containers
     , cborg:base
-    , indexed-traversable:base
-    , indexed-traversable-instances:base
-    , microstache:base
-    , parsec:base
-    , quickcheck-instances:base
-    , quickcheck-instances:bytestring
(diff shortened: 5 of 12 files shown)


cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
```diff
@@ -827,14 +827,14 @@ testDSIGNAggregatableWithContext _ genContext genKeyCtx genMsg name = testEnough
         forAllPoP $ prop_cbor_direct_vs_class encodePossessionProofDSIGN
   describe "aggregate" $ do
     prop "aggregate verify positive" $
-      withMaxSuccess 1000 .
+      withNumTests 1000 .
```
erikd, line 830: Bumped the min version on QuickCheck as well.

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -188,7 +188,7 @@ library testlib
   build-depends:
     FailT,
     HUnit,
-    QuickCheck,
+    QuickCheck >=2.18,
```
lehins: We can't all of a sudden enforce everyone to use latest version of QuickCheck.
Other components might have dependencies that are not compatible with QuickCheck and will not be for a while.

So, if your goal is to add compatibility with QuickChekc-2.18, then some CPP is in order.
I recommend adding a module to `cardano-base:testlib:Test.Cardano.Base.QuickCheck` with this function:
```haskell
#if !MIN_VERSION_QuickCheck(2, 18, 0)
withNumTests = withMaxSuccess
#endif
```
Then you could use it in this repo and in other components without worrying about bounds for QuickCheck

.github/workflows/haskell.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ["9.6.7", "9.8.4", "9.10.2", "9.12.2"]
+        ghc: ["9.6.7", "9.8.4", "9.10.2", "9.12.2", "9.14.1", ]
```
lehins:
```suggestion
        ghc: ["9.6.7", "9.8.4", "9.10.2", "9.12.2", "9.14.1"]
```

lehins (changes requested):

cardano-crypto-class/cardano-crypto-class.cabal
erikd: Done

.github/workflows/haskell.yml
erikd: Done.

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
```diff
@@ -894,3 +898,9 @@ testDSIGNAggregatableWithContext _ genContext genKeyCtx genMsg name = testEnough
                    ]
           sigs = [ signDSIGN ctx msg sk | sk <- sks ]
       pure (ctx, msg, vksPops, sigs)
+
+#if !MIN_VERSION_QuickCheck(2, 18, 0)
+withNumTests :: Int -> Property -> Property
+withNumTests = withMaxSuccess
```
lehins: I meant to put it here: https://github.com/IntersectMBO/cardano-base/tree/master/cardano-base/testlib/Test/Cardano/Base

I am fine merging this PR as is for now, but the point is you will run into the exact same problem in Ledger, Consensus, Network, etc. And with a hell a lot of more occurrences than this one. You'll be tired of adding the same function over and over in each repo in some random place. So, I'd suggest to put it in `cardano-base` package, in `testlib` in `Test.Cardano.Base.QuickCheck` module and let it be reused by all downstream packages.

lehins (changes requested):

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
erikd: I will fix it.

erikd: @lehins This is actually rather subtle and leads to subtle problems.

We can put `withNumTests` in `Test.Cardano.Base.QuickCheck` but should it be unconditionally exported? Both an uncondiftional and a conditional export causes issues.

If we export `withNumTests` unconditionally from  `Test.Cardano.Base.QuickCheck` we will get a duplicate identifier error for `QuickCheck > 2.18` requiring a CPP conditional `hiding (withNumTests)`.

If we export `withNumTests` conditionally from  `Test.Cardano.Base.QuickCheck` we will get a duplicate identifier error for `QuickCheck > 2.18` about the `Test.Cardano.Base.QuickCheck` import being redundant.

The QuickCheck devs should not have renamed this function and immediately deprecated the old name. They should have waited a couple of major releases before deprecating the old name.

Although far from ideal, I think the least painful solution is to use `QuickCheck >= 2.18` but I am happy to hear if you have a better suggestion. I would also like to know what you prefer:

* Unconditional export of `withNumTests` from  `Test.Cardano.Base.QuickCheck`.
* Conditional export of `withNumTests` from  `Test.Cardano.Base.QuickCheck`.
* Use `QuickCheck >= 2.18`.
* Something else.

erikd: Actually my commit below titled "Work around QuickCheck deprcation" has the neatest solution I can currently come up with.

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -170,6 +170,7 @@ library testlib
   hs-source-dirs: testlib
   visibility: public
   exposed-modules:
+    Test.Cardano.Base.QuickCheck
```
lehins: Looks like you forgot to `git add` the module

lehins (changes requested):

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
lehins: This sort of problem is always very easily solvable with `qualified` imports:
```haskell
module Test.Cardano.Base.QuickCheck (withNumTests) where

import qualified Test.QuickCheck as QC

withNumTests :: Int -> QC.Property -> QC.Property
#if MIN_VERSION_QuickCheck(2, 18, 0)
withNumTests = QC.withNumTests
#else
withNumTests = QC.withMaxSuccess
#endif
```

lehins: > The QuickCheck devs should not have renamed this function and immediately deprecated the old name. They should have waited a couple of major releases before deprecating the old name.

~I 100% agree with this! Although this wouldn't help us in this particula situation too much, since deprecation warnings would result in CI not passing. However, it would, at least, not cause all of the current code we have on CHaP to fail of a sudden.~
My apologies, I misread your message as "The QuickCheck devs should not have **removed** this function". 
QuickCheck devs done exactly as they were suppose to. Function is still there and gives you a warning that it should not be used anymore

erikd:
My end of day here. Will continue tomorrow.

cardano-crypto-class/testlib/Test/Cardano/Base/QuickCheck.hs
```diff
@@ -0,0 +1,25 @@
+{-# LANGUAGE CPP #-}
+module Test.Cardano.Base.QuickCheck
+  ( withMaxSuccess
```
lehins: This is an incorrect solution, because it continues to use the old name.
The point about deprecating a function is so all users downstream can switch to a new function. With this solution you just postponing the work until later
```suggestion
  ( withNumTests
```

lehins (changes requested):

cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
```diff
@@ -152,6 +153,25 @@ deriving instance HashAlgorithm h => Storable (Hash h a)
...
+-- ...
+-- >>> let hash = $$("123") :: Hash ShortHash ()
+-- ...
+--     • <Hash blake2b_prefix_8>: Malformed hex: invalid bytestring size
+--     • In the typed Template Haskell splice: $$("123")
+-- ...
+--       In the expression: $$("123") :: Hash ShortHash ()
+--       In an equation for ‘hash’: hash = $$("123") :: Hash ShortHash ()
+-- ...
+#else
```
neilmayhew, line 174: I think this can be handled better by omitting the redundant text from the error message and letting it be matched by `...` instead:

```haskell
-- >>> let hash = $$("deadbeef") :: Hash ShortHash ()
-- ...
--     • <Hash blake2b_prefix_8>: Expected in decoded form to be: 8 bytes, but got: 4
-- ...
-- >>> let hash = $$("123") :: Hash ShortHash ()
-- ...
--     • <Hash blake2b_prefix_8>: Malformed hex: invalid bytestring size
-- ...
```

The same applies below.

cardano-crypto-class/testlib/Test/Cardano/Base/QuickCheck.hs
```diff
@@ -0,0 +1,25 @@
...
+  )
+where
+
+-- QuickCheck 2.18 replaces `withMaxSuccess` with `withNumTests` and
+-- immediately deprecates the latter.
+-- We handle this for the whole Cardano stack here. Any other code
+-- that uses withMaxSuccess should import this module and switch to
```
neilmayhew: "or switch"

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
neilmayhew: You probably need to take this out, otherwise the code won't compile with later versions of QuickCheck

As it is, I get

```
    Ambiguous occurrence ‘withMaxSuccess’.
    It could refer to
       either ‘Test.QuickCheck.withMaxSuccess’,
              imported from ‘Test.QuickCheck’ at testlib/Test/Crypto/DSIGN.hs:31:3-16
              (and originally defined in ‘Test.QuickCheck.Property’),
           or ‘Test.Cardano.Base.QuickCheck.withMaxSuccess’,
              imported from ‘Test.Cardano.Base.QuickCheck’ at testlib/Test/Crypto/DSIGN.hs:92:1-35
              (and originally defined
                 at testlib/Test/Cardano/Base/QuickCheck.hs:24:1-14).
```

when I `cabal build` in `nix develop .#ghc9141`

cardano-crypto-class/testlib/Test/Cardano/Base/QuickCheck.hs
```diff
@@ -0,0 +1,25 @@
...
+  withNumTests,
+#else
+  withMaxSuccess,
+#endif
+  )
+
+#if MIN_VERSION_QuickCheck(2, 18, 0)
+withMaxSuccess :: Int -> Property -> Property
+withMaxSuccess = withNumTests
+#endif
```
neilmayhew: Building on @lehins comment above,

```suggestion
#if !MIN_VERSION_QuickCheck(2, 18, 0)
withNumTests :: Int -> Property -> Property
withNumTests = withMaxSuccess
#endif
```

neilmayhew: Of course, you'd also need to change `withMaxSuccess` to `withNumTests` in two places in `cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs`

neilmayhew (changes requested):

cardano-crypto-class/testlib/Test/Cardano/Base/QuickCheck.hs
neilmayhew: Actually, with the other change I suggested, this should probably be "and" after all.

However, you could mention that if the project is able to use a `>=2.18` constraint then importing this module isn't needed.

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
erikd: That code version was broken. Should be fine now.

cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
erikd: Not sure how this relates to this PR.

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
```diff
@@ -28,7 +28,9 @@ import Test.QuickCheck (
   forAllShow,
   ioProperty,
   counterexample,
-  withMaxSuccess,
+#if MIN_VERSION_QuickCheck(2, 18, 0)
+  withNumTests,
+#endif
```
lehins: The point of adding `Test.Cardano.Base.QuickCheck` is that you no longer need to use CPP anywhere else to solve this problem

cardano-crypto-class/testlib/Test/Crypto/DSIGN.hs
```diff
@@ -89,6 +91,9 @@ import Cardano.Crypto.EllipticCurve.BLS12_381 (Curve1, Curve2, blsCompress, blsG
 import Cardano.Crypto.EllipticCurve.BLS12_381.Internal (blsZero)
 import Cardano.Crypto.PinnedSizedBytes (PinnedSizedBytes)
 import Cardano.Crypto.DirectSerialise
+#if !MIN_VERSION_QuickCheck(2, 18, 0)
+import Test.Cardano.Base.QuickCheck
+#endif
```
lehins:
```suggestion
import Test.Cardano.Base.QuickCheck (withNumTests)
```

cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
lehins: It relates to this PR in a way that with @neilmayhew's suggestion you do not need to use CPP, which is always a win

cardano-crypto-class/cardano-crypto-class.cabal
lehins: This module needs to go into `cardano-base` package, not into `cardano-crypto-class` package!
https://github.com/IntersectMBO/cardano-base/blob/master/cardano-base/cardano-base.cabal

It will also require a:
* changelog entry 
* a version bump to `0.1.4.0` I
* and `cardano-crypto-class` will need a lower bound `cardano-base:testlib >= 0.1.4`

lehins (changes requested):
Still needs work

erikd:
* Moved `Test.Cardano.Base.QuickCheck` module to `cardano-base/testlib`
* Bumped version of `cardano-base` to `0.1.4` and updated `cardano-base` changelog.
* `cardano-crypto-class:testlib` now depends on `cardano-base:testlib >= 0.1.4`.
* Removed some unneeded CPP.

cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
lehins: That's ok though, we can change it at some point later, since using CPP also does fix the compatibility.

lehins (approved):
Looks great. Thank you!

neilmayhew (approved):

PR #653, opened 2026-05-11
feat: add cardano-crypto-wallet package

disassembler:
Introduces a new package providing authenticated envelope wrapping for HD wallet extended private keys.

cbits are vendored from cardano-crypto. This is done in the first commit which can be largely ignored by reviewers.

Key design:
- C layer handles ed25519 key generation, public-key derivation, signing, and BIP32 child derivation; it stores and operates on plaintext key material only; public symbols are prefixed cardano_ to avoid linker conflicts with the legacy cardano-crypto C library
- Haskell layer owns all encryption: Argon2id key derivation (128 MiB, t=3, p=4) and XChaCha20-Poly1305 AEAD wrapping with a freshly randomized salt and nonce per write
- CBOR-encoded versioned envelope with explicit associated data binding format version, KDF parameters, cipher, payload shape, public key, and chain code — preventing silent swaps without detection
- Fail-closed public API: passphrase-using operations return Either and reject wrong passphrases at authentication time
- Test helpers for fast-KDF and deterministic randomness modes keep the test suite sub-second without touching production paths
- Benchmark suite covers create, validate, sign, and change-passphrase under production Argon2id parameters

# Description

# Checklist

- [ ] Self-reviewed the diff


AndrewWestberg (approved):

cardano-crypto-wallet/testlib/Test/Cardano/Crypto/WalletHD/Instances.hs
```diff
@@ -0,0 +1,27 @@
...
+
+instance Arbitrary DerivationScheme where
+  arbitrary = elements [DerivationScheme1, DerivationScheme2]
+
+-- | Generates a v2-wrapped 'EncryptedKey' from a random 32-byte seed and
+-- chain code, always encrypted with the empty passphrase.
+instance Arbitrary EncryptedKey where
+  arbitrary = do
+    seed <- BS.pack <$> replicateM 32 arbitrary
+    cc <- BS.pack <$> replicateM 32 arbitrary
```
lehins: We have a helper for this that will do a much better job at generating random `ByteString`:
https://github.com/IntersectMBO/cardano-base/blob/7a8a991945d401d89e27f53b3d3bb464a354ad4c/cardano-base/testlib/Test/Cardano/Base/Bytes.hs#L38-L39
```suggestion
    seed <- genByteString 32
    cc <- genByteString 32
```

erikd:
I am not in any way a crypto guy, but I am a Haskell guy and I did spend decades coding in C, but have not done an C in over a decade. I will review this, but it should also be reviewed by a crypto guy that also knows Haskell and C.

For the Haskell code:
* All top level constants initialized with `unsafePerformIO` tagged with a `NOINLINE` pragma. :heavy_check_mark:
* Sensitive data held in `ScrubbedBytes`. :heavy_check_mark: . Possibly also store `salt` and `nonce` to `ScrubbedBytes`. :question: 
* Maybe use `{-# LANGUAGE Strict #-}` to ensure sensitive data is not held in an un-evaludated thunk longer than it should be. :question: 
* This package currently depends on the `memory` package which i think is unmaintained. The `ram` package is supposed to be a drop in replacement. :question: 

For the C code (`cardano-crypto-wallet/cbits/encrypted_sign.c` only):
* Seems fine. :heavy_check_mark:

cardano-crypto-wallet/testlib/Test/Cardano/Crypto/WalletHD/Instances.hs
```diff
@@ -0,0 +1,27 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | QuickCheck 'Arbitrary' instances for wallet types.
+--
+-- 'Arbitrary' 'EncryptedKey' always produces v2 envelopes wrapped with the
+-- empty passphrase so that tests can call passphrase-dependent operations
+-- without tracking a per-key passphrase.
+module Test.Cardano.Crypto.WalletHD.Instances () where
```
lehins: Convention is to call this module `Arbitrary` (we need to do the same rename in `cardano-crypto-class`)
```suggestion
module Test.Cardano.Crypto.WalletHD.Arbitrary () where
```

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,873 @@
...
+-- | Replace system randomness with a deterministic counter for reproducible
+-- test output.
+withDeterministicRandomnessForTesting :: IO a -> IO a
+withDeterministicRandomnessForTesting = bracket install restore . const
+  where
+    install = do
+      original <- readIORef randomModeRef
+      writeIORef randomModeRef (DeterministicRandom 0)
+      pure original
+    restore original = writeIORef randomModeRef original
```
lehins, line 203: Anything that is needed for testing only should go into `testlib`, that is the whole the reason for its existence. Let's leave it here for now and I'l add a new module `Test.Cardano.Crypto.WalletHD.RandomMode` and put it there in a follow up PR.

lehins:
I think this is a very useful for reviewing this PR summary from Claude about the migrated C files:

<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">The PR vendors 21 files into <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano-crypto-wallet/cbits/</code> (2 <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">.c</code> + 19 <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">.h</code>), all originally from <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano-crypto/cbits/</code>. I diffed each against its counterpart on <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">master</code> of <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">IntersectMBO/cardano-crypto</code>.</p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><strong>Per-file status:</strong></p>
<div class="overflow-x-auto w-full px-2 mb-6">

<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">Of the 21 vendored files, 10 are byte-identical, 9 have whitespace-only differences, and <strong>2 have real changes</strong>.</p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><strong><code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">ed25519-hash.h</code></strong> — the SHA-512 backend is swapped. Previously it called the vendored <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">crypton_sha512_{init,update,finalize}</code> from <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">crypton_sha512.h</code>; now it includes <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">&lt;sodium.h&gt;</code> and calls <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">crypto_hash_sha512_{init,update,final}</code>, with <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">memset</code> replaced by <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">sodium_memzero</code> for the context wipe.</p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><strong><code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">encrypted_sign.c</code></strong> — substantially rewritten. The semantic changes:</p>
<ul class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-disc flex flex-col gap-1 pl-8 mb-3">
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>All passphrase-based encryption removed from C.</strong> The original file did Argon2-replacement-style key stretching (<code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">crypton_fastpbkdf2_hmac_sha512</code>, 15000 iterations, fixed salt <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">"encrypted wallet salt"</code>) and encrypted/decrypted the secret key with ChaCha20 inside C (<code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">crypton_chacha_init</code>/<code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">crypton_chacha_combine</code>). All of that — <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">memory_combine</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">unencrypt_start</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">unencrypt_stop</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">stretch</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">clear</code>, the <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">SYM_*</code> constants — is deleted. The C struct's <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">ekey</code> field now holds <strong>plaintext</strong> secret-key bytes; encryption moves to the Haskell layer (Argon2id + XChaCha20-Poly1305, per the PR description).</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>All public functions renamed with <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_</code> prefix</strong> (<code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_from_secret</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_sign</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_derive_private</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_derive_public</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_new_from_mkg</code>) to avoid linker conflicts with the legacy <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano-crypto</code> library if both end up linked.</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>All passphrase parameters dropped.</strong> Signatures like <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">wallet_encrypted_sign(encrypted_key, pass, pass_len, data, …)</code> become <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_sign(encrypted_key, data, …)</code>. <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">wallet_encrypted_change_pass</code> is gone entirely (passphrase rotation now happens in Haskell).</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong><code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">wallet_encrypted_change_pass</code> is replaced by <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_decrypt</code></strong>, which doesn't decrypt anything in the crypto sense — it just verifies that the stored public key matches the secret key (recomputing it with <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_crypto_ed25519_publickey</code> and comparing via <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">sodium_memcmp</code>).</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>HMAC backend swapped.</strong> Vendored <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">DECL_HMAC(sha512, ...)</code> and <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">HMAC_sha512_{init,update,final}</code> are replaced by libsodium's <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">crypto_auth_hmacsha512_{init,update,final}</code>. The <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">#include &lt;hmac.h&gt;</code> is dropped.</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>Memory hygiene tightened.</strong> A new static helper <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">secure_clear</code> uses <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">sodium_memzero</code> instead of the old <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">clear</code> (which only called <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">memset</code> with a FIXME comment about the compiler optimising it away). Every function now zeroes its local secret-key buffers, HMAC contexts, index buffers, and intermediate <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">z</code> / <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">hmac_out</code> arrays before returning.</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>Return codes added.</strong> <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_sign</code> and <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_derive_private</code> now return <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">int</code> (always 0); <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">cardano_wallet_encrypted_from_secret</code> zeros the secret-key buffer on the error path.</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>Five new libsodium wrapper functions appended</strong> for the Haskell FFI to call: <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">wallet_sodium_randombytes</code>, <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">wallet_sodium_argon2id</code> (Argon2id at output length 32, configurable opslimit/memlimit), and <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">wallet_sodium_xchacha20poly1305_encrypt</code>/<code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">_decrypt</code> (both hardcoded to a 64-byte plaintext, i.e. <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">ENCRYPTED_KEY_SIZE</code>). Each calls <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">sodium_init()</code> defensively via <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">ensure_sodium</code>.</li>
</ul>

disassembler:
Good summary, two corrections:

  1. Four new libsodium wrapper functions, not five: wallet_sodium_randombytes,
   wallet_sodium_argon2id, wallet_sodium_xchacha20poly1305_encrypt,
  wallet_sodium_xchacha20poly1305_decrypt.
  2. cardano_wallet_encrypted_sign recomputes the public key via
  cardano_crypto_ed25519_publickey(in->ekey, pub_key) rather than using the
  cached in->pkey. Deliberate — signing always derives the pubkey fresh from
  the secret, so a corrupted pkey field can't produce a mismatched signature.
  The separate cardano_wallet_encrypted_decrypt exists for out-of-band
  consistency checks, not as a signing precondition.

  Rest is accurate.

cardano-crypto-wallet/cbits/encrypted_sign.c
```diff
@@ -0,0 +1,452 @@
...
+	memcpy(cc_out, hmac_out + (sizeof(hmac_out) - CHAIN_CODE_SIZE), CHAIN_CODE_SIZE);
+	secure_clear(z, 64);
+	secure_clear(hmac_out, 64);
+	secure_clear(idxBuf, sizeof(idxBuf));
+	secure_clear(&hmac_ctx, sizeof(hmac_ctx));
+
+	return 0;
+}
+
+int wallet_sodium_randombytes(uint8_t *out, uint32_t out_len)
```
lehins: Types do not match up with `randombytes_buf`
```suggestion
int wallet_sodium_randombytes(void * const out, size_t const out_len)
```

cardano-crypto-wallet/cbits/encrypted_sign.c
```diff
@@ -0,0 +1,452 @@
...
+{
+	if (ensure_sodium() != 0) {
+		return 1;
+	}
+	randombytes_buf(out, (size_t) out_len);
```
lehins: Casting no longer necessary
```suggestion
	randombytes_buf(out, out_len);
```

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+    PublicKeyPtr ->
+    ChainCodePtr ->
+    DerivationIndex ->
+    PublicKeyPtr ->
+    ChainCodePtr ->
+    CDerivationScheme ->
+    IO CInt
+
+foreign import ccall "wallet_sodium_randombytes"
+  wallet_sodium_randombytes :: Ptr Word8 -> Word32 -> IO CInt
```
lehins: This is not quite correct typing. Haskell's [`Word32` is really `HsWord32` on the C side](https://downloads.haskell.org/ghc/latest/docs/libraries/ghc-internal-9.1401.0-555c/src/GHC.Internal.Word.html#Word32).
However, we do not need it to be `HsWord32` on the C side, we need it to be `size_t`, therefore it is better to use [`CSize`](https://downloads.haskell.org/ghc/latest/docs/libraries/base-4.22.0.0-66f8/Foreign-C-Types.html#t:CSize) type here
```suggestion
  wallet_sodium_randombytes :: Ptr a -> CSize -> IO CInt
```

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+  if (memoryKiB, timeCost, parallelism, outputLength)
+    /= ( productionArgonMemoryKiB
+       , productionArgonTimeCost
+       , productionArgonParallelism
+       , productionArgonOutputLength
+       )
+    then failDecoder XPrvInvalidKdfParams
+    else pure ()
+  salt <- decodeBytes
+  if BS.length salt /= saltSize then failDecoder XPrvInvalidSaltLength else pure ()
```
lehins: Please replace all calls to `if x then y else pure ()` with `when x y`
```suggestion
  when (BS.length salt /= saltSize) $ failDecoder XPrvInvalidSaltLength
```

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+                memBytes
+      pure (if status == 0 then Right key else Left XPrvInternalError)
+
+randomBytesIO :: Int -> IO (Either XPrvError ByteString)
+randomBytesIO len = do
+  mode <- readIORef randomModeRef
+  case mode of
+    SystemRandom -> do
+      (status, bytes) <- B.allocRet len $ \out ->
+        wallet_sodium_randombytes out (fromIntegral len)
```
lehins: We need to annotate calls to `fromIntegral` with concrete types, which make possible integer overflows obvious.
```suggestion
        wallet_sodium_randombytes out (fromIntegral @Int @CSize len)
```

cardano-crypto-wallet/test/Main.hs
```diff
@@ -0,0 +1,21 @@
...
+ )
+import qualified Test.Cardano.Crypto.Wallet.RoundTripSpec as RoundTrip
+import qualified Test.Cardano.Crypto.Wallet.SignSpec as Sign
+import qualified Test.Cardano.Crypto.Wallet.V2FormatSpec as V2Format
+
+main :: IO ()
+main = do
+  sodiumInit
+  withFastKdfForTesting . withDeterministicRandomnessForTesting $
+    hspec $ do
```
lehins: The correct way to set up the environment for all tests with hspec is this way:
 ```suggestion
main = do
  let withInit runTests = do
        sodiumInit
        withFastKdfForTesting $ withDeterministicRandomnessForTesting runTests
    hspec $ aroundAll_ withInit $ do
```

File: cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs [hunk 10]
```diff
@@ -0,0 +1,909 @@
...
+  memoryKiB <- decodeWord
+  timeCost <- decodeWord
+  parallelism <- decodeWord
+  outputLength <- decodeWord
+  if (memoryKiB, timeCost, parallelism, outputLength)
+    /= ( productionArgonMemoryKiB
+       , productionArgonTimeCost
+       , productionArgonParallelism
+       , productionArgonOutputLength
+       )
```
lehins: Using tuples adds unnecessary overhead and makes the code less readable, since it is harder to see what is being compared to what
```suggestion
  if memoryKiB /= productionArgonMemoryKiB
     || timeCost /= productionArgonTimeCost
     || parallelism /= productionArgonParallelism
     || outputLength /= productionArgonOutputLength
```

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+                  withByteArray aad $ \ad ->
+                    withByteArray (v2Nonce envelope) $ \np ->
+                      withByteArray wrappingKey $ \kp ->
+                        wallet_sodium_xchacha20poly1305_decrypt
+                          (coerce ptextPtr)
+                          (coerce ct)
+                          (fromIntegral $ BS.length (v2Ciphertext envelope))
+                          (coerce tg)
+                          ad
+                          (fromIntegral $ BS.length aad)
```
lehins: This is susceptible to integer overflows, because the type used is incorrect. It should be:
```suggestion
                          (fromIntegral @Int @CULLong $ BS.length aad)
```

cardano-crypto-wallet/cbits/encrypted_sign.c
```diff
@@ -0,0 +1,452 @@
...
+
+int wallet_sodium_xchacha20poly1305_encrypt(
+	uint8_t *ciphertext,
+	uint8_t tag[crypto_aead_xchacha20poly1305_ietf_ABYTES],
+	uint8_t const *plaintext,
+	uint32_t plaintext_len,
+	uint8_t const *aad,
+	uint32_t aad_len,
+	uint8_t const nonce[crypto_aead_xchacha20poly1305_ietf_NPUBBYTES],
+	uint8_t const key[crypto_aead_xchacha20poly1305_ietf_KEYBYTES])
```
lehins, line 386: We must be using the same types as expected by libsodium!
Also on the Haskell side we should be using type designed specifically for C interface from `Foreign.C.Types` module, not the usual Haskell types.

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+randomBytesIO len = do
+  mode <- readIORef randomModeRef
+  case mode of
+    SystemRandom -> do
+      (status, bytes) <- B.allocRet len $ \out ->
+        wallet_sodium_randombytes out (fromIntegral len)
+      pure $ if status == 0 then Right bytes else Left XPrvInternalError
+    DeterministicRandom counter -> do
+      let bytes = deterministicBytes len counter
+      writeIORef randomModeRef (DeterministicRandom (counter + 1))
```
lehins, line 790: This isn't thread safe. Running tests in parallel will lead non-deterministic results.

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+                      withByteArray aad $ \ad ->
+                        withByteArray nonce $ \np ->
+                          withByteArray wrappingKey $ \kp ->
+                            wallet_sodium_xchacha20poly1305_encrypt
+                              (coerce outCipher)
+                              (coerce outTag)
+                              (coerce skPtr)
+                              (fromIntegral legacyKeySize)
+                              ad
+                              (fromIntegral $ BS.length aad)
```
lehins: Ditto
```suggestion
                              (fromIntegral @Int @CUULong $ BS.length aad)
```

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+  , kdfParallelism :: !Word
+  , kdfOutputLength :: !Word
+  }
+
+productionKdfParams, fastTestKdfParams :: KdfParams
+productionKdfParams = KdfParams 131072 3 4 32
+fastTestKdfParams = KdfParams 4096 1 1 32
+
+runtimeKdfParamsRef :: IORef KdfParams
+runtimeKdfParamsRef = unsafePerformIO (newIORef productionKdfParams)
```
lehins, line 152: Usage of global mutable variables for the purpose of testing like this is absolutely atrocious. Both here and in `randomModeRef`.
It is not thread safe, nor is it user friendly.
We can merge it as is and I'll fix it later

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+foreign import ccall "cardano_wallet_encrypted_from_secret"
+  wallet_encrypted_from_secret ::
+    SecretKeyPtr ->
+    ChainCodePtr ->
+    EncryptedKeyPtr ->
+    IO CInt
+
+foreign import ccall "cardano_wallet_encrypted_new_from_mkg"
+  wallet_encrypted_new_from_mkg ::
+    SecretKeyPtr ->
```
lehins: This is not a secret key. Secret key has 64 bytes, while master key has 96, because it has both secret key and chain code
```suggestion
    MasterKeyPtr ->
```

cardano-crypto-wallet/test/Test/Cardano/Crypto/Wallet/SignSpec.hs
```diff
@@ -0,0 +1,104 @@
...
+
+verifySignature :: BS.ByteString -> BS.ByteString -> Signature -> Bool
+verifySignature pub msg (Signature sig) = unsafePerformIO $
+  BS.useAsCStringLen msg $ \(mp, ml) ->
+    BS.useAsCString pub $ \pkp ->
+      BS.useAsCString sig $ \sigp ->
+        (== 0)
+          <$> c_ed25519_sign_open
+            (castPtr mp)
+            (fromIntegral ml)
```
lehins: At least in the test suite types are safe, but annotation for `fromIntegral` is a good habit to get into.
```suggestion
            (fromIntegral @Int @CSize ml)
```

cardano-crypto-wallet/output
lehins, line 1: What is this file. Doesn't look like it was meant to be added to the repo

cardano-crypto-wallet/NOTICE
```diff
@@ -0,0 +1,14 @@
+Copyright 2019-2021 Input Output (Hong Kong) Ltd.
```
lehins: Should we update the year?
```suggestion
Copyright 2019-2026 Input Output (Hong Kong) Ltd.
```

cardano-crypto-wallet/cardano-crypto-wallet.cabal
```diff
@@ -0,0 +1,110 @@
...
+  implementation.
+
+license: Apache-2.0
+license-files:
+  LICENSE
+  NOTICE
+
+author: IOHK
+maintainer: operations@iohk.io
+copyright: 2019-2025 IOHK
```
lehins: Living in the past 😄 
```suggestion
copyright: 2019-2026 IOHK
```

cardano-crypto-wallet/testlib/Test/Cardano/Crypto/WalletHD/Instances.hs
```diff
@@ -0,0 +1,28 @@
...
+
+-- | Generates a v2-wrapped 'EncryptedKey' from a random 32-byte seed and
+-- chain code, always encrypted with the empty passphrase.
+instance Arbitrary EncryptedKey where
+  arbitrary = do
+    seed <- BS.pack <$> replicateM 32 arbitrary
+    cc <- BS.pack <$> replicateM 32 arbitrary
+    case unsafePerformIO (encryptedCreate seed (BS.empty :: BS.ByteString) cc) of
+      Right k -> pure k
+      Left _ -> arbitrary
```
lehins:
```suggestion
      -- Approximately 50% of the time `encryptedCreate` will fail due to
      -- an invalid `cc`, since it is generated uniformly.
      - It is OK to retry half the time.
      Left _ -> arbitrary
```

cardano-crypto-wallet/cbits/encrypted_sign.c
```diff
@@ -0,0 +1,452 @@
...
+		return 1;
+	}
+	return crypto_pwhash(out,
+		(unsigned long long) out_len,
+		(const char *) pass,
+		(unsigned long long) pass_len,
+		salt,
+		(unsigned long long) opslimit,
+		(size_t) memlimit,
+		crypto_pwhash_ALG_ARGON2ID13);
```
lehins, line 375: I don't understand. How is this ever a good idea to cast a type first to a smaller type only later to cast back to an even bigger type?
We must align our cbits and FFI to match libsodium types

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+-- ---------------------------------------------------------------------------
+-- Internal: CBOR V2 envelope codec
+-- ---------------------------------------------------------------------------
+
+decodeV2Envelope :: ByteString -> Either XPrvError V2Envelope
+decodeV2Envelope bs = do
+  (rest, envelope) <-
+    either (const $ Left XPrvDecodeError) Right $
+      CBOR.deserialiseFromBytes decodeEnvelope (BL.fromStrict bs)
+  if BL.null rest then pure envelope else Left XPrvDecodeError
```
lehins: Reduction in duplication
```suggestion
  case CBOR.deserialiseFromBytes decodeEnvelope (BL.fromStrict bs) of
    Right (rest, envelope) 
      | BL.null rest -> Right envelope
    _ -> Left XPrvDecodeError
```

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+      , encodeBytes cc
+      ]
+
+decodeAad :: AadContext -> Either XPrvError (PublicKey, ChainCode)
+decodeAad bs =
+  case CBOR.deserialiseFromBytes decodeAadFields (BL.fromStrict bs) of
+    Left _ -> Left XPrvDecodeError
+    Right (rest, result)
+      | BL.null rest -> Right result
+      | otherwise -> Left XPrvDecodeError
```
lehins: Redundant case
```suggestion
    Right (rest, result)
      | BL.null rest -> Right result
    _ -> Left XPrvDecodeError
```

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -0,0 +1,909 @@
...
+encryptedChangePass ::
+  (ByteArrayAccess oldPassPhrase, ByteArrayAccess newPassPhrase) =>
+  oldPassPhrase -> newPassPhrase -> EncryptedKey -> IO (Either XPrvError EncryptedKey)
+encryptedChangePass oldPass newPass ekey = do
+  emat <- decryptKeyMaterial ekey oldPass
+  case emat of
+    Left err -> pure (Left err)
+    Right mat ->
+      finally (wrapKeyMaterial newPass mat) (mlsbFinalize (kmSecretKey mat))
+
```
lehins: This patters is completely incorrect. It is absolutely possible for an async exception to happen in between call to `decryptKeyMaterial` and scope of `finally` is entered, which will lead to `kmSecretKey` not being finalized.

lehins (approved):
There is way too much stuff that needs to be fixed in this PR. 
@disassembler Fix whatever you can, let's merge it and I will fix the rest of the problems in it in a separate PR

PR #658, opened 2026-05-19
Add `Ord` constraint on `CertVRF` in `VRFAlgorithm`

neilmayhew:
# Description

We want to be able to store ledger predicate failures in a `Set`, which requires an `Ord` constraint. Some of these failures transitively embed VRF certs so they too need to have `Ord` instances.

This change adds `Ord` instances for the three types of VRF provided in base. It also upgrades the `Eq` constraint to `Ord` in the `VRFAlgorithm` class to avoid having to repeat more specific constraints in many places in the ledger codebase.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 2.5.0.0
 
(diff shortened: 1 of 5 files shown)


lehins (approved):
Very nice.
Thank you!

PR #660, opened 2026-05-20
Add Cardano.Base.Typeable with type TypeName a

aniketd:
# Description

Add `IsString` and `Show` instances for `TypeName a`, where `a` can be any type, so that it is convenient to use `show` on any type using the `TypeName` constructor, as well as just use a `String` when necessary.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-base/src/Cardano/Base/Typeable.hs
+++ b/cardano-base/src/Cardano/Base/Typeable.hs
@@ -1,0 +1,23 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+
+-- | Wrapper for Data.Typeable, plus convenience functions
+module Cardano.Base.Typeable (
+  module X,
+  TypeName (..),
+) where
+
+import Data.String (IsString (fromString))
+import Data.Typeable as X
+
+data TypeName a where
+  TypeNameString :: String -> TypeName a
+  TypeName :: Typeable a => TypeName a
+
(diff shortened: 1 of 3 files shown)


cardano-base/src/Cardano/Base/Typeable.hs
```diff
@@ -0,0 +1,22 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE GADTs #-}
+
+-- | Wrapper for Data.Typeable, plus convenience functions
```
aniketd, line 4: I have reused the module haddock from `Cardano.Base.Proxy` 😜 , I hope this is suitable.

cardano-base/src/Cardano/Base/Typeable.hs
```diff
@@ -0,0 +1,22 @@
...
+module Cardano.Base.Typeable (
+  module X,
+) where
+
+import Data.String (IsString (fromString))
+import Data.Typeable as X
+
+data TypeName a where
```
aniketd, line 13: This type needs to be exported!

cardano-base/src/Cardano/Base/Typeable.hs
lehins: Absolutely! 😄

lehins (approved):
Perfect. Thank you!

PR #662, opened 2026-05-25
Generalize type signature for withNumTests

erikd:
# Description

Code in `cardano-base` depends on the generalized type sig.

When this is merged, I will release it to CHaP.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-base/CHANGELOG.md
+++ b/cardano-base/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog for `cardano-base`
 
-## 0.1.4.1
+## 0.1.5.0
 
-*
+* Generalize type signature for `withNumTests` by making it polymorphic.
 
 ## 0.1.4.0
 
--- a/cardano-base/cardano-base.cabal
+++ b/cardano-base/cardano-base.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: cardano-base
-version: 0.1.4.0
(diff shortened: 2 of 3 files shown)


cardano-base/cardano-base.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: cardano-base
-version: 0.1.4.0
+version: 0.1.4.1
```
lehins: This is a backwards compatible, but semantically visible change. Therefore patch version is not applicable and minor version should be bumped instead,
```suggestion
version: 0.1.5.0
```

cardano-base/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 0.1.4.1
```
lehins:
```suggestion
## 0.1.5.0
```

cardano-base/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
...
 
-*
+* Generalize type signature for `withNumTests`.
```
lehins:
```suggestion
* Generalize type signature for `withNumTests` by making it polymorphic
```

lehins (changes requested):
Aside from version inconsistency it looks good

lehins (approved):
👍

PR #665, opened 2026-06-01
Add `FixedSizeCodec`, lift VRF sizes to the type level

Soupstraw:
# Description

This PR lifts the sizes of `VerKeyVRF`, `SignKeyVRF` and `CertVRF` to the type level, which is similar to what we already do for `DSIGN` and `KES`. I also added the `FixedSizeBytes` typeclass, which gives a uniform interface to all the `rawDeserialise*` and `encode*` functions for  `DSIGN`/`KES`/`VRF` types.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
+++ b/cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
@@ -1,0 +1,55 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Cardano.Binary.FixedSizeCodec (
+  FixedSizeCodec (..),
+  decodeFixedSized,
+  encodeFixedSized,
+  fixedSize,
+  guardFixedSized,
+) where
+
+import Cardano.Binary (Decoder, Encoding, decodeBytes, encodeBytes)
+import Control.Monad (when)
+import qualified Data.ByteString as BS
+import Data.Proxy (Proxy (..))
+import Data.Typeable (Typeable, typeRep)
+import GHC.TypeLits (KnownNat, Nat, natVal)
+
+class KnownNat (FixedSize a) => FixedSizeCodec a where
+  type FixedSize a :: Nat
+  rawEncodeFixedSized :: a -> BS.ByteString
+  rawDecodeFixedSized :: MonadFail m => BS.ByteString -> m a
+
+fixedSize :: forall a proxy. FixedSizeCodec a => proxy a -> Word
+fixedSize _ = fromInteger @Word . natVal $ Proxy @(FixedSize a)
+
+guardFixedSized ::
+  forall a m.
+  (FixedSizeCodec a, MonadFail m, Typeable a) =>
+  BS.ByteString -> m a -> m a
+guardFixedSized bs action = do
+  when (actualSize /= expectedSize) $
+    fail
+      ( tyName
+          ++ ": wrong length, expected "
+          ++ show expectedSize
+          ++ " bytes but got "
+          ++ show actualSize
+      )
+  action
+  where
+    tyName = show (typeRep $ Proxy @a)
+    expectedSize = fixedSize $ Proxy @a
+    actualSize = fromIntegral @Int @Word $ BS.length bs
+{-# INLINE guardFixedSized #-}
+
+decodeFixedSized :: FixedSizeCodec a => Decoder s a
+decodeFixedSized = decodeBytes >>= rawDecodeFixedSized
+
+encodeFixedSized :: FixedSizeCodec a => a -> Encoding
+encodeFixedSized = encodeBytes . rawEncodeFixedSized
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog for `cardano-crypto-class`
 
+## 2.6.0.0
+
+* Add `VerKeySizeVRF`, `SignKeySizeVRF` and `CertSizeVRF` type-level sizes to `VRFAlgorithm`
+* Change `sizeVerKeyVRF`, `sizeSignKeyVRF` and `sizeCertVRF` from `VRFAlgorithm` methods to standalone functions
+
 ## 2.5.1.0
 
 * Add `psbToByteArray`
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -1,9 +1,15 @@
 # Changelog for `cardano-crypto-class`
 
-## 2.6.0.0
+## 2.5.2.0
 
 * Add `VerKeySizeVRF`, `SignKeySizeVRF` and `CertSizeVRF` type-level sizes to `VRFAlgorithm`
 * Change `sizeVerKeyVRF`, `sizeSignKeyVRF` and `sizeCertVRF` from `VRFAlgorithm` methods to standalone functions
+* Add `FixedSizeCodec` as a superclass of `DSIGNAlgorithm`, `KESAlgorithm`, `VRFAlgorithm`, `UnsoundPureKESAlgorithm` and `DSIGNAggregatable`
+* Remove `failSizeCheck`
+* Deprecate the `encode*` and `decode*` functions for DSIGN, KES and VRF keys, signatures and certificates in favor of `encodeFixedSized` and `decodeFixedSized`
+* Add `FixedSizeCodec` instances for `PinnedSizedBytes` and `PackedBytes`
+* Add `psbFromByteStringForM`
+* Add `hashFromByteStringM`
 
 ## 2.5.1.0
 
--- a/cardano-crypto-class/CHANGELOG.md
+++ b/cardano-crypto-class/CHANGELOG.md
@@ -7,6 +7,9 @@
 * Add `FixedSizeCodec` as a superclass of `DSIGNAlgorithm`, `KESAlgorithm`, `VRFAlgorithm`, `UnsoundPureKESAlgorithm` and `DSIGNAggregatable`
 * Remove `failSizeCheck`
 * Deprecate the `encode*` and `decode*` functions for DSIGN, KES and VRF keys, signatures and certificates in favor of `encodeFixedSized` and `decodeFixedSized`
+* Deprecate the `rawSerialise*` and `rawDeserialise*` methods for DSIGN, KES and VRF keys, signatures, certificates and proofs of possession in favor of `rawEncodeFixedSized` and `rawDecodeFixedSized`
+* Deprecate `verKeySizeDSIGN`, `signKeySizeDSIGN`, `sigSizeDSIGN`, `verKeySizeKES`, `sigSizeKES`, `sizeVerKeyVRF`, `sizeSignKeyVRF` and `sizeCertVRF` in favor of `fixedSize`
+* Re-export `fixedSize` from `Cardano.Crypto.KES.Class`
 * Add `FixedSizeCodec` instances for `PinnedSizedBytes` and `PackedBytes`
 * Add `psbFromByteStringForM`
 * Add `hashFromByteStringM`
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -128,7 +128,6 @@ library
     cardano-base >=0.1.6,
     cardano-binary >=1.7.3,
     cardano-strict-containers,
-    cborg,
     crypton ^>=1.0,
     deepseq,
     heapwords,
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: cardano-crypto-class
-version: 2.6.0.0
+version: 2.5.2.0
 synopsis:
   Type classes abstracting over cryptography primitives for Cardano
 
@@ -126,7 +126,7 @@ library
     base16-bytestring >=1,
     bytestring,
     cardano-base >=0.1.6,
-    cardano-binary >=1.7.3,
+    cardano-binary >=1.9.1,
     cardano-strict-containers,
     crypton ^>=1.0,
     deepseq,
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: cardano-crypto-class
-version: 2.5.1.0
+version: 2.6.0.0
 synopsis:
   Type classes abstracting over cryptography primitives for Cardano
 
--- a/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
+++ b/cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
@@ -33,6 +33,11 @@ module Cardano.Crypto.DSIGN.BLS12381.Internal (
 #include "blst_util.h"
 
 import Cardano.Binary (FromCBOR (fromCBOR), ToCBOR (encodedSizeExpr, toCBOR))
+import Cardano.Binary.FixedSizeCodec (
+  FixedSizeCodec (..),
+  decodeFixedSized,
+  encodeFixedSized,
+ )
 import Cardano.Crypto.DSIGN.Class (
   DSIGNAggregatable (..),
   DSIGNAlgorithm (
@@ -40,37 +45,21 @@ import Cardano.Crypto.DSIGN.Class (
     KeyGenContextDSIGN,
     SeedSizeDSIGN,
     SigDSIGN,
-    SigSizeDSIGN,
     SignKeyDSIGN,
-    SignKeySizeDSIGN,
     Signable,
     VerKeyDSIGN,
-    VerKeySizeDSIGN,
     algorithmNameDSIGN,
     deriveVerKeyDSIGN,
     genKeyDSIGN,
     genKeyDSIGNWithContext,
-    rawDeserialiseSigDSIGN,
-    rawDeserialiseSignKeyDSIGN,
-    rawDeserialiseVerKeyDSIGN,
-    rawSerialiseSigDSIGN,
-    rawSerialiseSignKeyDSIGN,
-    rawSerialiseVerKeyDSIGN,
     signDSIGN,
     verifyDSIGN
   ),
-  decodePossessionProofDSIGN,
-  decodeSigDSIGN,
-  decodeSignKeyDSIGN,
-  decodeVerKeyDSIGN,
-  encodePossessionProofDSIGN,
-  encodeSigDSIGN,
-  encodeSignKeyDSIGN,
-  encodeVerKeyDSIGN,
   encodedPossessionProofDSIGNSizeExpr,
   encodedSigDSIGNSizeExpr,
   encodedSignKeyDSIGNSizeExpr,
   encodedVerKeyDSIGNSizeExpr,
+  rawSerialiseVerKeyDSIGN,
   seedSizeDSIGN,
  )
 import Cardano.Crypto.EllipticCurve.BLS12_381.Internal (
@@ -105,6 +94,7 @@ import Cardano.Crypto.PinnedSizedBytes (
 import Cardano.Crypto.Seed (getBytesFromSeedT)
 import Cardano.Crypto.Util (SignableRepresentation (getSignableRepresentation))
 import Control.DeepSeq (NFData)
+import Control.Monad (when)
 import Data.Bifunctor (first)
 import Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
@@ -119,6 +109,10 @@ import GHC.TypeLits (KnownSymbol, Symbol, symbolVal)
 import GHC.TypeNats (KnownNat, type (+))
 import NoThunks.Class (NoThunks)
 import System.IO.Unsafe (unsafeDupablePerformIO)
+
+failDecodeBLS :: MonadFail m => String -> String -> m a
+failDecodeBLS ty msg =
+  fail $ ty <> " BLS12381DSIGN: deserialisation failed (" <> msg <> ")"
 
 data BLS12381DSIGN curve
 
@@ -275,12 +269,9 @@ instance
   DSIGNAlgorithm (BLS12381DSIGN curve)
   where
   type SeedSizeDSIGN (BLS12381DSIGN curve) = CARDANO_BLST_SCALAR_SIZE
-  type SignKeySizeDSIGN (BLS12381DSIGN curve) = CARDANO_BLST_SCALAR_SIZE
 
   -- These *Sizes* are used in the serialization/deserialization
   -- so these use the compressed sizes of the BLS12-381 `Point curve`
-  type VerKeySizeDSIGN (BLS12381DSIGN curve) = CompressedPointSize curve
-  type SigSizeDSIGN (BLS12381DSIGN curve) = CompressedPointSize (DualCurve curve)
   type Signable (BLS12381DSIGN curve) = SignableRepresentation
 
   -- Context can hold domain separation tag and/or augmentation data for signatures
@@ -378,51 +369,6 @@ instance
                     infoPtr
                     (fromIntegral @Int @CSize infoLen)
 
-  -- Note that this also compresses the signature according to the ZCash standard
-  {-# INLINE rawSerialiseSigDSIGN #-}
-  rawSerialiseSigDSIGN (SigBLS12381 sigPSB) = blsCompress @(DualCurve curve) sigPSB
-
-  {-# INLINE rawSerialiseVerKeyDSIGN #-}
-  -- Note that this also compresses the verification key according to the ZCash standard
-  rawSerialiseVerKeyDSIGN (VerKeyBLS12381 vkPSB) = blsCompress @curve vkPSB
-
-  {-# INLINE rawSerialiseSignKeyDSIGN #-}
-  rawSerialiseSignKeyDSIGN (SignKeyBLS12381 skPSB) = scalarToBS skPSB
-
-  {-# INLINE rawDeserialiseVerKeyDSIGN #-}
-  rawDeserialiseVerKeyDSIGN bs =
-    -- Note that this also performs a group membership check.
-    -- That is, the deserialised point is in the subgroup of Curve1/Curve2.
-    case blsUncompress @curve bs of
-      Left _ -> Nothing
-      Right vkPsb ->
-        -- Reject the identity (point at infinity) as a verification key
-        if blsIsInf @curve vkPsb
-          then Nothing
-          else Just (VerKeyBLS12381 vkPsb)
-
-  {-# INLINE rawDeserialiseSignKeyDSIGN #-}
-  rawDeserialiseSignKeyDSIGN bs =
-    -- A signing key is strictly a BE integer mod the curve order.
-    -- The `DSIGN` interface via PSB would ensure at the type level that
-    -- they are of size 32 bytes (256 bits). But we must even ensure
-    -- they are valid Scalars, i.e., less than the curve order (255 bits).
-    case scalarFromBS bs of
-      Left _ -> Nothing
-      Right skScalar ->
-        -- Reject the zero scalar as a signing key
-        if BS.all (== 0) (scalarToBS skScalar)
-          then Nothing
-          else Just (SignKeyBLS12381 skScalar)
-
-  {-# INLINE rawDeserialiseSigDSIGN #-}
-  rawDeserialiseSigDSIGN bs =
-    -- Note that this also performs a group membership check.
-    -- That is, the deserialised point is in the subgroup of Curve1/Curve2.
-    case blsUncompress @(DualCurve curve) bs of
-      Left _ -> Nothing
-      Right sigPsb -> Just (SigBLS12381 sigPsb)
-
 deriving stock instance
   BLS curve =>
   Eq (VerKeyDSIGN (BLS12381DSIGN curve))
@@ -447,35 +393,89 @@ instance Show (SignKeyDSIGN (BLS12381DSIGN curve)) where
 
 instance
   BLS12381CurveConstraints curve =>
+  FixedSizeCodec (VerKeyDSIGN (BLS12381DSIGN curve))
+  where
+  type FixedSize (VerKeyDSIGN (BLS12381DSIGN curve)) = CompressedPointSize curve
+
+  -- Note that this also compresses the verification key according to the ZCash standard
+  {-# INLINE rawEncodeFixedSized #-}
+  rawEncodeFixedSized (VerKeyBLS12381 vkPSB) = blsCompress @curve vkPSB
+  {-# INLINE rawDecodeFixedSized #-}
+  rawDecodeFixedSized bs =
+    -- Note that this also performs a group membership check.
+    -- That is, the deserialised point is in the subgroup of Curve1/Curve2.
+    case blsUncompress @curve bs of
+      Left err -> failDecodeBLS "VerKeyDSIGN" $ show err
+      Right vkPsb ->
+        -- Reject the identity (point at infinity) as a verification key
+        if blsIsInf @curve vkPsb
+          then failDecodeBLS "VerKeyDSIGN" "infinity point"
+          else pure (VerKeyBLS12381 vkPsb)
+
+instance FixedSizeCodec (SignKeyDSIGN (BLS12381DSIGN curve)) where
+  type FixedSize (SignKeyDSIGN (BLS12381DSIGN curve)) = CARDANO_BLST_SCALAR_SIZE
+  {-# INLINE rawEncodeFixedSized #-}
+  rawEncodeFixedSized (SignKeyBLS12381 skPSB) = scalarToBS skPSB
+  {-# INLINE rawDecodeFixedSized #-}
+  rawDecodeFixedSized bs = do
+    -- A signing key is strictly a BE integer mod the curve order.
+    -- We must ensure they are valid Scalars, i.e., less than the curve order (255 bits).
+    case scalarFromBS bs of
+      Left err -> failDecodeBLS "SignKeyDSIGN" $ show err
+      Right skScalar ->
+        -- Reject the zero scalar as a signing key
+        if BS.all (== 0) (scalarToBS skScalar)
+          then failDecodeBLS "SignKeyDSIGN" "zero scalar"
+          else pure (SignKeyBLS12381 skScalar)
+
+instance
+  BLS12381CurveConstraints curve =>
+  FixedSizeCodec (SigDSIGN (BLS12381DSIGN curve))
+  where
+  type FixedSize (SigDSIGN (BLS12381DSIGN curve)) = CompressedPointSize (DualCurve curve)
+
+  -- Note that this also compresses the signature according to the ZCash standard
+  {-# INLINE rawEncodeFixedSized #-}
+  rawEncodeFixedSized (SigBLS12381 sigPSB) = blsCompress @(DualCurve curve) sigPSB
+  {-# INLINE rawDecodeFixedSized #-}
+  rawDecodeFixedSized bs =
+    -- Note that this also performs a group membership check.
+    -- That is, the deserialised point is in the subgroup of Curve1/Curve2.
+    case blsUncompress @(DualCurve curve) bs of
+      Left err -> failDecodeBLS "SigDSIGN" $ show err
+      Right sigPsb -> pure (SigBLS12381 sigPsb)
+
+instance
+  BLS12381CurveConstraints curve =>
   ToCBOR (VerKeyDSIGN (BLS12381DSIGN curve))
   where
-  toCBOR = encodeVerKeyDSIGN
+  toCBOR = encodeFixedSized
   encodedSizeExpr _ = encodedVerKeyDSIGNSizeExpr
 
 instance
   BLS12381CurveConstraints curve =>
   FromCBOR (VerKeyDSIGN (BLS12381DSIGN curve))
   where
-  fromCBOR = decodeVerKeyDSIGN
+  fromCBOR = decodeFixedSized
 
 instance
   BLS12381CurveConstraints curve =>
   ToCBOR (SignKeyDSIGN (BLS12381DSIGN curve))
   where
-  toCBOR = encodeSignKeyDSIGN
+  toCBOR = encodeFixedSized
   encodedSizeExpr _ = encodedSignKeyDSIGNSizeExpr
 
 instance
   BLS12381CurveConstraints curve =>
   FromCBOR (SignKeyDSIGN (BLS12381DSIGN curve))
   where
-  fromCBOR = decodeSignKeyDSIGN
+  fromCBOR = decodeFixedSized
 
 instance
   BLS12381CurveConstraints curve =>
   ToCBOR (SigDSIGN (BLS12381DSIGN curve))
   where
-  toCBOR = encodeSigDSIGN
+  toCBOR = encodeFixedSized
   encodedSizeExpr _ = encodedSigDSIGNSizeExpr
 
 -- | Helper functions to extract the internal Point representation
@@ -490,7 +490,7 @@ instance
   BLS12381CurveConstraints curve =>
   FromCBOR (SigDSIGN (BLS12381DSIGN curve))
   where
-  fromCBOR = decodeSigDSIGN
+  fromCBOR = decodeFixedSized
 
 instance
   BLS12381CurveConstraints curve =>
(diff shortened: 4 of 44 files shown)


cardano-crypto-class/src/Cardano/Crypto/VRF/Class.hs
```diff
@@ -162,8 +169,14 @@ class
...
 
   sizeVerKeyVRF :: proxy v -> Word
+  sizeVerKeyVRF _ = fromInteger @Word . natVal $ Proxy @(VerKeySizeVRF v)
+
   sizeSignKeyVRF :: proxy v -> Word
+  sizeSignKeyVRF _ = fromInteger @Word . natVal $ Proxy @(SignKeySizeVRF v)
+
   sizeCertVRF :: proxy v -> Word
+  sizeCertVRF _ = fromInteger @Word . natVal $ Proxy @(CertSizeVRF v)
+
```
neilmayhew: I don't think these need to be methods any more, but could be top-level functions instead. Then, you wouldn't need a test to ensure that the default definitions haven't been overridden.

This would be a breaking change, because it changes the way the functions need to be imported in client code that uses explicit imports, but it would be a small breakage.

The only place it's used in Ledger is `libs/cardano-protocol-tpraos/testlib/Test/Cardano/Protocol/Crypto/VRF/Fake.hs`

However, maybe it's better to retain the backward compatibility for now.

cardano-crypto-class/src/Cardano/Crypto/VRF/Class.hs
```diff
@@ -100,6 +103,10 @@ class
   data SignKeyVRF v :: Type
   data CertVRF v :: Type
 
+  type VerKeySizeVRF v :: Nat
+  type SignKeySizeVRF v :: Nat
+  type CertSizeVRF v :: Nat
```
neilmayhew, line 120: Just wondering: why didn't you make the names match the methods you're replacing? (`SizeVerKeyVRF` etc.)

If you prefer not to have `Size` at the front, I think it would still be better to have it at the back instead of splitting the `VRF` part away from the rest.

cardano-crypto-praos/testlib/Test/Crypto/VRF.hs
```diff
@@ -265,18 +265,23 @@ testVRFAlgorithm _ n =
...
+              rawSerialiseSignKeyVRF
+              (sizeSignKeyVRF (Proxy @v))
+          prop "Cert" $
+            prop_size_serialise @(CertVRF v)
+              rawSerialiseCertVRF
+              (sizeCertVRF (Proxy @v))
+        describe "type-level sizes match methods" $ do
+          it "VerKey" $ natVal (Proxy @(VerKeySizeVRF v)) @?= toInteger (sizeVerKeyVRF $ Proxy @v)
+          it "SignKey" $ natVal (Proxy @(SignKeySizeVRF v)) @?= toInteger (sizeSignKeyVRF $ Proxy @v)
+          it "Cert" $ natVal (Proxy @(CertSizeVRF v)) @?= toInteger (sizeCertVRF $ Proxy @v)
```
neilmayhew: This test would be unnecessary if `sizeVerKeyVRF` etc. were functions rather than methods.

cardano-crypto-praos/CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 # Changelog for `cardano-crypto-praos`
 
-## 2.2.3.1
+## 2.2.4.0
 
-*
+* Add `VerKeySizeVRF`, `SignKeySizeVRF` and `CertSizeVRF` instances
```
neilmayhew:
```suggestion
* Add `VerKeySizeVRF`, `SignKeySizeVRF` and `CertSizeVRF` types to the `VRFAlgorithm` instances for
  `PraosVRF`, `PraosBatchCompatVRF`, `SimpleVRF`, `MockVRF`, `NeverVRF`
```

neilmayhew (approved):
LGTM. My comments are just suggestions.

cardano-crypto-class/src/Cardano/Crypto/FixedSizeBytes.hs
```diff
@@ -0,0 +1,15 @@
...
+
+import qualified Cardano.Binary as C
+import qualified Data.ByteString as BS
+import Data.Typeable (Typeable)
+import GHC.TypeLits (KnownNat, Nat)
+
+class (Typeable a, KnownNat (FixedSize a)) => FixedSizeBytes a where
+  type FixedSize a :: Nat
+  encodeFixed :: a -> C.Encoding
+  rawDeserialise :: BS.ByteString -> Maybe a
```
neilmayhew: It's odd to be mixing (en)code and (de)serialise. Could we have encode/decode or serialise/deserialise?

BTW I prefer encode/decode because of potential confusion over International vs American spelling of serialise/deserialise.

cardano-crypto-class/src/Cardano/Crypto/VRF/Class.hs
Soupstraw: I just copied how it's done in `DSIGN` and `KES`. In fact it used to be like you suggested until https://github.com/IntersectMBO/cardano-base/pull/610.

cardano-crypto-class/src/Cardano/Crypto/FixedSizeBytes.hs
Soupstraw: renamed `rawDeserialize` to `rawDecodeFixed`

cardano-crypto-class/src/Cardano/Crypto/VRF/Class.hs
neilmayhew: Fair enough. The consistency is the most important thing, and I vaguely remember seeing that previous PR.

lehins (changes requested):
I am blocking this PR because I want to give my review before it gets merged

cardano-crypto-class/src/Cardano/Crypto/FixedSizeBytes.hs
```diff
@@ -0,0 +1,15 @@
...
+
+import qualified Cardano.Binary as C
+import qualified Data.ByteString as BS
+import Data.Typeable (Typeable)
+import GHC.TypeLits (KnownNat, Nat)
+
+class (Typeable a, KnownNat (FixedSize a)) => FixedSizeBytes a where
+  type FixedSize a :: Nat
+  encodeFixed :: a -> C.Encoding
+  rawDecodeFixed :: BS.ByteString -> Maybe a
```
lehins: Three things that need to change in this interface:
* it should not be coupled to CBOR (i.e. `Encoding` should not be used here).
* naming needs to be made consistent
* decoder should return `MonadFail`, instead of `Maybe`
```suggestion
  rawEncodeFixedSized :: a -> BS.ByteString
  rawDecodeFixedSized :: MoandFail m => BS.ByteString -> m a
```

Strategy is to:

* Add `FixedSizeCodec` superclass constraint to all classes (eg. `FixedSizeCodec (VerKeyDSIGN algo), ... => `
* move this function next to the type class and use it for instance declarations: https://github.com/IntersectMBO/cardano-base/blob/52b93c60f8394be8bb75018c444b5fa311b29498/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs#L290-L305
* Add CBOR encoder/decoder that use this `FixedSizeCodec`, eg.
  ```haskell
  decodeFixedSized = decodeBytes >>= rawDecodeFixedSizedBytes
  ```
* Use above helper `decodeFixedSized` in actual decoder, that can also be deprecated, eg: `decodeSigDSIGN = decodeFixedSized` instead of https://github.com/IntersectMBO/cardano-base/blob/52b93c60f8394be8bb75018c444b5fa311b29498/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs#L283-L288

cardano-crypto-class/src/Cardano/Crypto/FixedSizeBytes.hs
```diff
@@ -0,0 +1,15 @@
...
+{-# LANGUAGE TypeFamilies #-}
+
+module Cardano.Crypto.FixedSizeBytes (FixedSizeBytes (..)) where
+
+import qualified Cardano.Binary as C
+import qualified Data.ByteString as BS
+import Data.Typeable (Typeable)
+import GHC.TypeLits (KnownNat, Nat)
+
+class (Typeable a, KnownNat (FixedSize a)) => FixedSizeBytes a where
```
lehins: Not a very descriptive name of the interface. It is not really about the bytes, but encoding types with fixed size to and from bytes
```suggestion
class (Typeable a, KnownNat (FixedSize a)) => FixedSizeCodec a where
```

cardano-crypto-class/src/Cardano/Crypto/FixedSizeBytes.hs
```diff
@@ -0,0 +1,15 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module Cardano.Crypto.FixedSizeBytes (FixedSizeBytes (..)) where
```
lehins: This is a general interface and does not belong in `cardano-crypto-class`
Let's move it into `cardano-binary` into `Cardano.Binary.FixedSizeCodec`

lehins (changes requested):
This is a very nice direction! Still needs a bit of love though to make ideal.

cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
```diff
@@ -0,0 +1,63 @@
...
+import qualified Data.ByteString as BS
+import Data.Proxy (Proxy (..))
+import Data.Typeable (Typeable, tyConName, typeRep, typeRepTyCon)
+import GHC.TypeLits (KnownNat, Nat, natVal)
+
+class (Typeable a, KnownNat (FixedSize a)) => FixedSizeCodec a where
+  type FixedSize a :: Nat
+  rawEncodeFixedSized :: a -> BS.ByteString
+  rawDecodeFixedSized :: MonadFail m => BS.ByteString -> m a
+
```
lehins, line 27: I think it will be nice to define a top level helper like this


```suggestion
fixedSize :: FixedSizeCodec a => proxy a -> Int
fixedSize = fromInteger @Word . natVal
```

cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
```diff
@@ -0,0 +1,63 @@
...
+            ++ show expectedSize
+            ++ " bytes but got "
+            ++ show actualSize
+        )
+  | otherwise = fail $ tyName ++ ": cannot decode " ++ name
+  where
+    tyName = tyConName (typeRepTyCon (typeRep (Proxy @a)))
+    expectedSize = fromInteger @Word (natVal (Proxy @(FixedSize a)))
+    actualSize = fromIntegral @Int @Word (BS.length bs)
+{-# NOINLINE failFixedSized #-}
```
lehins: This will become just a helper that ensures the size is correct.
```suggestion
guardFixedSized ::
  forall a m.
  (Typeable a, FixedSizeCodec a, MonadFail m) =>
  String -> Proxy a -> BS.ByteString -> m ()
guardFixedSized name proxy bs =
  when (BS.length bs /= fixedSize proxy) $ 
    fail
      ( tyName
          ++ ": wrong length, expected "
          ++ show expectedSize
          ++ " bytes but got "
          ++ show actualSize
      )
  where
    tyName = show (typeRep proxy)
{-# INLINE guardFixedSized #-}
```

cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
```diff
@@ -0,0 +1,63 @@
...
+-- | Build a raw fixed-size decoder, failing with an informative message when
+-- the underlying decoder returns 'Nothing'. The first argument should be a
+-- noun describing the type (e.g. "key", "signature")
+mkRawFixedSizedDecoder ::
+  forall a m.
+  (Typeable a, KnownNat (FixedSize a), MonadFail m) =>
+  String -> (BS.ByteString -> Maybe a) -> BS.ByteString -> m a
+mkRawFixedSizedDecoder name rawDecode bs =
+  maybe (failFixedSized name (Proxy @a) bs) pure (rawDecode bs)
+{-# INLINE mkRawFixedSizedDecoder #-}
```
lehins: I think you missed the point of my suggestion about `MonadFail`

The goal is to:
* avoid duplicate logic of checking the size
* improve failure messages
* and later deprecation or extraction into standalone functions of individual `rawDeserialise*` functions. 

Which means we need an instance of `FixedSizeCodec` for each algorithm, eg. `instance FixedSizeCodec (VerKeyDSIGN Ed25519DSIGN)` and all of the `raw(Des|S)erialise*` can get this implementation `rawDeserialise* = rawDecodeFixedSized`

```suggestion
```

lehins (changes requested):
We need to reduce duplication if we introducing this new interface, which requires inverting the approach of `rawEncodeFixedSized`/`rawDecodeFixedSized` using `rawSerialise*`/`rawDeserialise*`

cardano-base/src/Cardano/Base/Bytes.hs
```diff
@@ -40,11 +39,11 @@ slice offset size =
...
 splitsAt :: [Int] -> ByteString -> [ByteString]
-splitsAt = go 0
-  where
-    go !_ [] bs
-      | BS.null bs = []
-      | otherwise = [bs]
-    go !off (sz : szs) bs
-      | BS.length bs >= sz = BS.take sz bs : go (off + sz) szs (BS.drop sz bs)
-      | otherwise = []
+splitsAt [] bs
```
lehins: If you gonna rewrite this function, then you should make it tail recursive.
However, it is best to avoid introducing changes to logic in large refactoring PRs like this and make such changes in standalone PRs where it is much more clear where potentially dangerous can be introduced in.

cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
```diff
@@ -0,0 +1,58 @@
...
+) where
+
+import Cardano.Binary (Decoder, Encoding, decodeBytes, encodeBytes)
+import Control.Monad (when)
+import qualified Data.ByteString as BS
+import Data.Proxy (Proxy (..))
+import Data.Typeable (Typeable, typeRep)
+import GHC.TypeLits (KnownNat, Nat, natVal)
+
+type family FixedSize a :: Nat
```
lehins: Why did you move it outside of the type class?

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
```diff
@@ -568,13 +565,27 @@ deriving stock instance
...
+  FixedSizeCodec (PossessionProofDSIGN (BLS12381DSIGN curve))
+  where
+  type
+    FixedSize (PossessionProofDSIGN (BLS12381DSIGN curve)) =
+      PossessionProofSizeDSIGN (BLS12381DSIGN curve)
+  rawEncodeFixedSized = rawSerialisePossessionProofDSIGN
+  rawDecodeFixedSized bs = do
+    case rawDeserialisePossessionProofDSIGN bs of
+      Nothing -> fail "PossessionProofDSIGN BLS12381DSIGN: deserialisation failed"
+      Just p -> pure p
```
lehins: Instead make `rawDeserialisePossessionProofDSIGN` use `rawDecodeFixedSized`
```suggestion
  rawDecodeFixedSized bs = do
    -- Note that these also perform group membership and size checks.
    -- It will also ensure that all of the supplied `ByteString` is consumed
    -- through the size checks.
    let failMsg msg = "PossessionProofDSIGN BLS12381DSIGN: deserialisation failed (" <> msg " <> ")"
    case blsUncompress @(DualCurve curve) bs of
      Left err -> fail $ failMsg (show err)
      Right mu1Point -> do 
        -- Reject the zero point (point at infinity) for both mu1 and mu2
        when (blsIsInf @(DualCurve curve) mu1Point) $ do
          fail $ failMsg "infinity point"
        pure $ PossessionProofBLS12381 mu1Point
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/BLS12381/Internal.hs
```diff
@@ -445,37 +387,92 @@ instance Eq (SignKeyDSIGN (BLS12381DSIGN curve)) where
...
+  type FixedSize (VerKeyDSIGN (BLS12381DSIGN curve)) = CompressedPointSize curve
+  -- Note that this also compresses the verification key according to the ZCash standard
+  {-# INLINE rawEncodeFixedSized #-}
+  rawEncodeFixedSized (VerKeyBLS12381 vkPSB) = blsCompress @curve vkPSB
+  {-# INLINE rawDecodeFixedSized #-}
+  rawDecodeFixedSized bs =
+    -- Note that this also performs a group membership check.
+    -- That is, the deserialised point is in the subgroup of Curve1/Curve2.
+    case blsUncompress @curve bs of
+      Left err -> fail $ "VerKeyDSIGN BLS12381DSIGN: deserialisation failed (" <> show err <> ")"
```
lehins: Might be useful adding a helper in this module:
```
failDecodeBLS :: MoandFail m => String -> String -> m ()
failDecodeBLS ty msg =
  fail $ ty <> " BLS12381DSIGN: deserialisation failed (" <> msg <> ")"
```

cardano-crypto-class/src/Cardano/Crypto/Hash/Class.hs
```diff
@@ -210,16 +210,18 @@ hashToBytes (UnsafeHashRep h) = unpackPinnedBytes h
...
-  HashAlgorithm h =>
+  forall h a m.
+  ( HashAlgorithm h
+  , MonadFail m
+  ) =>
   -- | It must have an exact length, as given by 'hashSize'.
   ByteString ->
-  Maybe (Hash h a)
+  m (Hash h a)
 hashFromBytes bytes
```
lehins: Please add a new function with this implementation:
```suggestion
hashFromByteStringM = fmap UnsafeHashRep . packByteString
```

and keep the type signature for original `hashFromBytes` function intact, but with this new implementation:
```haskell
hashFromBytes = hashFromByteStringM
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -185,53 +174,62 @@ instance DSIGNAlgorithm SchnorrSecp256k1DSIGN where
...
-            _ -> Nothing
-  rawDeserialiseSignKeyDSIGN bs =
-    SignKeySchnorrSecp256k1 <$> psbFromByteStringCheck bs
-  rawDeserialiseSigDSIGN bs =
-    SigSchnorrSecp256k1 <$> psbFromByteStringCheck bs
+  {-# NOINLINE rawDecodeFixedSized #-}
+  rawDecodeFixedSized bs = do
+    result <- pure . unsafeDupablePerformIO . unsafeUseAsCStringLen bs $ \(ptr, len) -> do
+      unless (len == fixedSize (Proxy @(VerKeyDSIGN SchnorrSecp256k1DSIGN))) $
+        fail "VerKeyDSIGN SchnorrSecp256k1DSIGN: invalid length"
```
lehins: This is very wrong. Here you switched to using `fail` in `IO`, which becomes a run time exception
```suggestion
  rawDecodeFixedSized bs = do
    guardFixedSized (Proxy @(VerKeyDSIGN SchnorrSecp256k1DSIGN)) bs
    result <- pure . unsafeDupablePerformIO . unsafeUseAsCStringLen bs $ \(ptr, len) -> do
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -185,53 +174,62 @@ instance DSIGNAlgorithm SchnorrSecp256k1DSIGN where
...
+        1 -> pure . VerKeySchnorrSecp256k1 $ vkPsb
+        _ -> Nothing
+    case result of
+      Nothing -> fail "VerKeyDSIGN: cannot decode key"
+      Just vk -> pure vk
+
+instance FixedSizeCodec (SignKeyDSIGN SchnorrSecp256k1DSIGN) where
+  type FixedSize (SignKeyDSIGN SchnorrSecp256k1DSIGN) = SECP256K1_SCHNORR_PRIVKEY_BYTES
+  rawEncodeFixedSized (SignKeySchnorrSecp256k1 skPSB) = psbToByteString skPSB
+  rawDecodeFixedSized = fixedSizedFromByteStringM SignKeySchnorrSecp256k1
```
lehins: Please get rid of `fixedSizedFromByteStringM`
```suggestion
  rawEncodeFixedSized (SignKeySchnorrSecp256k1 skPSB) = psbToByteString skPSB
  rawDecodeFixedSized = SignKeySchnorrSecp256k1 <$> rawDecodeFixedSized
```

lehins (changes requested):

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
Soupstraw: With your suggestion the error becomes:
```
  7) cardano-crypto-class, Secp256k1 Test Vectors, SchnorrSecp256k1, Parsing should fail when using invalid length signatures.
       Expected invalid length signature error..
       expected: "wrong length, expected 64 bytes but got 63"
        but got: "PinnedSizedBytes 64: wrong length, expected 64 bytes but got 63"
```

lehins: Why is that a problem? Just fix the test.
I am not too worried about actual types being included in the failure message

cardano-crypto-class/src/Cardano/Crypto/DSIGN/EcdsaSecp256k1.hs
```diff
@@ -129,14 +120,21 @@ newtype MessageHash = MH (PinnedSizedBytes SECP256K1_ECDSA_MESSAGE_BYTES)
...
+  type FixedSize MessageHash = SECP256K1_ECDSA_MESSAGE_BYTES
+  rawEncodeFixedSized (MH psb) = psbToByteString psb
+  rawDecodeFixedSized bs = MH <$> rawDecodeFixedSized bs
+
 -- | Take a blob of bytes (which is presumed to be a 32-byte hash), verify its
 -- length, and package it into a 'MessageHash' if that length is exactly 32.
 toMessageHash :: ByteString -> Maybe MessageHash
-toMessageHash bs = MH <$> psbFromByteStringCheck bs
+toMessageHash = rawDecodeFixedSized
+{-# DEPRECATED toMessageHash "Use rawDecodeFixedSized instead" #-}
```
lehins: Could you please go through deprecations and make sure suggested functions have links included for replacements, which is done this way:
```suggestion
{-# DEPRECATED toMessageHash "Use `rawDecodeFixedSized` instead" #-}
```

cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
```diff
@@ -0,0 +1,54 @@
...
+      ( tyName
+          ++ ": wrong length, expected "
+          ++ show expectedSize
+          ++ " bytes but got "
+          ++ show actualSize
+      )
+  where
+    tyName = show (typeRep proxy)
+    expectedSize = fromIntegral $ fixedSize proxy
+    actualSize = BS.length bs
```
lehins: Any concern of overflow is removed with this approach. Not that it was a real concern, but I think it is a tiny bit better to cast an `Int` that is guaranteed to be non-negative (due to `length`) to `Word`, instead of `Word` to `Int` 
```suggestion
    expectedSize = fixedSize proxy
    actualSize = fromIntegral @Int @Word $ BS.length bs
```

cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
```diff
@@ -0,0 +1,54 @@
...
+) where
+
+import Cardano.Binary (Decoder, Encoding, decodeBytes, encodeBytes)
+import Control.Monad (when)
+import qualified Data.ByteString as BS
+import Data.Proxy (Proxy (..))
+import Data.Typeable (Typeable, typeRep)
+import GHC.TypeLits (KnownNat, Nat, natVal)
+
+class (Typeable a, KnownNat (FixedSize a)) => FixedSizeCodec a where
```
lehins: Please remove this `Typeable` constraint, since it is not needed
```suggestion
class KnownNat (FixedSize a) => FixedSizeCodec a where
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs
```diff
@@ -332,44 +348,40 @@ verifySignedDSIGN ::
...
 
-encodeSignedDSIGN :: DSIGNAlgorithm v => SignedDSIGN v a -> Encoding
-encodeSignedDSIGN (SignedDSIGN s) = encodeSigDSIGN s
+encodeSignedDSIGN :: (DSIGNAlgorithm v, Typeable a) => SignedDSIGN v a -> Encoding
+encodeSignedDSIGN = encodeFixedSized
+{-# DEPRECATED encodeSignedDSIGN "Use `encodeFixedSized` instead" #-}
 
-decodeSignedDSIGN :: DSIGNAlgorithm v => Decoder s (SignedDSIGN v a)
-decodeSignedDSIGN = SignedDSIGN <$> decodeSigDSIGN
+decodeSignedDSIGN :: (DSIGNAlgorithm v, Typeable a) => Decoder s (SignedDSIGN v a)
```
lehins: You need to revert these `Typeable` constraint, otherwise it would be concidered a breaking change. Also it is not needed.
```suggestion
decodeSignedDSIGN :: DSIGNAlgorithm v => Decoder s (SignedDSIGN v a)
```

cardano-crypto-class/src/Cardano/Crypto/DSIGN/SchnorrSecp256k1.hs
```diff
@@ -185,53 +174,68 @@ instance DSIGNAlgorithm SchnorrSecp256k1DSIGN where
...
+        fromIntegral @Word @Int . fixedSize $ Proxy @(VerKeyDSIGN SchnorrSecp256k1DSIGN)
+    unsafeDupablePerformIO . unsafeUseAsCStringLen bs $ \case
+      (ptr, len)
+        | len /= expectedSize ->
+            pure . fail $
+              "VerKeyDSIGN SchnorrSecp256k1DSIGN: wrong length, expected "
+                <> show expectedSize
+                <> " bytes but got "
+                <> show len
+        | otherwise -> do
```
lehins: Why not use the existing helper?
```suggestion
    guardFixedSized (Proxy @(VerKeyDSIGN SchnorrSecp256k1DSIGN)) bs
    unsafeDupablePerformIO . unsafeUseAsCString bs $ \ptr -> do
```

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -226,37 +237,30 @@ psbFromBytes ws0 = PSB (pinnedByteArrayFromListN size ws)
...
-    size = fromInteger (natVal (Proxy :: Proxy n))
+  forall m a.
+  (MonadFail m, FixedSizeCodec a) =>
+  Proxy a -> BS.ByteString -> m (PinnedSizedBytes (FixedSize a))
+psbFromByteStringM proxy bs = do
+  let size = fromInteger (natVal (Proxy @(FixedSize a)))
+  guardFixedSized proxy bs
+  unsafeDupablePerformIO $
+    BS.useAsCStringLen bs $ \(Ptr addr#, _) -> do
+      marr@(MutableByteArray marr#) <- newPinnedByteArray size
```
lehins: You do not need an extra binding, since `guardFixedSize` will ensure they are equal
```suggestion
psbFromByteStringM proxy bs = do
  guardFixedSized proxy bs
  unsafeDupablePerformIO $
    BS.useAsCStringLen bs $ \(Ptr addr#, size) -> do
      marr@(MutableByteArray marr#) <- newPinnedByteArray size
```

cardano-binary/src/Cardano/Binary/FixedSizeCodec.hs
```diff
@@ -0,0 +1,54 @@
...
+  Proxy a -> BS.ByteString -> m ()
+guardFixedSized proxy bs =
+  when (actualSize /= expectedSize) $
+    fail
+      ( tyName
+          ++ ": wrong length, expected "
+          ++ show expectedSize
+          ++ " bytes but got "
+          ++ show actualSize
+      )
```
lehins, line 43: While doing a final review, I realized that using `guardFixedSized` as it is defined here is a bit error prone, since there is no guarantee that supplied `Proxy` will have the same type as the algorithm that it is being applied to.

In other words instead of:
```haskell
  rawDecodeFixedSized bs = do
    guardFixedSized (Proxy @(SignKeyDSIGN Ed25519DSIGN)) bs
    pure . genKeyDSIGN . mkSeedFromBytes $ bs
```
it is easy to make a very-hard-to-spot mistake like this:
```haskell
  rawDecodeFixedSized bs = do
    guardFixedSized (Proxy @(VerKeyDSIGN Ed25519DSIGN)) bs
    pure . genKeyDSIGN . mkSeedFromBytes $ bs
```
So, I think it will be much better if we implement it like this:
```suggestion
guardFixedSized ::
  forall a m.
  (FixedSizeCodec a, MonadFail m) =>
  BS.ByteString -> m a -> m a
guardFixedSized bs action =
  when (actualSize /= expectedSize) $
    fail
      ( tyName
          ++ ": wrong length, expected "
          ++ show expectedSize
          ++ " bytes but got "
          ++ show actualSize
      )
  action
```

With this change above example becomes:
```haskell
  rawDecodeFixedSized bs =
    guardFixedSized bs $ do
      pure . genKeyDSIGN $ mkSeedFromBytes bs
```

lehins (changes requested):

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -180,7 +185,14 @@ instance KnownNat n => IsString (Q (TExp (PinnedSizedBytes n))) where
     let n = fromInteger $ natVal (Proxy :: Proxy n)
     case decodeHexString hexStr n of
       Left err -> fail $ "<PinnedSizedBytes>: " ++ err
-      Right _ -> examineSplice [||either error psbFromByteString (decodeHexString hexStr n)||]
+      Right _ ->
+        examineSplice
+          [||
+          either
+            error
+            (fromMaybe (error "<PinnedSizedBytes>: size mismatch") . rawDecodeFixedSized)
```
lehins: It will be better to use this helper [`errorFail`](https://hackage.haskell.org/package/FailT-0.1.2.0/docs/Control-Monad-Trans-Fail-String.html#v:errorFail), so the user will receive the original failure message
```suggestion
            (errorFail . rawDecodeFixedSized)
```

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -226,37 +238,32 @@ psbFromBytes ws0 = PSB (pinnedByteArrayFromListN size ws)
...
-psbFromByteString :: KnownNat n => BS.ByteString -> PinnedSizedBytes n
-psbFromByteString bs =
-  case psbFromByteStringCheck bs of
+psbFromByteString ::
+  FixedSizeCodec a =>
+  Proxy a -> BS.ByteString -> PinnedSizedBytes (FixedSize a)
+psbFromByteString proxy bs =
+  case psbFromByteStringCheck proxy bs of
     Nothing -> error $ "psbFromByteString: Size mismatch, got: " ++ show (BS.length bs)
     Just psb -> psb
```
lehins: Same here
```suggestion
psbFromByteString = errorFail . psbFromByteStringM
```

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -226,37 +238,32 @@ psbFromBytes ws0 = PSB (pinnedByteArrayFromListN size ws)
...
-    size :: Int
-    size = fromInteger (natVal (Proxy :: Proxy n))
+  forall a m.
+  (MonadFail m, FixedSizeCodec a) =>
+  Proxy a -> BS.ByteString -> m (PinnedSizedBytes (FixedSize a))
+psbFromByteStringM _ bs =
+  guardFixedSized bs
+    . unsafeDupablePerformIO
+    . BS.useAsCStringLen bs
+    $ \(Ptr addr#, size) -> do
```
lehins: Man, this is so unreadable when all of these unsafe functions are composed like this. 
Also let's change to `unafeUseAsCStringLen`, since supplied bytestring is used for read only purpose 
```suggestion
  guardFixedSized bs $ do
    unsafeDupablePerformIO . BS.unafeUseAsCStringLen bs $ \(Ptr addr#, size) -> do
```

cardano-crypto-class/CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
...
+* Add `VerKeySizeVRF`, `SignKeySizeVRF` and `CertSizeVRF` type-level sizes to `VRFAlgorithm`
+* Change `sizeVerKeyVRF`, `sizeSignKeyVRF` and `sizeCertVRF` from `VRFAlgorithm` methods to standalone functions
+* Add `FixedSizeCodec` as a superclass of `DSIGNAlgorithm`, `KESAlgorithm`, `VRFAlgorithm`, `UnsoundPureKESAlgorithm` and `DSIGNAggregatable`
+* Remove `failSizeCheck`
+* Deprecate the `encode*` and `decode*` functions for DSIGN, KES and VRF keys, signatures and certificates in favor of `encodeFixedSized` and `decodeFixedSized`
+* Deprecate the `rawSerialise*` and `rawDeserialise*` methods for DSIGN, KES and VRF keys, signatures, certificates and proofs of possession in favor of `rawEncodeFixedSized` and `rawDecodeFixedSized`
+* Deprecate `verKeySizeDSIGN`, `signKeySizeDSIGN`, `sigSizeDSIGN`, `verKeySizeKES`, `sigSizeKES`, `sizeVerKeyVRF`, `sizeSignKeyVRF` and `sizeCertVRF` in favor of `fixedSize`
+* Re-export `fixedSize` from `Cardano.Crypto.KES.Class`
+* Add `FixedSizeCodec` instances for `PinnedSizedBytes` and `PackedBytes`
+* Change `psbFromByteString`, `psbFromByteStringCheck` and `psbFromByteStringM` to take a `Proxy` argument
```
lehins: Instead of introducing a breaking change like this it will be better to add new functions instead that can do this
```suggestion
* Add `psbFromByteStringForM`
```

cardano-crypto-class/CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Changelog for `cardano-crypto-class`
 
+## 2.6.0.0
```
lehins:
```suggestion
## 2.5.2.0
```

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -226,37 +238,32 @@ psbFromBytes ws0 = PSB (pinnedByteArrayFromListN size ws)
 
 -- | Convert a ByteString into PinnedSizedBytes. Input should contain the exact
 -- number of bytes as expected by type level @n@ size, otherwise error.
-psbFromByteString :: KnownNat n => BS.ByteString -> PinnedSizedBytes n
-psbFromByteString bs =
-  case psbFromByteStringCheck bs of
+psbFromByteString ::
+  FixedSizeCodec a =>
+  Proxy a -> BS.ByteString -> PinnedSizedBytes (FixedSize a)
```
lehins: Please revert change to the type signature and add HasCallStack
```suggestion
psbFromByteString :: (HasCallStack, KnownNat n) => BS.ByteString -> PinnedSizedBytes n
```

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -226,37 +238,32 @@ psbFromBytes ws0 = PSB (pinnedByteArrayFromListN size ws)
...
+psbFromByteString proxy bs =
+  case psbFromByteStringCheck proxy bs of
     Nothing -> error $ "psbFromByteString: Size mismatch, got: " ++ show (BS.length bs)
     Just psb -> psb
 
-psbFromByteStringCheck :: forall n. KnownNat n => BS.ByteString -> Maybe (PinnedSizedBytes n)
+psbFromByteStringCheck ::
+  FixedSizeCodec a =>
+  Proxy a -> BS.ByteString -> Maybe (PinnedSizedBytes (FixedSize a))
```
lehins: Ditto
```suggestion
psbFromByteStringCheck :: KnownNat n => BS.ByteString -> Maybe (PinnedSizedBytes n)
```

lehins (changes requested):
This PR has changed in unexpected ways. Please revert the changes.

This PR has no business changing like `cardano-crypto-wallet`

cardano-crypto-class/src/Cardano/Crypto/PinnedSizedBytes.hs
```diff
@@ -226,37 +243,37 @@ psbFromBytes ws0 = PSB (pinnedByteArrayFromListN size ws)
...
-
-psbFromByteStringCheck :: forall n. KnownNat n => BS.ByteString -> Maybe (PinnedSizedBytes n)
+psbFromByteString :: (HasCallStack, KnownNat n) => BS.ByteString -> PinnedSizedBytes n
+psbFromByteString = errorFail @String . psbFromByteStringM
+
+psbFromByteStringFor ::
+  forall a.
+  KnownNat (FixedSize a) =>
+  Proxy a -> BS.ByteString -> PinnedSizedBytes (FixedSize a)
+psbFromByteStringFor _ = psbFromByteString @(FixedSize a)
```
lehins: Please get rid of this one. `psbFromByteString` shouldn't exist either, but we'll deal with it later
```suggestion
```

cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: cardano-crypto-class
-version: 2.5.1.0
+version: 2.6.0.0
```
lehins:
```suggestion
version: 2.5.2.0
```

cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -40,7 +40,7 @@ library
     bytestring,
     cardano-base >=0.1.6,
     cardano-binary,
-    cardano-crypto-class,
+    cardano-crypto-class >=2.6,
```
lehins:
```suggestion
    cardano-crypto-class >=2.5.2,
```

cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -59,7 +59,7 @@ library testlib
   build-depends:
     QuickCheck,
     cardano-base:testlib,
-    cardano-crypto-class,
+    cardano-crypto-class >=2.6,
```
lehins:
```suggestion
    cardano-crypto-class,
```

cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -78,7 +78,7 @@ test-suite tests
     bytestring,
     cardano-base:testlib,
     cardano-binary,
```
lehins:
```suggestion
    cardano-binary >= 1.9.1,
```

cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -78,7 +78,7 @@ test-suite tests
...
-    cardano-crypto-class,
+    cardano-crypto-class >=2.6,
```
lehins:
```suggestion
    cardano-crypto-class >=2.5.2,
```

cardano-crypto-leios/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for `cardano-crypto-leios`
 
+## 0.1.1.0
```
lehins: This is a patch version change (not visible to the user)
```suggestion
## 0.1.0.1
```

cardano-crypto-leios/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+
+* Change lower bound of `cardano-crypto-class` to `>=2.6`
```
lehins: Hence changelog entry is not needed
```suggestion
* 
```

cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: cardano-crypto-leios
-version: 0.1.0.0
+version: 0.1.1.0
```
lehins:
```suggestion
version: 0.1.0.1
```

lehins (approved):
Awesome work!
Thank you!

PR #667, opened 2026-06-04
Fix initial version of cardano-crypto-wallet

lehins:
# Description

This is a PR that fixes initial version of `cardano-crypto-wallet` that was added in #653 and reverts #666

Collection of issues that is fixed:
* integer overflows in multiple places
* interface that promoted leakage of EncryptedKey and no guarantees for finalization
* incorrect usage of [`crypto_aead_xchacha20poly1305_ietf_encrypt`](https://libsodium.gitbook.io/doc/secret-key_cryptography/aead/chacha20-poly1305/xchacha20-poly1305_construction#combined-mode). Libsodium API specified:
  > At most `mlen` + `crypto_aead_xchacha20poly1305_ietf_ABYTES` bytes are put into `c`, and the actual number of bytes is stored into `clen` ...
  
  while previous implementation assumed exact number of bytes, not *at most*
* Reduction in number of redundant arguments passed to C
* Remove redundant allocations
* Improve type safety of the implementation
* Ensure that the salt size is static at the type level and not customizable with `KdfParams`
* Drop duplicate values from the `Envelope`
* Expose all safe parts of the implementation as public API
* Reduce duplication in the test suite

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-wallet/cbits/encrypted_sign.c
+++ b/cardano-crypto-wallet/cbits/encrypted_sign.c
@@ -407,7 +407,6 @@ int wallet_sodium_xchacha20poly1305_encrypt(
 int wallet_sodium_xchacha20poly1305_decrypt(
 	uint8_t *secret_to_decrypt,
 	uint8_t const *ciphertext,
-	unsigned long long ciphertext_len,
 	uint8_t const tag[crypto_aead_xchacha20poly1305_ietf_ABYTES],
 	uint8_t const *aad,
 	unsigned long long aad_len,
@@ -420,17 +419,14 @@ int wallet_sodium_xchacha20poly1305_decrypt(
 	if (ensure_sodium() != 0) {
 		return 1;
 	}
-	if (ciphertext_len != UNENCRYPTED_KEY_SIZE) {
-		return 1;
-	}
-	memcpy(combined, ciphertext, (size_t) ciphertext_len);
-	memcpy(combined + ciphertext_len, tag, crypto_aead_xchacha20poly1305_ietf_ABYTES);
+	memcpy(combined, ciphertext, (size_t) UNENCRYPTED_KEY_SIZE);
+	memcpy(combined + UNENCRYPTED_KEY_SIZE, tag, crypto_aead_xchacha20poly1305_ietf_ABYTES);
 	if (crypto_aead_xchacha20poly1305_ietf_decrypt(
 		secret_to_decrypt,
 		&plen,
 		NULL,
 		combined,
-		ciphertext_len + crypto_aead_xchacha20poly1305_ietf_ABYTES,
+		UNENCRYPTED_KEY_SIZE + crypto_aead_xchacha20poly1305_ietf_ABYTES,
 		aad,
 		aad_len,
 		nonce,
@@ -439,5 +435,5 @@ int wallet_sodium_xchacha20poly1305_decrypt(
 		return 1;
 	}
 	secure_clear(combined, sizeof(combined));
-	return (plen == ciphertext_len) ? 0 : 1;
+	return (plen == UNENCRYPTED_KEY_SIZE) ? 0 : 1;
 }
--- a/cardano-crypto-wallet/cbits/encrypted_sign.c
+++ b/cardano-crypto-wallet/cbits/encrypted_sign.c
@@ -347,7 +347,6 @@ int wallet_sodium_randombytes(void * const out, size_t const out_len)
 }
 
 int wallet_sodium_argon2id(uint8_t *out,
-	unsigned long long out_len,
 	uint8_t const *pass,
 	unsigned long long pass_len,
 	uint8_t const salt[crypto_pwhash_SALTBYTES],
@@ -357,11 +356,8 @@ int wallet_sodium_argon2id(uint8_t *out,
 	if (ensure_sodium() != 0) {
 		return 1;
 	}
-	if (out_len != 32) {
-		return 1;
-	}
 	return crypto_pwhash(out,
-		out_len,
+		crypto_aead_xchacha20poly1305_ietf_KEYBYTES,
 		(const char *) pass,
 		pass_len,
 		salt,
--- a/cardano-crypto-wallet/cbits/encrypted_sign.c
+++ b/cardano-crypto-wallet/cbits/encrypted_sign.c
@@ -16,14 +16,14 @@ static int ensure_sodium(void)
 }
 
 #define SECRET_KEY_SEED_SIZE 32
-#define ENCRYPTED_KEY_SIZE   64
+#define UNENCRYPTED_KEY_SIZE 64
 #define PUBLIC_KEY_SIZE      32
 #define CHAIN_CODE_SIZE      32
 
-#define FULL_KEY_SIZE (ENCRYPTED_KEY_SIZE + PUBLIC_KEY_SIZE + CHAIN_CODE_SIZE)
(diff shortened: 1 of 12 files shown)


lehins:
@perturbing and @disassembler, this PR is ready for review

disassembler (approved):
I integrated this and confirmed that I can migrate from pbkdf2-hmac-sha512 to argon2id-v2.

cardano-crypto-wallet/cbits/encrypted_sign.c
```diff
@@ -16,97 +16,92 @@ static int ensure_sodium(void)
...
 #define SECRET_KEY_SEED_SIZE 32
-#define ENCRYPTED_KEY_SIZE   64
+#define UNENCRYPTED_KEY_SIZE 64
 #define PUBLIC_KEY_SIZE      32
 #define CHAIN_CODE_SIZE      32
 
-#define FULL_KEY_SIZE (ENCRYPTED_KEY_SIZE + PUBLIC_KEY_SIZE + CHAIN_CODE_SIZE)
+#define FULL_KEY_SIZE (UNENCRYPTED_KEY_SIZE + PUBLIC_KEY_SIZE + CHAIN_CODE_SIZE)
 
 typedef struct {
```
perturbing, line 29:
```suggestion
/* This captures the unencrypted BIP-39-Ed25519 extended root key `root = kL ‖ kR ‖ c ` as per CIP-1852.
 * where `kL` is the left half of the extended Ed25519 secret scalar,
 * `kR` is its right half (used only as keying material during child derivation),
 * and `c` is the 32-byte chain code. */
typedef struct {
```
And given this comment, perhaps we should add the comment
```
typedef struct {
    uint8_t skey[64]; // kL ‖ kR
    uint8_t pkey[32]; // A = compress(kL · B)
    uint8_t cc[32];   // c
} key_material;
```

Minor; given that, perhaps we should also rename `key_material` -> `extended_key_material`

cardano-crypto-wallet/cbits/encrypted_sign.c
```diff
@@ -16,97 +16,92 @@ static int ensure_sodium(void)
...
+	uint8_t skey[UNENCRYPTED_KEY_SIZE];
 	uint8_t pkey[PUBLIC_KEY_SIZE];
 	uint8_t cc[CHAIN_CODE_SIZE];
-} encrypted_key;
+} key_material;
 
 typedef struct {
 	uint8_t pkey[PUBLIC_KEY_SIZE];
 	uint8_t cc[CHAIN_CODE_SIZE];
 } public_key;
```
perturbing: Consider renaming this to the `extended_public_key` (that is, the combination of the public key and chain code, as per CIP-1852

cardano-crypto-wallet/test/Test/Cardano/Crypto/Wallet/V2FormatSpec.hs
```diff
@@ -23,128 +26,125 @@ wrongPass = BS.replicate 32 0x00
 -- ---------------------------------------------------------------------------
 -- Public-key golden vector
```
perturbing, line 27: What is the source of this test vector? Perhaps good to add as well. I know that the cip has multiple test vectors (for each variant (ledger/trezor/icarus)).

I guess the Icarus one is the interesting one? If so then

**CIP-3 Icarus test vectors** (from the
[CIP-0003 Icarus annex](https://cips.cardano.org/cip/cip-0003/annex/Icarus)),
15-word phrase
`eight country switch draw meat scout mystery blade tip drift useless good keep usage title`.

With an **empty passphrase** (`""`):

```
c065afd2832cd8b087c4d9ab7011f481ee1e0721e78ea5dd609f3ab3f156d245   (kL)
d176bd8fd4ec60b4731c3918a2a72a0226c0cd119ec35b47e4d55884667f552a   (kR)
23f7fdcd4a10c6cd2c7393ac61d877873e248f417634aa3d812af327ffe9d620   (c)
```

With the **passphrase `"foo"`**
```
70531039904019351e1afb361cd1b312a4d0565d4ff9f8062d38acf4b15cce41   (kL)
d7b5738d9c893feea55512a3004acb0d222c35d3e3d5cde943a15a9824cbac59   (kR)
443cf67e589614076ba01e354b1a432e0e6db3b59e37fc56b5fb0222970a010e   (c)
```

perturbing (changes requested):
A few remarks regarding names and testing. 

Also, I think that there are 3 places where we do not properly clear private material in `cardano-crypto-wallet/cbits/encrypted_sign.c`

1. in `add_left()`; the `zl8` buffer contains the scalar material but is never cleared
2. Similarly in `add_left_public()`
3. `xchacha20poly1305_decrypt()`: An edge case that should not happen, but if decrypt succeeds but `plen != UNENCRYPTED_KEY_SIZE` it will error without clearing `secret_to_decrypt`.

cardano-crypto-wallet/test/Test/Cardano/Crypto/Wallet/V2FormatSpec.hs
lehins: @disassembler This question is for you, since you are the original author of this package.

lehins:
Fixed: 
> in add_left(); the zl8 buffer contains the scalar material but is never cleared

Fixed: 
> Similarly in add_left_public()

This one is not correct, since GHC's RTS will trigger libsodium functionality for memory clearing when `mlsbCreate` exits or when Haskell's type `SecretKey` gets GCed. Checkout how `MLockedSizedBytes` works in order to get an idea why this is the case. In other words `secret_to_decrypt` is guaranteed to be cleared.
> xchacha20poly1305_decrypt(): An edge case that should not happen, but if decrypt succeeds but plen != UNENCRYPTED_KEY_SIZE it will error without clearing secret_to_decrypt.

cardano-crypto-wallet/cbits/encrypted_sign.c
lehins: This is fun. This struct is not even used :sweat_smile: 
Removed.

cardano-crypto-wallet/cbits/encrypted_sign.c
lehins: I've added the comments, but I'd prefer to get rid of this struct completely in a future iteration, instead of renaming it. Reason for this is that current implementation allocates the whole of struct in libsodium's mlocked memory, which is scarce resource. We really only need `skey`, while allocating `pkey` and `cc` in regular memory. For this to happen, we need to get rid of this struct.
That being said, this is not an urgent matter, so instead I'll create a ticket about it and we'll take care of it later.

lehins:
@perturbing Addressed all comments

cardano-crypto-wallet/test/Test/Cardano/Crypto/Wallet/V2FormatSpec.hs
disassembler: The `expectedPublicKey` bytes in `V2FormatSpec.hs` were bootstrapped: the code
was run once with `testSeed = BS.replicate 32 0x02 and testCC = BS.replicate
32 0xAB`, the public key it produced was extracted, and those bytes were
hardcoded back as the expected value.

There is no external reference. It is a self-seeded golden test — useful for
detecting silent regressions in the C code, but it does not verify
correctness against an independent standard.

The CIP-3 Icarus vectors are more valuable: they provide
an independently-specified output that a conforming implementation must
produce, so a match is meaningful rather than circular.

perturbing (approved):
Thanks, looks awesome!

PR #670, opened 2026-06-11
Scaffold an initial cardano-crypto-leios package

### LINKED ISSUE #790: Prototype Voting
## Why
Within our prototyping effort, we also want to "trace the path" of voting and uncover any problems on it. While #774 deliberately takes the short-cut of treating any EB as certified, thus not needing voting. We want to close that gap and see vote traffic affect the prototype networks already.

Furthermore, we are currently also discussing the committee selection scheme #856. Within that, we are going to need evidence about the vote traffic load #805 from simulations, but ideally can also validate the simulation behavior with a more "real" implementation. Lastly, adding voting to the prototype will test drive the network protocol and message structures.

We want to have this self-contained and not require #776 for this to work.

## What
- Out of scope: key registration and committee selection
  - Re-use existing key material (hackily re-interpret Ed25519 keys as BLS key material)
  - All pools are voters ("everyone votes" scheme in https://github.com/cardano-scaling/leios-wfa-ls-demo/issues/2 or https://github.com/IntersectMBO/ouroboros-consensus/pull/1975)
- Create, validate and propagate votes
  - A basic voting thread issues votes on EBs that are not too old (between 3 * `L_hdr` and the following `L_vote`)
  - Out of scope: Equivocation detection and EB validation
  - Votes are kept in memory (no persistence)
  - Votes are sent to downstream nodes (if they requested more votes)
  - Votes are individually validated before relayed
- Collect votes and aggregate a certificate
  - Trace all received votes and progress towards threshold (hard-coded)
  - Once threshold is reached, create a certificate and keep it in memory
  - Out of scope: use the certificate in block production (maybe easy after #774)

## How
- Drive development using `ThreadNet` property tests
- ~~Add vote messages to LeiosNotify/LeiosFetch mini-protocols~~
- Sketch `LeiosVotes` protocol as speculatively discussed in https://github.com/cardano-foundation/CIPs/pull/1167
  - [ ] https://github.com/cardano-scaling/cardano-blueprint/pull/65
- Create a basic vote forge loop
  - Maybe be optimistically driven by completing a tx closure
  - Trace EB candidates and when voted (don't spam)
- Implement a simple, but deduplicating vote fetching logic
  - Using an in-memory a `LeiosVotingState`
- Store certs along side votes in the `LeiosVotingState`

ch1bo:
Adds a new package for leios cryptographic types and operations. This was done in course of https://github.com/IntersectMBO/ouroboros-consensus/pull/2068, I'm currently integrating this with the `cardano-ledger` `master` and expect a follow-up PR there.

The digital signature scheme is BLS12-381 and fixed in the module. Contrary to the CIP-164, the certificate does not contain a slot or `EbHash` anymore. This makes definition in `cardano-base` a lot easier and in the current block structure design, the "message" against which the certificate is signed would be available from the (block) context in which the certificate is used.

Most importantly, this module contains encoders/decoders for the `LeiosCert` type including roundtrip and golden tests. This should be enough for the `cardano-ledger` to use this type confidently in `Dijkstra` era blocks.

There are also property tests about aggregating and verifying certificates. The `Committee` is part of this package, but how it is selected is deliberately kept out of scope.

image


cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -0,0 +1,81 @@
...
+  import: base, project-config
+  visibility: public
+  hs-source-dirs: testlib
+  exposed-modules:
+    Test.Cardano.Crypto.Leios.Gen
+
+  build-depends:
+    cardano-crypto-class,
+    cardano-crypto-leios,
+    hedgehog,
```
lehins: Hedgehog is not used in this repository. Please switch to QuickCheck with `hspec` combination.

```suggestion

```

cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -0,0 +1,81 @@
...
+  main-is: Main.hs
+  other-modules:
+    Test.Cardano.Crypto.Leios
+
+  build-depends:
+    bytestring,
+    cardano-binary:{cardano-binary, testlib},
+    cardano-crypto-class,
+    cardano-crypto-leios:{cardano-crypto-leios, testlib},
+    cardano-prelude-test,
```
lehins: `cardano-prelude` and `cardano-prelude-test` are not used except for some places in byron relevant code
Please avoid this dependency.
```suggestion
```

cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -0,0 +1,81 @@
...
+    containers,
+    hedgehog,
```
lehins: Ditto
```suggestion
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+--
+-- This package intentionally does not provide committee selection — sampling
+-- voters from the active stake distribution lives in consensus/ledger.
+-- However, callers are responsible for ensuring that every voter's BLS
+-- proof-of-possession has been verified before a 'Committee' value is built;
+-- 'verifyLeiosCert' and 'aggregateLeiosCert' both rely on this invariant to
+-- skip per-key PoP checks (they use 'uncheckedAggregateVerKeysDSIGN' /
+-- 'aggregateSigsDSIGN' under the hood). Passing in unchecked keys defeats
+-- the security of the aggregate signature.
+newtype Committee = Committee {committeeVoters :: Vector (Weight, LeiosVerificationKey)}
```
lehins: `Vector` is lazy, which will easily lead to space leaks.
Having a lazy tuple as an element will exacerbate this issue.

At the very least I suggest creating a proper data type for the vector element, instead of using a tuple and be really diligent in forcing elements of a vector. Something like that:

```suggestion
newtype Committee = Committee {committeeVoters :: Vector LeiosVoter}
```

```haskell
data LeiosVoter = LeiosVoter
  { leiosVoterWeight :: !Weight
  , leiosVoterVerKey :: !LeiosVerificationKey
  }
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+    <$> (bitFieldFromBytes <$> decodeBytes)
+    <*> decodeSigDSIGN
+
+-- ** Construction
+
+data AggregationError
+  = -- | A voter index in the contributions is past the committee bound.
+    VoterIdOutOfBounds !VoterId
+  | -- | BLS signature aggregation failed (e.g. malformed input signature).
+    BLSAggregationFailed !String
```
lehins: List is lazy. Making it strict in WHNF is not very useful. Therefore it is best to switch to Text
```suggestion
    BLSAggregationFailed !Text
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+  Either AggregationError LeiosCert
+aggregateLeiosCert committee contributions = do
+  let n = committeeSize committee
+      entries = Map.toAscList contributions
+  case [v | (v, _) <- entries, fromIntegral (voterIndex v) >= n] of
+    v : _ -> Left (VoterIdOutOfBounds v)
+    [] -> pure ()
+  aggSig <-
+    case aggregateSigsDSIGN (map snd entries) of
+      Left e -> Left (BLSAggregationFailed e)
```
lehins: From my previous suggestion
```suggestion
      Left e -> Left (BLSAggregationFailed (T.pack e))
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+-- 'verifyLeiosCert', and the CBOR codecs handle on-wire transport.
+
+-- | The @signers@ bitfield of a 'LeiosCert': a @⌈committeeSize\/8⌉@-byte
+-- MSB-first packed-bits representation of which committee voters contributed
+-- to the aggregate signature.
+--
+-- A 'newtype' wrapper around 'ByteArray' so type signatures throughout the
+-- aggregate / verify path say what they're working on, and so the on-wire
+-- form cannot be accidentally confused with arbitrary @bytes@.
+newtype BitField = BitField {bitFieldBytes :: ByteArray}
```
lehins, line 413: Why don't you use an existing package for handling this?
This is a very reputable package with a reputable maintainer: https://hackage.haskell.org/package/bitvec 

It seems counter productive to use efficient `ByteArray` for representation and then resort to lists and conversion to pinned ByteString for various operations on it.

cardano-crypto-leios/testlib/Test/Cardano/Crypto/Leios/Gen.hs
```diff
@@ -0,0 +1,81 @@
+{-# LANGUAGE TypeApplications #-}
+
+-- | Hedgehog generators for 'Cardano.Crypto.Leios' types, intended for
+-- downstream test suites (e.g. @cardano-ledger@) that want a real
+-- structurally-valid 'LeiosCert' without depending on the BLS plumbing.
```
lehins:
```suggestion
-- | QuickCheck generators for 'Cardano.Crypto.Leios' types, intended for
-- downstream test suites (e.g. @cardano-ledger@) that want a real
-- structurally-valid 'LeiosCert' without depending on the BLS plumbing.
```

Ledger does not use `hedgehog`!

.gitattributes
```diff
@@ -4,3 +4,8 @@
 
 nix/.stack.nix/*.nix linguist-generated=true
 nix/sources.nix linguist-generated=true
+
+# Golden test files are byte-exact hex dumps compared verbatim against the
+# encoder output. Without this, Windows checkouts translate LF -> CRLF and
+# the comparison fails (see cardano-crypto-leios:test:tests on Windows CI).
+cardano-crypto-leios/test/golden/** text eol=lf
```
lehins: Golden binary tests should be using binary format, not hex. 
```suggestion
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+-- ** Verification
+
+data VerificationError
+  = -- | 'signers' bitfield is longer than @⌈committeeSize/8⌉@ bytes.
+    MalformedSigners
+  | -- | The aggregate-BLS verification failed (wrong message, tampered
+    -- signature, or a bitfield/aggregate mismatch).
+    InvalidSignature
+  | -- | Sum of signers' weights is below the required threshold.
+    InsufficientWeight {got :: !Weight, required :: !Weight}
```
lehins: Hell no! This is never getting into master!
```suggestion
    InsufficientWeight !Weight !Weight
```

If you really like to distinguish between received and expected, I'd recommend either:
* creating a separate type in this module
* or even better moving our specialized ledger type `Mismatch` into `cardano-base:Cardano.Base.Mismatch` and using it here: https://github.com/IntersectMBO/cardano-ledger/blob/c80260469ac695aca97bb322a6ec74f9f5ab4b66/libs/cardano-ledger-core/src/Cardano/Ledger/BaseTypes.hs#L824-L829


```suggestion
    InsufficientWeight !(Mismatch RelGTEQ Weight)
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+-- Named fields on the 'InsufficientWeight' variant of 'VerificationError'
+-- are wanted at the call site; the project-wide @-Wpartial-fields@ would
+-- otherwise reject record syntax on a single sum-type alternative.
+{-# OPTIONS_GHC -Wno-partial-fields #-}
```
lehins: With all due respect, I will never allow this warning to be turned off!
Partial field accessors should have never been a language feature to begin with
```suggestion
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+
+-- | Cryptographic data types and operations used in Leios per CIP-164. Leios
+-- uses BLS12-381 MinSig as its signature scheme and defines a 'LeiosCert' that
+-- can be included into blocks. This module deliberately not includes a
+-- 'LeiosVote' because the vote itself is not an artifact that is on-chain.
+module Cardano.Crypto.Leios where
```
lehins: Please using explicit exports. We need to make sure no internal function leak out into the interface and we want to make sure ghc does it best at optimizing code in this module

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+-- | Plain CBOR encoder for 'LeiosCert', matching the CDDL in 'LeiosCert'.
+encodeLeiosCert :: LeiosCert -> Encoding
+encodeLeiosCert cert =
+  encodeListLen 2
+    <> encodeBytes (bitFieldToBytes cert.signers)
+    <> encodeSigDSIGN cert.aggregatedSignature
+
+-- | Plain CBOR decoder for 'LeiosCert', matching the CDDL in 'LeiosCert'.
+decodeLeiosCert :: Decoder s LeiosCert
+decodeLeiosCert = do
```
lehins, line 256: This decoder will be unusable in ledger, since it doesn't support indefinite length encoding

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+-- This is the only way to construct a 'LeiosCert' from outside the package;
+-- the bitfield layout is an internal wire-format detail.
+aggregateLeiosCert ::
+  Committee ->
+  Map VoterId LeiosSignature ->
+  Either AggregationError LeiosCert
+aggregateLeiosCert committee contributions = do
+  let n = committeeSize committee
+      entries = Map.toAscList contributions
+  case [v | (v, _) <- entries, fromIntegral (voterIndex v) >= n] of
```
lehins: Creating a separate binding `entries` will cause that list to be allocated. However if we use Map directly in both places where it is used then we will tap into list fusion and avoid that list being allocated. Also we need to make sure all `fromIntegral` in this repo are type annotated
```suggestion
  case [v | v <- Map.keys contributions, fromIntegral @Word32 @Int (voterIndex v) >= n] of
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+    v : _ -> Left (VoterIdOutOfBounds v)
+    [] -> pure ()
+  aggSig <-
+    case aggregateSigsDSIGN (map snd entries) of
```
lehins:
```suggestion
    case aggregateSigsDSIGN (Map.elems contributions) of
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+-- identities and eligibility proofs are not carried in the certificate;
+-- 'signers' is a @⌈N\/8⌉@-byte bitfield over the committee where bit @i@ is
+-- set iff voter index @i@ signed.
+--
+-- Producers should build 'LeiosCert' values via 'aggregateLeiosCert' and
+-- consumers verify them via 'verifyLeiosCert'; the bitfield layout is an
+-- implementation detail of the wire format.
+data LeiosCert = LeiosCert
+  { signers :: !BitField
+  , aggregatedSignature :: !LeiosSignature
```
lehins, line 240: This is a pretty terrible naming, since `signers` can easily be a local binding anywhere in the cardano-node codebase. I suggest something more descriptive like:

```suggestion
  { leisCertSigners :: !BitField
  , leisCertSignature :: !LeiosSignature
  -- ^ Aggregated BLS signature
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,348 @@
...
+    case uncheckedAggregateVerKeysDSIGN (reverse vks) of
+      Left _ -> Left InvalidSignature
+      Right k -> Right k
+  case verifyDSIGN leiosSignContext aggVk msg cert.aggregatedSignature of
+    Left _ -> Left InvalidSignature
+    Right () -> Right got
+  where
+    -- The bitfield decoder already enforced @i < n@; if the committee is
+    -- shorter than the decoder's idea of @n@ we treat it as a malformed cert.
+    accumSigner voters (w, ks) i = case voters V.!? i of
```
lehins: We should avoid space leaks as much as we can!
```suggestion
    accumSigner voters (!w, !ks) i = case voters V.!? i of
```

lehins (changes requested):
Consistency is one of the most important parts in software development. It is important to use consistent dependencies as the rest of the project, in this case cardano-base repo being that project.

cardano-crypto-leios/cardano-crypto-leios.cabal
ch1bo: Fixed in a52fe70690cc7ad5bd5571adb1b274557c138c9e

cardano-crypto-leios/cardano-crypto-leios.cabal
ch1bo: Fixed in a52fe70690cc7ad5bd5571adb1b274557c138c9e

cardano-crypto-leios/cardano-crypto-leios.cabal
ch1bo: Fixed in a52fe70690cc7ad5bd5571adb1b274557c138c9e

cardano-crypto-leios/testlib/Test/Cardano/Crypto/Leios/Gen.hs
ch1bo: Fixed in a52fe70690cc7ad5bd5571adb1b274557c138c9e

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: True, I should have turned off [Field selectors](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/field_selectors.html) instead. But your reaction tells me that this would also not be fine for you?

I can add a sub-type (as you suggested) for the one case it's used in this module, no problem.

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Resolved in https://github.com/intersectmbo/cardano-base/commit/510055053dc3dcefdfe1f822f1994d3a5e620ea3

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Resolved in https://github.com/intersectmbo/cardano-base/commit/510055053dc3dcefdfe1f822f1994d3a5e620ea3

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Resolved in https://github.com/intersectmbo/cardano-base/commit/510055053dc3dcefdfe1f822f1994d3a5e620ea3

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Thanks for the pointer. Looked into bitvec, but it uses LSB byte order and we'd need to reverse it. However, I have revised the construction to use a mutable byte array and cleaned up the indexing function. Can you have another look?

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Addressed in https://github.com/intersectmbo/cardano-base/commit/510055053dc3dcefdfe1f822f1994d3a5e620ea3

I went with a dedicated type, because this `InsufficientWeight !Weight !Weight` easily results in mixing the two like I encountered in `cardano-ledger-binary` [here](https://github.com/IntersectMBO/cardano-ledger/pull/5872#discussion_r3426371971)

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Done in e85caae2c4183b7a1cae3a274572984e6ee599a5

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Added support for indefinite length encoding and tests in 82f6884fd7e70913053e6c51b8b822c52670ca51

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Refactored now, please have another look

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: I deliberately wanted to match the CDDL as close as possible. The call sites should be all in this module, outside construction and inspection are not really intended (to be convenient) and we could even choose to not export the field selectors. However, I didn't want to be too prescriptive on this type.

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Thanks, added

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+-- | Find a voter's 'VoterId' on the 'Committee' by its
+-- 'LeiosVerificationKey', or 'Nothing' if the key is not on the committee.
+--
+-- If the committee carries duplicate verification keys, returns the smallest
+-- index matching @vk@ (committee selection is expected to deduplicate, but
+-- this module does not enforce it).
+getVoterId :: LeiosVerificationKey -> Committee -> Maybe VoterId
+getVoterId vk committee =
+  VoterId . fromIntegral @Int @Word16
+    <$> V.findIndex ((== vk) . voterVKey) committee.committeeVoters
```
lehins: Where is `getVoterId` going to be used. There are two issues with this function:
* It has O(n) complexity, which can lead to significant performance degradation if used in an inner loop.
* It is susceptible to integer overflow for a committee size larger than 65535

In theory neither one of these is dangerous, as long as we are sure these edge cases are not triggered. For example today we know that we can't have 65K stake pools, but in 20 years time this might change and this function could become a silent killer.

In any case, if it is only used in testing, I'd suggest moving it into the `testlib`. However if it will be used in production then we need to properly document those points and instead of simply calling `fromIntegral`, we should make this function partial upon overflow (eg. `error "Impossible: committee is too large")`.

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+  -- The bitfield must be exactly the canonical 'committee-many bits, padded
+  -- to a whole byte' length. Trailing bytes (zero-padded or otherwise) are
+  -- not accepted; the wire form is fixed for a given committee size.
+  when (sizeofByteArray (bitFieldBytes cert.signers) /= (n + 7) `div` 8) $
+    Left MalformedSigners
+  (got, vks) <- foldlM accumSigner (0, []) $ bitFieldMembers cert.signers
+  when (got < required) $
+    Left (InsufficientWeight WeightMismatch {got, required})
+  aggVk <-
+    uncheckedAggregateVerKeysDSIGN (reverse vks) -- REVIEW: reverse needed?
```
lehins: It is needed when used with `foldlM`. Alternatively you could use `foldrM`

Normally this would not be desired during validation, since:
* `foldrM` will not short-circuit on the very first failure. But that is ok for us, since we need to optimize the most common case which is the valid case
* with `foldrM` validation happens from the end of the list in the opposite direction, but since we only report `MalformedSigners` failure without any other information this point is irrelevant

That being said, it would be **a whole lot more efficient** to implement a manual nested loop over bits in the byte array, instead of converting BitArray to a list of indices and monadically folding over that list. That being said, such optimization can be postpone until a later time.
```suggestion
    uncheckedAggregateVerKeysDSIGN (reverse vks)
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+    InsufficientWeight !WeightMismatch
+  deriving stock (Eq, Show, Generic)
+  deriving anyclass (NFData)
+
+-- | The mismatch between the actual contributing weight and the minimum
+-- threshold a 'LeiosCert' is required to meet. Carried by
+-- 'InsufficientWeight'.
+data WeightMismatch = WeightMismatch
+  { got :: !Weight
+  , required :: !Weight
```
lehins: If the definition is exported into a public api, could I please ask you give those bindings decent names. Whatever you come up, but not something that is so short and not descriptive as `got`. Maybe:

```suggestion
  { weightReceived :: !Weight
  , weightRequired :: !Weight
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
lehins: I understand the desire, but I am speaking from experience people doing this precise mistake in the past of trying to name bindings in Haskell the same way as they were in the spec and it is not worth it!
It causes problems in such a huge codebase as ours.
So, there are only two options:
* If these fields are not used outside of this module, then they should not be exported. 
* If they are used outside of this module, then they need decent names. Do not worry about matching to the CDDL spec exactly, people can figure out what maps to what

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+--
+-- This package intentionally does not provide committee selection — sampling
+-- voters from the active stake distribution lives in consensus/ledger.
+-- However, callers are responsible for ensuring that every voter's BLS
+-- proof-of-possession has been verified before a 'Committee' value is built;
+-- 'verifyLeiosCert' and 'aggregateLeiosCert' both rely on this invariant to
+-- skip per-key PoP checks (they use 'uncheckedAggregateVerKeysDSIGN' /
+-- 'aggregateSigsDSIGN' under the hood). Passing in unchecked keys defeats
+-- the security of the aggregate signature.
+newtype Committee = Committee {committeeVoters :: Vector LeiosVoter}
```
lehins: Can we please call it `LeiosCommittee`, since we already have [`Committee`](https://github.com/IntersectMBO/cardano-ledger/blob/41ade60e9321fa5ab442fcf0b2d1a97fc0129344/eras/conway/impl/src/Cardano/Ledger/Conway/Governance/Procedures.hs#L560) in Ledger and it will be too confusing if we do not disambiguate
```suggestion
newtype LeiosCommittee = LeiosCommittee {leiosCommitteeVoters :: Vector LeiosVoter}
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+  -- 'nothunks' ships no instance for 'Data.Vector.Strict.Vector' and we don't
+  -- want to add an orphan. A WHNF-only check on the wrapper is sufficient here:
+  -- the strict 'Vector' forces every cell to WHNF, and a WHNF 'LeiosVoter'
+  -- forces both of its strict fields, so "Committee in WHNF" structurally
+  -- implies no thunks anywhere inside.
+  deriving (NoThunks) via OnlyCheckWhnfNamed "Committee" Committee
+
+-- | Number of seats in the committee.
+committeeSize :: Committee -> Int
+committeeSize Committee {committeeVoters} = length committeeVoters
```
lehins: Ditto
```suggestion
leiosCommitteeSize :: Committee -> Int
leiosCommitteeSize = length . leiosCommitteeVoters
```

cardano-crypto-leios/cardano-crypto-leios.cabal
```diff
@@ -0,0 +1,90 @@
...
+  build-depends:
+    bytestring,
+    cardano-base,
+    cardano-binary,
+    cardano-crypto-class,
+    cborg,
+    containers,
+    deepseq,
+    nothunks,
+    primitive,
```
lehins: To account for `runByteArray` usage
```suggestion
    primitive >= 0.7.3,
```

cardano-crypto-leios/test/Test/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,337 @@
...
+-- | Default committee size range exercised by the verify/aggregate properties.
+-- 1 ≤ n ≤ 16 covers single-voter (n=1), single-byte bitfield (n ≤ 8) and the
+-- two-byte boundary (n=9..16).
+genN :: QC.Gen Int
+genN = chooseInt (1, 16)
+
+genMsg :: QC.Gen BS.ByteString
+genMsg = do
+  len <- chooseInt (0, 64)
+  BS.pack <$> QC.vectorOf len QC.arbitrary
```
lehins: There is a more efficient way to generate `ByteString`s with [`genByteString`](https://github.com/IntersectMBO/cardano-base/blob/cf30402767f92ce6b690de2ecb59091c8e84fec2/cardano-base/testlib/Test/Cardano/Base/Bytes.hs#L38)
```suggestion
genMsg = chooseInt (0, 64) >>= genByteString
```

cardano-crypto-leios/testlib/Test/Cardano/Crypto/Leios/Gen.hs
```diff
@@ -0,0 +1,100 @@
...
+import Test.QuickCheck (Gen, arbitrary, choose, chooseInt, elements, shuffle, vectorOf)
+import Test.QuickCheck.Gen (unGen)
+import Test.QuickCheck.Random (mkQCGen)
+
+-- | Generate a 'LeiosSigningKey' from a uniformly random seed of the
+-- algorithm's expected size.
+genLeiosSigningKey :: Gen LeiosSigningKey
+genLeiosSigningKey = do
+  let seedLen = fromIntegral @Word @Int (seedSizeDSIGN (Proxy @LeiosDSIGN))
+  seedBytes <- BS.pack <$> vectorOf seedLen arbitrary
```
lehins: Ditto
```suggestion
  seedBytes <- genByteString seedLen
```

cardano-crypto-leios/testlib/Test/Cardano/Crypto/Leios/Gen.hs
```diff
@@ -0,0 +1,100 @@
...
+-- | Generate a real BLS 'LeiosSignature' by signing a random message with a
+-- freshly-generated signing key. Suitable as a byte-generator source for
+-- CDDL specs that need on-wire bytes which round-trip through
+-- 'Cardano.Crypto.DSIGN.rawDeserialiseSigDSIGN' — uniformly random 48-byte
+-- strings do /not/ decode to valid BLS G1 points and will crash there.
+genLeiosSignature :: Gen LeiosSignature
+genLeiosSignature = do
+  sk <- genLeiosSigningKey
+  msgLen <- choose (0, 256)
+  msg <- BS.pack <$> vectorOf msgLen arbitrary
```
lehins:
```suggestion
  sk <- genLeiosSigningKey
  msgLen <- choose (0, 256)
  msg <- genByteSring msgLen
```

cardano-crypto-leios/testlib/Test/Cardano/Crypto/Leios/Gen.hs
```diff
@@ -0,0 +1,100 @@
...
+-- Coverage of bitfield byte-length boundaries (CBOR uint widths > 256
+-- bytes) is not exercised here; that belongs in this package's own test
+-- suite, not in the shared testlib.
+genLeiosCert :: Gen LeiosCert
+genLeiosCert = do
+  n <- elements [1, 8, 9, 16, 17, 24]
+  sks <- vectorOf n genLeiosSigningKey
+  let committee =
+        Committee . V.fromList $
+          [LeiosVoter (1 / fromIntegral n) (deriveVerKeyDSIGN sk) | sk <- sks]
```
lehins: That is a much nicer way to convert to a Rational
```suggestion
          [LeiosVoter (1 % toInteger n) (deriveVerKeyDSIGN sk) | sk <- sks]
```

cardano-crypto-leios/testlib/Test/Cardano/Crypto/Leios/Gen.hs
```diff
@@ -0,0 +1,100 @@
...
+  k <- chooseInt (1, n)
+  signerIxs <- take k <$> shuffle [0 .. n - 1]
+  msgLen <- choose (0, 64)
+  msg <- BS.pack <$> vectorOf msgLen arbitrary
```
lehins:
```suggestion
  msg <- genByteString msgLen
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+-- the security of the aggregate signature.
+newtype Committee = Committee {committeeVoters :: Vector LeiosVoter}
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (NFData)
+  -- 'nothunks' ships no instance for 'Data.Vector.Strict.Vector' and we don't
+  -- want to add an orphan. A WHNF-only check on the wrapper is sufficient here:
+  -- the strict 'Vector' forces every cell to WHNF, and a WHNF 'LeiosVoter'
+  -- forces both of its strict fields, so "Committee in WHNF" structurally
+  -- implies no thunks anywhere inside.
+  deriving (NoThunks) via OnlyCheckWhnfNamed "Committee" Committee
```
lehins: We could ask for a new release of `nothunks`: https://github.com/input-output-hk/nothunks/blob/master/CHANGELOG.md#032---

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+      <*> decodeSigDSIGN
+  when isIndef $ do
+    isBreak <- decodeBreakOr
+    unless isBreak $
+      fail "LeiosCert: expected break after 2 elements of indefinite-length list"
+  pure cert
+
+data AggregationError
+  = -- | A voter index in the sigs is past the committee bound.
+    VoterIdOutOfBounds !VoterId
```
lehins: Why not make it a list?
Also failures don't really need to be strict, since they are never retained in the state
```suggestion
    VoterIdsOutOfBounds (NonEmpty VoterId)
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+-- This is the only way to construct a 'LeiosCert' from outside the package;
+-- the bitfield layout is an internal wire-format detail.
+aggregateLeiosCert ::
+  Committee ->
+  Map VoterId LeiosSignature ->
+  Either AggregationError LeiosCert
+aggregateLeiosCert committee sigs = do
+  case outOfBoundsVoterIds of
+    v : _ -> Left (VoterIdOutOfBounds v)
+    [] -> pure ()
```
lehins:
```suggestion
  case NE.nonEmpty outOfBoundsVoterIds of
    Just vs -> Left (VoterIdsOutOfBounds vs)
    Nothing -> Right ()
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+    isBreak <- decodeBreakOr
+    unless isBreak $
+      fail "LeiosCert: expected break after 2 elements of indefinite-length list"
+  pure cert
+
+data AggregationError
+  = -- | A voter index in the sigs is past the committee bound.
+    VoterIdOutOfBounds !VoterId
+  | -- | BLS signature aggregation failed (e.g. malformed input signature).
+    BLSAggregationFailed !Text
```
lehins: Ditto
```suggestion
    BLSAggregationFailed Text
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+      aggregateSigsDSIGN (Map.elems sigs)
+  pure LeiosCert {signers, aggregatedSignature}
+  where
+    outOfBoundsVoterIds =
+      [vid | vid <- Map.keys sigs, isNothing $ resolveVoter committee vid]
+
+    -- Builds directly into a mutable 'ByteArray' via a single allocation and
+    -- writes one bit per member of the input set.
+    signers = BitField $ runByteArray $ do
+      let len = (n + 7) `div` 8
```
lehins: `n` isn't used outside this binding, this will be cleaner:
```suggestion
      let n = committeeSize committee
          len = (n + 7) `div` 8
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+-- aggregate / verify path say what they're working on, and so the on-wire
+-- form cannot be accidentally confused with arbitrary @bytes@.
+newtype BitField = BitField {bitFieldBytes :: ByteArray}
+  deriving stock (Show, Eq, Generic)
+  deriving anyclass (NFData)
+  deriving (NoThunks) via OnlyCheckWhnfNamed "BitField" BitField
+
+-- | Encode a 'BitField' to CBOR bytes.
+encodeBitField :: BitField -> Encoding
+encodeBitField = encodeBytes . byteArrayToByteString . bitFieldBytes
```
lehins: Conversion to `ByteString` is unnecessary.
```suggestion
encodeBitField = toCBOR . bitFieldBytes
```

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+encodeLeiosCert cert =
+  encodeListLen 2
+    <> encodeBitField cert.signers
+    <> encodeSigDSIGN cert.aggregatedSignature
+
+-- | Plain CBOR decoder for 'LeiosCert', matching the CDDL in 'LeiosCert'.
+-- Accepts both definite-length and indefinite-length encodings of the
+-- outer 2-element array.
+decodeLeiosCert :: Decoder s LeiosCert
+decodeLeiosCert = do
```
lehins, line 256: To be honest with you, looking at this implementation makes me believe that encoders/decoders should be implemented in ledger in `cardano-ledger-binary` instead, because:
* we already have proper helpers for this stuff (for this case eg. `decodeRecordNamed`) 
* we'll probably have to ignore some of those decoders in this module anyways and rewrite them in ledger using versioned decoder.

We stopped using plain decoder/encoder for almost anything in ledger, except for byron era of course.

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -0,0 +1,419 @@
...
+  deriving anyclass (NFData)
+  deriving (NoThunks) via OnlyCheckWhnfNamed "BitField" BitField
+
+-- | Encode a 'BitField' to CBOR bytes.
+encodeBitField :: BitField -> Encoding
+encodeBitField = encodeBytes . byteArrayToByteString . bitFieldBytes
+
+-- | Decode a 'BitField' from CBOR bytes.
+decodeBitField :: Decoder s BitField
+decodeBitField = BitField . byteArrayFromByteString <$> decodeBytes
```
lehins, line 424: To drive my point about defining these decoders in ledger further: we have introduced indefinite length decoding for bytes for Dijkstra era, which plain `decodeBytes` does not support.

lehins (approved):
More feedback, none of which is terribly critical and we can fix it up later if needed.

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Resolved in https://github.com/intersectmbo/cardano-base/commit/9bfac4c6e697cbd8ddf6af60e30f661d0da54bd6

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Removed the type

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: I was following the CertifiedVRF precedent (plain codec next to the type in Cardano.Crypto.VRF.Class, wrapped via fromPlainEncoding/fromPlainDecoder in Cardano.Ledger.Binary.Crypto). Keeping the codec next to the type for cohesion as the current consumer is actually consensus, not ledger. That may shift later, and moving it to ledger likely makes sense once the codec becomes versioned.

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: https://github.com/intersectmbo/cardano-base/commit/4f89d33aed8f3a371f6204edb90942fc8c721ea6

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: https://github.com/intersectmbo/cardano-base/commit/4f89d33aed8f3a371f6204edb90942fc8c721ea6

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: https://github.com/intersectmbo/cardano-base/commit/4f89d33aed8f3a371f6204edb90942fc8c721ea6

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: https://github.com/intersectmbo/cardano-base/commit/4f89d33aed8f3a371f6204edb90942fc8c721ea6

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: https://github.com/intersectmbo/cardano-base/commit/4f89d33aed8f3a371f6204edb90942fc8c721ea6

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: https://github.com/input-output-hk/nothunks/issues/72

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: That being said, you would prefer orphan instances and a golden roundtrip in `cardano-ledger-binary` instead?

lehins: > That may shift later, and moving it to ledger likely makes sense once the codec becomes versioned.

It is already versioned. BlockHeader in both TPraos and Praos use versioned Decoder

> That being said, you would prefer orphan instances and a golden roundtrip in cardano-ledger-binary instead?

Yes, almost all of the serializarion testing logic lives in ledger

PR #671, opened 2026-06-19
Add cardano-config package

jasagredo:
NOTE: this PR was superseded by https://github.com/IntersectMBO/cardano-config/pull/1

Introduce a new `cardano-config` library that provides a single entry-point for parsing the `cardano-node` configuration.

It defines:

- A CLI options parser based on `optparse-applicative` (`parseCliArgs`).
- JSON/YAML parsing of the configuration files, derived from `autodocodec` codecs (`parseConfigurationFiles`).
- `resolveConfiguration`, combining both into a `NodeConfiguration`.

The configuration keys are grouped into components (Storage, Consensus, Mempool, Network, Protocol, Testing, Tracing). A bundled `cardano-config-schema` executable dumps the JSON Schema derived from the same codecs as the authoritative key listing.

Register the package in `cabal.project`.


cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,145 @@
...
+
+-- | Parse the configuration file, but do not parse the children files
+-- referenced from it yet.
+parseConfigurationFile ::
+  FilePath -> Value -> Parser (NodeConfigurationFromFileF IO)
+parseConfigurationFile root val =
+  withObject
+    "Configuration"
+    ( \v -> do
+        configVersion <- v .:? "ConfigurationVersion"
```
crocodile-dentist: This could be problematic if we want to change the V2 schema (eg. rename something, or move a key under some new object). If we moved the `ConfigurationVersion` to the top level, and have it wrap the configuration keys/sections, then evolving the format internally can be easier.

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,145 @@
...
+  Parser (IO a)
+subFileParser root sectionName p val = do
+  withObject
+    "Configuration"
+    ( \v -> do
+        v .:? fromString sectionName >>= \case
+          Just fn ->
+            pure $
+              do
+                exists <- doesFileExist $ root </> fn
```
crocodile-dentist: If the file name is misspelled it will try to parse with JSON and fail with a not so clear codec error.

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,145 @@
...
+  (Value -> Parser a) ->
+  Value ->
+  Parser (IO a)
+subFileParser root sectionName p val = do
+  withObject
+    "Configuration"
+    ( \v -> do
+        v .:? fromString sectionName >>= \case
+          Just fn ->
+            pure $
```
crocodile-dentist: This interleaves Parsing with IO, and the sequencing is performed by `parseConfigurationFiles`. In case of a parsing error in some sub-component, will a clear message be produced by `ConfigurationParsingError` identifying the offending file/line? Perhaps a more explicit approach can be taken to first identify whether we are parsing an object inline or it is a file path, and in the case of the second, a second pass reads the file and runs the codec on the resulting value. In this case, better error messages can be given.

cardano-config/src/Cardano/Configuration/File/Network.hs
```diff
@@ -0,0 +1,148 @@
...
+        <*> requiredField "softLimit" "Soft limit on the number of connections"
+          .= (\(AcceptedConnectionsLimit _ s _) -> s)
+        <*> requiredFieldWith "delay" diffTimeCodec "Delay, in seconds, applied once the soft limit is reached"
+          .= (\(AcceptedConnectionsLimit _ _ d) -> d)
+
+-- | Options related to Networking configuration. Most of the fields are
+-- @Maybe@ such that the networking layer can then set the appropriate
+-- defaults.
+data NetworkConfiguration = NetworkConfiguration
+  { pncDiffusionMode :: String
```
crocodile-dentist: Could we instead give this a `DiffusionMode` type (eg. similar in spirit to `LedgerDbBackendSelector`), and define things like:

```haskell
data DiffusionMode = InitiatorOnly | InitiatorAndResponder
  deriving (Generic, Show, Eq, Enum, Bounded)
  deriving (FromJSON, ToJSON) via (Autodocodec DiffusionMode)

instance HasCodec DiffusionMode where
  codec = shownBoundedEnumCodec
```

The schema would enumerate better values for tooling support and catch typos at parse time.

There are other things that can be handled this way, such as `byronReqNetworkMagic` which is just `RequiresNoMagic`, but I only give this one example to be brief.

cardano-config/src/Cardano/Configuration/File/Consensus.hs
```diff
@@ -0,0 +1,98 @@
...
+-- | The consensus mode is selected by the @ConsensusMode@ key; the Genesis
+-- flags only apply in Genesis mode, so they are read (from the
+-- @LowLevelGenesisOptions@ key) only then.
+instance HasCodec (ConsensusConfiguration Maybe) where
+  codec =
+    bimapCodec toConfig fromConfig $
+      object "ConsensusConfiguration" $
+        (,)
+          <$> optionalField "ConsensusMode" "Which consensus mode to run (PraosMode or GenesisMode)" .= fst
+          <*> optionalField "LowLevelGenesisOptions" "Low-level Genesis tuning (GenesisMode only)" .= snd
```
crocodile-dentist: Maybe use `discriminatedUnionCodec` to avoid silently dropping low level genesis options in case of using `PraosMode`?

cardano-config/README.md
```diff
@@ -0,0 +1,137 @@
...
+it with the bundled executable:
+
+```console
+$ cardano-config-schema          # the whole configuration
+$ cardano-config-schema --list   # the available components
+$ cardano-config-schema Storage  # one component
+```
+
+Keys that none of the parsers below recognise are **ignored** (parsing neither
+fails nor preserves them).
```
crocodile-dentist: This can be dangerous. Since we already expose a schema, we can be more strict and also add a `additionalProperties: False` to catch typos as errors. Or at least have a strict mode in `parseConfigurationFiles` that validates against `wholeConfigSchema` and emits a warning or error on unknown keys.

crocodile-dentist:
Some other general comments:

1. Maybe this could be moved to its own dedicated repo eventually, so as not to force `cardano-base` releases for configuration-related changes.
2. Some fields are marked with `optionalFieldWithDefault` (visible in schema) and others are `Maybe`. Would it be more principled to have the schema's job to be to just enumerate keys and their types, and the default values provided at the end?

cardano-config/src/Cardano/Configuration/CliArgs.hs
```diff
@@ -0,0 +1,395 @@
...
+        | null hostRev -> Left "parseHostPort: Empty host."
+        | null portRev -> Left "parseHostPort: Empty port."
+        | all isDigit portRev
+        , Just p <- readMaybe @PortNumber (reverse portRev) ->
+            if
+              | 0 <= p, p <= 65535 -> Right (Text.pack (reverse hostRev), p)
+              | otherwise -> Left ("parseHostPort: Numeric port '" ++ show p ++ "' out of range: 0 - 65535)")
+        | otherwise -> Left "parseHostPort: Non-numeric port."
+  | otherwise =
+      Left "parseHostPort: No colon found."
```
crocodile-dentist: This will fail on eg. `2001:db8::1`, where it will parse the host as `2001:db8::` and port as `1`. We can use `network-uri` to correctly parse IP's.

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,145 @@
...
+                    bs <- BS.readFile (root </> fn)
+                    json <- decodeThrow bs
+                    parseOrThrow p json
+                  else parseOrThrow p val
+          Nothing ->
+            pure $ parseOrThrow p val
+    )
+    val
+
+data ConfigurationParsingError = ConfigurationParsingError String
```
crocodile-dentist: We can improve it with:

```haskell
  data ConfigurationParsingError = ConfigurationParsingError   
    { errFile    :: Maybe FilePath        -- which sub-file, if any                                                            
    , errSection :: String                -- which top-level section                                                           
    , errPath    :: [JSONPathElement]     -- where in the JSON tree                                                            
    , errMessage :: String                -- aeson's "expected X, got Y"                                                       
    }             
    deriving (Show, Exception) 
```
The `JSONPathElement` gives us better structure for errors from nested values.

cardano-config/src/Cardano/Configuration/File.hs
jasagredo:
```
{
  "Version": 
  "Config": {
    ...
  }
}
```

We discussed this could probably be better, but have a decoder for the legacy format for some versions

cardano-config/src/Cardano/Configuration/File/Network.hs
jasagredo: We agreed on this

jasagredo:
> Maybe this could be moved to its own dedicated repo eventually, so as not to force cardano-base releases for configuration-related changes.

Let's piggyback on cardano-base until we have a satisfactory implementation, then I can move it to its own repo in one go, instead of merging it here :smile: 

> Some fields are marked with optionalFieldWithDefault (visible in schema) and others are Maybe. Would it be more principled to have the schema's job to be to just enumerate keys and their types, and the default values provided at the end?

`Maybe` are defaulted in the library, whereas `optionalFieldWithDefault` is defaulted in the node. Teams want the configs to live in their repos, while the whole spirit of this library necessitates of the defaults to live here. It seems we came to an agreement (maybe? still needs discussion) on having a verbatim copy of a file from network here, and a CI check that sees that they are real verbatim copies. That way network can temporarily modify their defaults in a branch, but when they commit a default, they will need to update it here too.

cardano-config/README.md
jasagredo: We agreed on the latter

cardano-config/test/Main.hs
```diff
@@ -0,0 +1,70 @@
...
+
+import Cardano.Configuration.File
+import Control.Exception (SomeException, evaluate, try)
+import Data.Aeson (eitherDecodeFileStrict')
+import System.Exit (exitFailure)
+
+main :: IO ()
+main = do
+  results <-
+    sequence
```
carbolymer, line 31: I think there is a risk for `config.schema.json` to drift apart from the codecs definition. You should consider adding a test checking that + CI gate.

cardano-config/src/Cardano/Configuration/Basics.hs
```diff
@@ -0,0 +1,13 @@
+-- | Basic types for configuration
+module Cardano.Configuration.Basics (
```
carbolymer: A nitpick: use singular nouns for all module names consistently.

cardano-config/src/Cardano/Configuration/File/Storage.hs
```diff
@@ -0,0 +1,177 @@
...
+-- or an explicit set of options.
+data SnapshotPolicy
+  = NamedSnapshotPolicy String
+  | CustomSnapshotPolicy SnapshotOptions
+  deriving (Generic, Show)
+
+instance HasCodec SnapshotPolicy where
+  codec =
+    dimapCodec toPolicy fromPolicy $
+      disjointEitherCodec (codec @String) (codec @SnapshotOptions)
```
carbolymer: I think `disjointEitherCodec` produces poor error messages here. For example when a `Snapshots` validation fails (e.g. zero interval or invalid delay range), the error includes the failure of the *other* branch of the disjoint union (the `"Mithril"` string form) which the user never intended to use.

**Input** (`zero-interval.json`):
```json
{
  "LedgerDB": {"Snapshots": {"SnapshotInterval": 0}},
  "ByronGenesisFile": "b.json",
  "ShelleyGenesisFile": "s.json",
  "AlonzoGenesisFile": "a.json",
  "ConwayGenesisFile": "c.json",
  "LastKnownBlockVersion-Major": 3,
  "LastKnownBlockVersion-Minor": 0
}
```

**Error**:
```
ConfigurationParsingError "Error in $.LedgerDB.Snapshots:
Both branches of a disjoint union failed:
Left:   Error in $: parsing Text failed, expected String, but encountered Object
Right:  Error in $.SnapshotInterval: Non-positive SnapshotInterval: 0
"
```

I think in your case if you're going to the right branch of the disjoint union parsing, only the right branch error message is relevant to the user here.

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,145 @@
...
+  FilePath -> Value -> Parser (NodeConfigurationFromFileF IO)
+parseConfigurationFile root val =
+  withObject
+    "Configuration"
+    ( \v -> do
+        configVersion <- v .:? "ConfigurationVersion"
+        case configVersion :: Maybe Int of
+          Nothing -> parseConfigurationVersion1 root val
+          Just 1 -> parseConfigurationVersion1 root val
+          _ -> fail $ "Unknown configuration version: " <> show configVersion
```
carbolymer:
```suggestion
          Just i -> fail $ "Unsupported configuration version: " <> show i
          _ -> fail $ "Unknown configuration version: " <> show configVersion
```

avoids having `Just` constructor in the rendered message

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,145 @@
...
+        v .:? fromString sectionName >>= \case
+          Just fn ->
+            pure $
+              do
+                exists <- doesFileExist $ root </> fn
+                if exists
+                  then do
+                    bs <- BS.readFile (root </> fn)
+                    json <- decodeThrow bs
+                    parseOrThrow p json
```
carbolymer: Similarly to what @crocodile-dentist reported - we don't know which file parsing fails if there are many specified. For example:


**Input** (`split-bad-sub.json`):
```json
{
  "Storage": "bad-storage.json",
  "ByronGenesisFile": "b.json",
  "ShelleyGenesisFile": "s.json",
  "AlonzoGenesisFile": "a.json",
  "ConwayGenesisFile": "c.json",
  "LastKnownBlockVersion-Major": 3,
  "LastKnownBlockVersion-Minor": 0
}
```

`bad-storage.json`:
```
not json at all
```

**Error**:
```
ConfigurationParsingError "Error in $: parsing StorageConfiguration failed, expected Object, but encountered String"
```

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,145 @@
...
+deriving instance Show (NodeConfigurationFromFileF (Either FilePath))
+deriving instance Show (NodeConfigurationFromFileF Identity)
+
+-- | If the consulted key is a filepath, then prepare an action to parse that
+-- other file.
+subFileParser ::
+  FilePath ->
+  String ->
+  (Value -> Parser a) ->
+  Value ->
```
carbolymer, line 140: Haddocks for the function parameters would be nice

cardano-config/src/Cardano/Configuration/Schema.hs
```diff
@@ -0,0 +1,143 @@
...
+-- draft-07-compatible schemas (only @oneOf@/@anyOf@/@const@/@properties@/
+-- @required@/@description@), and draft-07 is what validators such as ajv accept
+-- by default.
+draftURI :: Text
+draftURI = "http://json-schema.org/draft-07/schema#"
+
+-- | Make a schema friendlier to validators and editors: declare the draft and
+-- turn autodocodec's @$comment@ annotations into @description@s (which editors
+-- surface on hover and completion).
+polish :: Value -> Value
```
carbolymer: 🇵🇱

cardano-config/src/Cardano/Configuration/CliArgs.hs
```diff
@@ -0,0 +1,395 @@
+{-# LANGUAGE MultiWayIf #-}
+
+module Cardano.Configuration.CliArgs (
+  -- * CLI Arguments
+  CliArgs (..),
+  parseCliArgs,
+  ShutdownOn (..),
+
```
carbolymer, line 6: I think it would be useful to have all CLI parsers exported in some way from `cardano-config`. Some cli tools like cardano-cli could reuse e.g. port or socket parsing then.

cardano-config/src/Cardano/Configuration/File.hs
carbolymer: Definitely error messages should be improved.

cardano-config/src/Cardano/Configuration/File/Network.hs
```diff
@@ -0,0 +1,148 @@
...
+          .= (\(AcceptedConnectionsLimit h _ _) -> h)
+        <*> requiredField "softLimit" "Soft limit on the number of connections"
+          .= (\(AcceptedConnectionsLimit _ s _) -> s)
+        <*> requiredFieldWith "delay" diffTimeCodec "Delay, in seconds, applied once the soft limit is reached"
+          .= (\(AcceptedConnectionsLimit _ _ d) -> d)
+
+-- | Options related to Networking configuration. Most of the fields are
+-- @Maybe@ such that the networking layer can then set the appropriate
+-- defaults.
+data NetworkConfiguration = NetworkConfiguration
```
carbolymer: why all the fields are keeping `pnc` prefix?

cardano-config/src/Cardano/Configuration/File/Network.hs
```diff
@@ -0,0 +1,148 @@
...
+        <*> optionalField "ResponderCoreAffinityPolicy" "Whether responders are pinned to a core"
+          .= pncResponderCoreAffinityPolicy
+        <*> optionalField "ExperimentalProtocolsEnabled" "Enable experimental network protocols"
+          .= pncExperimentalProtocolsEnabled
+        <*> optionalField "TxSubmissionLogicVersion" "Which tx-submission inbound logic to run"
+          .= pncTxSubmissionLogicVersion
+        <*> optionalFieldWith "TxSubmissionInitDelay" diffTimeCodec "Tx-submission initial delay, in seconds"
+          .= pncTxSubmissionInitDelay
+
+-- | Connections for local clients
```
carbolymer: `LocalConnectionsConfig` stays partial (`Maybe` fields) even after resolution, unlike `StorageConfiguration`/`ConsensusConfiguration`/`ProtocolConfiguration` which use the `f` parameter to distinguish
  partial (`Maybe`) from resolved (`Identity`). This inconsistency means `resolveConfiguration` has to handle each sub-config differently rather than using a uniform resolution pattern.

carbolymer (dismissed):
The library is inconsistent about where defaults and validation live:

- Some fields bake in defaults at parse time via `optionalFieldWithDefault` (e.g. `DiffusionMode` defaults to `"InitiatorAndResponder"`, `EnableCSJ` defaults to `True`).
- Some fields stay `Maybe` because the downstream consumer owns the default (e.g. `ProtocolIdleTimeout`, `MaxConcurrencyBulkSync`).
- Some types use the `f` parameter to distinguish partial from resolved (`StorageConfiguration f`, `ConsensusConfiguration f`, `ProtocolConfiguration f`), but others use plain `Maybe` permanently (`NetworkConfiguration`, `LocalConnectionsConfig`, `MempoolConfiguration`, `TestingConfiguration`).

It would help to have a stated design principle in the README or module documentation:
- When should a default live in cardano-config vs be deferred to the consumer?
- Where should cross-field validation happen (e.g. "RPC enabled but no socket path")? At parse time? At resolution time? In the consumer?
- Should all sub-configs use the `f` parameter uniformly, so that `NodeConfiguration` carries `Identity`-wrapped fields guaranteeing completeness?

Without this, each new field or sub-config will make an ad-hoc choice and the inconsistency will grow.

cardano-config/src/Cardano/Configuration.hs
```diff
@@ -0,0 +1,247 @@
...
+  File.TracingConfiguration (..),
+  File.parseConfigurationFiles,
+  File.parseConfigurationFilesWith,
+  File.UnknownKeyPolicy (..),
+  File.ConfigurationParsingError (..),
+) where
+
+import qualified Cardano.Configuration.CliArgs as CLI
+import qualified Cardano.Configuration.Common as File
+import qualified Cardano.Configuration.File as File
```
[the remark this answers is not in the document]
jasagredo: Not an error — Haskell allows several modules under one qualifier; Common, File, File.Consensus, File.Protocol and File.Storage are all imported as File deliberately, so File. exposes the whole configuration surface from one prefix. The package builds (and CI runs cabal build/test). No change.

cardano-config/src/Cardano/Configuration.hs
```diff
@@ -0,0 +1,247 @@
...
+  ConfigResolutionError (..),
+
+  -- ** Storage
+  File.StorageConfiguration (..),
+  File.LedgerDbConfiguration (..),
+  File.SnapshotPolicy (..),
+  File.SnapshotOptions (..),
+  File.LedgerDbBackendSelector (..),
+  File.NodeDatabasePaths (..),
+
```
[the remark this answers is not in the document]
jasagredo: Same as above. No change

cardano-config/src/Cardano/Configuration/CliArgs.hs
```diff
@@ -0,0 +1,435 @@
...
+import Data.IP (IPv4, IPv6)
+import Data.Text (Text)
+import qualified Data.Text as Text
+import Data.Word (Word64)
+import Network.Socket (PortNumber)
+import Network.URI (URI (..), URIAuth (..), parseURIReference)
+import Options.Applicative
+import System.Posix.Types (Fd (..))
+import Text.Read (readEither, readMaybe)
+
```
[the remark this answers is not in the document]
jasagredo: Both are provided by Options.Applicative. No change.

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,365 @@
...
+    Object o
+      | Just config <- KM.lookup "Configuration" o -> pure (lookupVersion o, config)
+      | otherwise -> pure (lookupVersion o, value)
+    _ ->
+      throwIO $
+        ConfigurationParsingError Nothing Nothing [] "expected the configuration to be a JSON/YAML object"
+  where
+    lookupVersion o = case KM.lookup "Version" o of
+      Just (Number n) -> round n
+      _ -> 1
```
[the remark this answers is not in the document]
jasagredo: Good catch. splitEnvelope now parses Version with Data.Scientific.toBoundedInteger and raises a configurationParsingError for a non-integer or out-of-range value (a missing key still means legacy version 1).

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,365 @@
...
+import Control.Monad (unless)
+import Data.Aeson (FromJSON, Value (..), parseJSON)
+import qualified Data.Aeson.Key as K
+import qualified Data.Aeson.KeyMap as KM
+import Data.Aeson.Types (JSONPath, JSONPathElement (..), formatError, iparseEither)
+import Data.Foldable (toList)
+import Data.Functor.Identity (Identity (..))
+import Data.List (intercalate)
+import qualified Data.Text as T
+import qualified Data.Yaml as Yaml
```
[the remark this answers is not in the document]
jasagredo: added import Data.Scientific (toBoundedInteger) and scientific to the library dependencies.

cardano-config/README.md
```diff
@@ -0,0 +1,356 @@
...
+`parseCliArgs` is an `optparse-applicative` parser producing a `CliArgs` value.
+The flag names, metavars and help text match those historically accepted by
+`cardano-node`, so existing operator scripts keep working. The recognised flags
+are:
+
+| Group | Flag(s) | Metavar | Notes |
+| --- | --- | --- | --- |
+| | `--config` | `FILEPATH` | Main configuration file (defaults to `./configuration/cardano/mainnet-config.yaml`). |
+| | `--topology` | `FILEPATH` | Topology file (defaults to the mainnet `./configuration/cardano/mainnet-topology.yaml`). |
+| | `--socket-path` | `FILEPATH` | Socket for local clients; overrides `LocalConnections.SocketPath`. |
```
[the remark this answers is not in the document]
jasagredo: Fixed the README to mainnet-config.json / mainnet-topology.json, matching the CliArgs defaults

cardano-config/cardano-config.cabal
```diff
@@ -0,0 +1,127 @@
...
+
+category: Data
+build-type: Simple
+extra-doc-files:
+  CHANGELOG.md
+
+data-files:
+  defaults/**/*.json
+  examples/*.json
+
```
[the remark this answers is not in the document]
jasagredo: Fixed. Added schemas/*.json to data-files so the schema-drift test has the committed schemas available from a source distribution.

cardano-config/test/Main.hs
```diff
@@ -0,0 +1,158 @@
...
+              IO (Either String (LocalConnectionsConfig Maybe))
+          )
+      , parseCase "examples/fullconfig.json"
+      , parseCase "examples/split.json"
+      , parseCase "examples/split-all.json"
+      , listMergeCase
+      , shadowWarnCase
+      , shadowRejectCase
+      , resolveCase
+      ]
```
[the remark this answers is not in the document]
jasagredo: The packaging gap is fixed (above), so examples/ and schemas/ are now in the sdist and the tests resolve them from the package directory, which is the CWD under cabal test.

cardano-config/src/Cardano/Configuration/File.hs
jasagredo: Done. the format is now an optional { "Version": N, "Configuration": {…} } envelope, with Version at the top level wrapping the configuration object; a bare document is treated as legacy version 1. splitEnvelope reads the version and now validates it's an integer.

cardano-config/src/Cardano/Configuration/File.hs
jasagredo: Fixed. Sub-file references are resolved in an explicit pass (loadSectionSource): a path that doesn't exist now raises the referenced configuration file does not exist, naming the file and section, instead of being parsed as inline JSON.

cardano-config/src/Cardano/Configuration/File.hs
jasagredo: Reworked along these lines: each section is first classified (top-level keys, inline object, or a referenced sub-file/list), then read and run through the codec, rather than interleaving Parser with IO. Failures carry structured context (ConfigurationParsingError with file, section and JSON path).

cardano-config/src/Cardano/Configuration/File/Network.hs
jasagredo: Done

cardano-config/src/Cardano/Configuration/File/Consensus.hs
jasagredo: Fixed. The ConsensusMode codec now rejects LowLevelGenesisOptions unless the mode is GenesisMode (it errors rather than dropping them), so the options can't be silently ignored.

cardano-config/README.md
jasagredo: Added RejectUnknownKeys: parseConfigurationFilesWith RejectUnknownKeys rejects unrecognised keys (the default warns). The whole-config schema also now encodes section/top-level exclusivity, and ajv validate against it is documented under cardano-config schema --help.

cardano-config/src/Cardano/Configuration/CliArgs.hs
jasagredo: parseHostPort now parses the URI authority via network-uri, so bracketed IPv6 ([2001:db8::1]:3001) is handled and a bare IPv6 no longer mis-splits.

cardano-config/src/Cardano/Configuration/File.hs
jasagredo: Done. ConfigurationParsingError now carries errFile, errSection, errPath :: JSONPath and errMessage, and the Show instance renders the file/section plus aeson's path-aware message

cardano-config/test/Main.hs
jasagredo: Added. The test-suite asserts every committed schema (config.schema.json and one per component) equals the schema derived from the codecs, so they can't drift; cabal test all runs in CI. scripts/gen-schemas.sh regenerates them, and they're now packaged as data-files.

cardano-config/src/Cardano/Configuration/File/Storage.hs
jasagredo: Switched to matchChoiceCodec, which dispatches on the JSON shape: an object goes straight to the SnapshotOptions codec, so a SnapshotInterval/delay-range failure is reported on its own without the irrelevant string-branch error. (NodeDatabasePaths got the same treatment.)

cardano-config/src/Cardano/Configuration/File.hs
jasagredo: Addressed by the structured error: decodeValueFile/loadSectionSource set errFile to the offending sub-file, so a malformed bad-storage.json is named in the ConfigurationParsingError.

cardano-config/src/Cardano/Configuration/File.hs
jasagredo: The parsing helpers (decodeValueFile, runCodec, loadSectionSource, loadBaseDefault, parseSection) now carry per-parameter haddocks. I can add them to the public parseConfigurationFiles/…With signatures too if you'd like.

cardano-config/src/Cardano/Configuration/CliArgs.hs
jasagredo: Done. The individual optparse-applicative parsers and argument readers (parseSocketPath, parsePort, parseHostPort, parseNodeAddress, …) are now exported alongside parseCliArgs.

cardano-config/src/Cardano/Configuration/File/Network.hs
jasagredo: It was a leftover from the node's PartialNodeConfiguration and carried no meaning here; the fields are now unprefixed.

cardano-config/src/Cardano/Configuration/File/Network.hs
jasagredo: Fixed. Every component (including LocalConnectionsConfig, Network, Mempool, Testing) is now f-parameterised and gains a finalize that turns the merged partial form into Identity (erroring via requireField when a base default is missing), so resolution is uniform.

cardano-config/defaults/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+Genesis files are network-specific and only ever appear in a `Protocol.<network>`
+variant, never in the base: defaulting them to mainnet would silently
+mis-configure other networks.
+
+## Ownership
+
+`cardano-config` is currently the *origin* of these files, but each component is
+ultimately **owned by the layer that implements it** (networking owns
+`Network*.json`, consensus owns `Consensus.json`, and so on). The intended flow:
```
carbolymer, line 87: I'm not sure README is the best place to put future development plans. I think REAMDE should describe current state of things, not what the code might do later.

cardano-config/defaults/README.md
```diff
@@ -0,0 +1,124 @@
...
+  - the `*GenesisHash` fields
+  - `PBftSignatureThreshold`
+  - `CheckpointsFile`/`Hash`
+  - `SocketPath`
+  - `RpcSocketPath`
+  - `MempoolCapacityBytesOverride` (`NoOverride`)
+  - the three mempool timeouts (the node's default is no timeout),
+  - the Testing `Test<Era>HardForkAt*` / `DijkstraGenesis*` knobs.
+
+## Provenance / TODO
```
carbolymer, line 113: this also feels like it does not belong to a README

cardano-config/cardano-config.cabal
```diff
@@ -0,0 +1,135 @@
...
+    Cardano.Configuration.File
+    Cardano.Configuration.File.Consensus
+    Cardano.Configuration.File.Mempool
+    Cardano.Configuration.File.Network
+    Cardano.Configuration.File.Protocol
+    Cardano.Configuration.File.Storage
+    Cardano.Configuration.File.Testing
+    Cardano.Configuration.File.Tracing
+    Cardano.Configuration.Render
+    Cardano.Configuration.Schema
```
lehins, line 46: Any chance we can hide module that do not need to be public? If there are in fact such modules, of course

cardano-config/src/Cardano/Configuration/File/Protocol.hs
```diff
@@ -0,0 +1,142 @@
...
+    <*> optionalField "LastKnownBlockVersion-Alt" "Last known block version, alt"
+      .= byronSupportedProtocolVersionAlt
+
+-- | The genesis file (and optional hash) for the checkpoints.
+checkpointsObjectCodec :: JSONObjectCodec (Maybe (Hashed FilePath))
+checkpointsObjectCodec = optionalHashedFileObjectCodec "CheckpointsFile" "CheckpointsFileHash"
+
+-- | Configuration for the protocol
+data ProtocolConfiguration f = ProtocolConfiguration
+  { byronGenesis :: ByronGenesisConfiguration
```
lehins, line 111: Is it lazy on purpose? If so, it would be nice to have a comment about why it is. Otherwise it needs a strictness annotation. Considering that `ByronGenesisConfiguration` has its all fields strict, I suspect it is the latter
```suggestion
  { byronGenesis :: !ByronGenesisConfiguration
```

cardano-config/src/Cardano/Configuration/File/Protocol.hs
```diff
@@ -0,0 +1,142 @@
...
+import Data.ByteString (ByteString)
+import Data.Functor.Identity (Identity)
+import Data.Text (Text)
+import Data.Word
+import GHC.Generics
+
+-- | A maybe hashed entity, possibly a file.
+data Hashed a = Hashed
+  { hashed :: a
+  , hash :: Maybe (Hash Blake2b_256 ByteString)
```
lehins, line 28: Why is the Hash optional?
Am I reading this correctly that the hash of genesis files is not required?

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,372 @@
...
+
+-- | Check the top-level configuration keys against the recognised ones, warning
+-- on (or, under 'RejectUnknownKeys', rejecting) any that are unrecognised.
+checkUnknownKeys :: UnknownKeyPolicy -> FilePath -> Value -> IO ()
+checkUnknownKeys policy cfgFile value =
+  case value of
+    Object o -> do
+      let unknown = [k | k <- map K.toText (KM.keys o), k `notElem` recognisedKeys]
+      unless (null unknown) $ do
+        let msg = "unrecognised configuration key(s): " <> intercalate ", " (map T.unpack unknown)
```
lehins, line 318: Slight simplification
```suggestion
      let unknown = [K.toString k | k <- KM.keys o, K.toText k `notElem` recognisedKeys]
      unless (null unknown) $ do
        let msg = "unrecognised configuration key(s): " <> intercalate ", " unknown
```

cardano-config/src/Cardano/Configuration/Render.hs
```diff
@@ -0,0 +1,188 @@
...
+        [ ("HostAddr" .=) . show <$> hostAddr nc
+        , ("HostIPv6Addr" .=) . show <$> hostIPv6Addr nc
+        , ("Port" .=) . portNumber <$> port nc
+        , ("TracerSocket" .=) . tracerConnectionValue <$> tracerSocket nc
+        , ("ShutdownIPC" .=) . fdNumber <$> shutdownIPC nc
+        , ("ShutdownOn" .=) . shutdownOnValue <$> shutdownOnTarget nc
+        ]
+  where
+    portNumber p = toJSON (fromIntegral p :: Integer)
+    fdNumber fd = toJSON (fromIntegral fd :: Integer)
```
lehins, line 156: `toInteger` is always safer, since it is not susceptible to overflows, as such it also does not require any type annotations 
```suggestion
    portNumber p = toJSON (toInteger p)
    fdNumber fd = toJSON (toInteger fd)
```

cardano-config/src/Cardano/Configuration/Render.hs
```diff
@@ -0,0 +1,188 @@
...
+
+tracerConnectionValue :: CLI.TracerConnection -> Value
+tracerConnectionValue (CLI.TracerConnection name method) =
+  object ["Name" .= name, "Method" .= methodValue method]
+  where
+    methodValue :: CLI.TracerConnectionMethod -> Value
+    methodValue = \case
+      CLI.TracerConnectViaPipe p -> object ["Pipe" .= p]
+      CLI.TracerConnectViaRemote host p ->
+        object ["Host" .= host, "Port" .= (fromIntegral p :: Integer)]
```
lehins, line 183:
```suggestion
        object ["Host" .= host, "Port" .= toInteger p]
```

cardano-config/src/Cardano/Configuration/Schema.hs
```diff
@@ -0,0 +1,458 @@
...
+      , Just t <- branchTitle b =
+          Object (KM.insert "title" (String t) b)
+    addTitle v = v
+    -- Name a branch by its const value or its type; leave structural branches
+    -- (e.g. a bare @{ required: [..] }@ constraint) untitled.
+    branchTitle b = case KM.lookup "const" b of
+      Just (String s) -> Just s
+      _ -> case KM.lookup "type" b of
+        Just (String t) -> Just (T.toTitle t)
+        _ -> Nothing
```
lehins, line 372: Cleaner way to write the same thing
```suggestion
    branchTitle b = do
      String s <- KM.lookup "const" b
      String t <- KM.lookup "type" b
      Just $ T.toTitle t
```

cardano-config/src/Cardano/Configuration/Schema.hs
```diff
@@ -0,0 +1,458 @@
...
+-- flat top-level key space, so their defaults are merged into one overlay.
+wholeConfigSchemaWithDefaults :: [(Text, Value)] -> Value
+wholeConfigSchemaWithDefaults defs =
+  withDefaults (foldr (deepMerge . snd) (Object KM.empty) defs) wholeConfigSchema
+
+-- | Each component schema with its @default@s filled in from its
+-- @defaults\/<Component>.json@ (when one is supplied).
+configurationSchemasWithDefaults :: [(Text, Value)] -> [(Text, Value)]
+configurationSchemasWithDefaults defs =
+  [ (name, maybe s (\d -> withDefaults d s) (lookup name defs))
```
lehins, line 434: This is pretty bad. `lookup` is O(n), which makes this whole function O(n^2), furthermore there is also a `lookup` call on the result.
At the very least converting `defs` to a `Map`:

```suggestion
  let defsMap = Map.fromList defs in
  [ (name, maybe s (`withDefaults` s) (Map.lookup name defsMap))
```
Ideally switching to an `OMap` in all such places where `lookup` is used whenver order matters, or regular `Map` if order is irrelevant

cardano-config/src/Cardano/Configuration/Render.hs
carbolymer, line 1: There is a lot of boilerplate with rewriting `a Identity -> a Maybe`. Part of it comes from the fact that not all fields are `f a` where `f` is a configuration type parameter - it then could be generalised with `barbies`.

Tbh I don't have any good arguments against wrapping every configuration field in `f`. It just adds more type-level boilerplate everywhere - but you have that in most places already.

see also https://chrispenner.ca/posts/hkd-options

cardano-config/src/Cardano/Configuration/File.hs
```diff
@@ -0,0 +1,372 @@
+{-# LANGUAGE GADTs #-}
+
+-- | The representation of the configuration file
+module Cardano.Configuration.File (
```
carbolymer, line 4: `File.hs` currently handles error types, JSON layering, key linting, and orchestration in a single module (~250 lines). It's manageable today but likely to grow as more validation or versioning logic
lands.

A natural split:

| Module | Responsibility |
|--------|---------------|
| `File.Error` | `ConfigurationParsingError` and its instances |
| `File.Merge` | JSON plumbing: `decodeValueFile`, `mergeValues`, `loadSectionSource`, `loadBaseDefault`, `sectionUserLayer`, `parseSection`, `splitEnvelope`, `runCodec` |
| `File.Lint` | `UnknownKeyPolicy`, `checkUnknownKeys`, `checkShadowedKeys` - currently the only part that depends on `Schema.recognisedKeys`/`componentPropertyNames` |
| `File` | Orchestration (`parseConfigurationFiles`, `parseConfigurationVersion1`, `NodeConfigurationFromFileF`) and re-exports |

The main benefit is that linting (which pulls in the schema) separates from the layering engine (which doesn't), making both independently testable.

Not blocking - the file is short enough today.

cardano-config/README.md
```diff
@@ -0,0 +1,362 @@
...
+{
+    "$id": "https://raw.githubusercontent.com/IntersectMBO/cardano-base/master/cardano-config/schemas/NetworkConfig.schema.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "NetworkConfiguration",
+    "properties": {
+        "AcceptedConnectionsLimit": {
+...
+```
+
+### ... see the schema for the whole config (deprecated, please use sub-files)
```
carbolymer, line 77: so if this was not released already, can this pargraph be removed althogether?

lehins (commented):
As I've mention in a private chat, I believe it will be much better to have this package be in a standalone repo and make it also depend on Ledger, so it is capable of parsing Ledger Genesis files and verify their hash and potentially provide schema for their content as well.
Some minor suggestions are also included int this review

cardano-config/README.md
```diff
@@ -0,0 +1,362 @@
...
+  `GenesisMode` (the default mode is `PraosMode`), so they cannot live in the
+  base files and keep their defaults in the codec.
+- **Where validation lives.** Structural validation of a single value lives in
+  its codec (and thus in the schema). Cross-field validation — constraints that
+  span CLI and file values or several components — lives in `resolveConfiguration`
+  as a list of `ConfigCheck`s; consumers can add their own with
+  `resolveConfigurationWith`.
+- **Errors.** File/JSON failures are reported as `ConfigurationParsingError`,
+  which records the offending file, section and JSON path. Resolution failures
+  are reported as `ConfigResolutionError`, listing the violated checks.
```
carbolymer, line 362: I don't see much value in describing the code in the README.

cardano-config/README.md
jasagredo: If we don't parse the whole schema, we would break the configuration of every user.

cardano-config/src/Cardano/Configuration/File/Protocol.hs
jasagredo: That is indeed the case today in the node

cardano-config/src/Cardano/Configuration/File/Protocol.hs
```diff
@@ -0,0 +1,121 @@
...
```
Jimbo4350, line 26: A comment about how the hash is used and why it's optional would be useful.

cardano-config/src/Cardano/Configuration/CliArgs.hs
```diff
@@ -0,0 +1,395 @@
...
+import Data.IP (IPv4, IPv6)
+import Data.Text (Text)
+import qualified Data.Text as Text
+import Data.Word (Word64)
+import Network.Socket (PortNumber)
+import Options.Applicative
+import System.Posix.Types (Fd (..))
+import Text.Read (readEither, readMaybe)
+
+data ShutdownOn
```
Jimbo4350, line 31: Haddocks for at least some of these types would be useful i.e explaining what they do.

cardano-config/src/Cardano/Configuration/File/Protocol.hs
jasagredo: I made the hashes now mandatory in cardano-config

cardano-config/src/Cardano/Configuration/Schema.hs
jasagredo: Thanks!

cardano-config/src/Cardano/Configuration/Schema.hs
jasagredo: Indeed, switched to using maps. Sorry for this!

cardano-config/src/Cardano/Configuration/Render.hs
jasagredo: Right, barbies doesn't really buy us much and we need to refactor all datatypes. Using `Last` as the node does results in a runtime check for undefined values instead of being enforced statically by the types, so I think we should stick with the current design

cardano-config/defaults/README.md
jasagredo: Removed

cardano-config/defaults/README.md
jasagredo: Removed

cardano-config/README.md
jasagredo: I added a warning and still parse it

cardano-config/cardano-config.cabal
jasagredo: In the new version I am using private sublibraries, which results in the API boiling down entirely to just `Cardano.Configuration`

cardano-config/src/Cardano/Configuration/File/Testing.hs
```diff
@@ -1,18 +1,22 @@
...
 import GHC.Generics (Generic)
 
 -- | The testing configuration: knobs for forcing era transitions at specific
--- epochs/versions and for enabling the experimental era.
-data TestingConfiguration = TestingConfiguration
-  { experimentalHardForksEnabled :: Bool
+-- epochs/versions and for enabling the experimental era. Only
+-- @ExperimentalHardForksEnabled@ has a default; the rest are optional by nature.
+data TestingConfiguration f = TestingConfiguration
+  { experimentalHardForksEnabled :: f Bool
```
Jimbo4350, line 19: You have a few `Bool`s littered around the codebase. I generally don't use `Bool`s (and define sum types) to avoid boolean blindness. Makes the code more readable IMO.

Jimbo4350 (commented):
Overall LGTM. @carbolymer  pointed out some consistency issues that I agree with.

My main concern would be the readability of errors when they occur particularly when it comes to JSON decoding (e.g https://github.com/IntersectMBO/cardano-base/pull/671#discussion_r3453416830). The configuration is a complicated object and debugging obscure JSON errors will hinder the user experience of getting a (potentially misconfigured) node up and running.

cardano-config/src/Cardano/Configuration/File/Protocol.hs
jasagredo: The hash is no longer optional.

jasagredo:
Closing in favor of https://github.com/IntersectMBO/cardano-config/pull/1. @Jimbo4350, I will apply your suggestions there

cardano-config/src/Cardano/Configuration/File/Protocol.hs
jasagredo: But the machinery is there for the Checkpoints file, whose hash is still optional. Is this correct? In some sense why put a hash if it is optional and the file changes "rapidly", but on the other hand a hash gives you stability on what you are expected to read... 

Perhaps @johnalotoski could give his opinion on whether the Checkpoints file should have a mandatory hash or just an optional one.

cardano-config/src/Cardano/Configuration/CliArgs.hs
jasagredo: Done, added haddocks to `ShutdownOn`, `TracerConnectionMethod`, `TracerConnection`, `KESSource` and Credentials explaining what each represents, since they're exported for other tools (e.g. `cardano-cli`) to reuse.

cardano-config/src/Cardano/Configuration/File/Testing.hs
jasagredo: In general I agree. We use sum types where things are not really obvious, but the remaining ones are more obvious I think and real switches (`ExperimentalHardForksEnabled`, `EnableCSJ`, `PeerSharing`, ...).

We have 8 of those. I'd lean towards keeping them, but if you want any specifically changed to using a sum type, I can do it.

cardano-config/src/Cardano/Configuration/File/Protocol.hs
johnalotoski: For both the [previously] optional genesis file hashes and checkpoint file hashes, SRE has always included the hashes in config distributions.  I've only seen examples of the hashes not being used in some local testnet configs.  I don't see a reason we couldn't make them mandatory.  Even if a file changes relatively frequently, the level of effort to update the hash is low and provides confidence whatever file is read has the expected content.

PR #687, opened 2026-07-13
Fix encryptedDerivePublic to return Either instead of throwing on har…

disassembler:
…dened indices

Replace both partial `error` branches in `encryptedDerivePublic` with structured failures: hardened indices now return `Left XPrvHardenedDerivationUnsupported` and FFI failures return `Left XPrvInternalError`. The return type changes from the pure `(PublicKey, ChainCode)` to `Either XPrvError (PublicKey, ChainCode)`, matching the style of the sibling `encryptedDerivePrivate` API and eliminating the crash vector when callers forward attacker-controlled derivation paths.

# Description

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
+++ b/cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
@@ -427,6 +427,7 @@ data XPrvError
   | XPrvInvalidChainCode
   | XPrvPublicKeyMismatch
   | XPrvInternalError
+  | XPrvHardenedDerivationUnsupported
   deriving (Eq, Show)
 
 newtype EncryptedKey = EncryptedKey ByteString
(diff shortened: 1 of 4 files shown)


cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -673,26 +674,26 @@ encryptedDerivePublic ::
   DerivationScheme ->
   (PublicKey, ChainCode) ->
   DerivationIndex ->
-  (PublicKey, ChainCode)
+  Either XPrvError (PublicKey, ChainCode)
```
lehins, line 677: This is a breaking change to the interface, as such it needs a major version bump. Also a changelog entry is needed for change to type signature and addition of `XPrvHardenedDerivationUnsupported` case

lehins (changes requested):
Aside from missing necessary version bump and changelog it looks good. Thank you!

lehins:
Oh yeah, and it needs some `fourmolu`

lehins (approved):
Perfect! Thank you!

PR #689, opened 2026-07-14
Rename and expose key material

lehins:
# Description

See #688 for original intent that is being solved by this PR. 

It is probably easier to review this PR one commit at a time

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
+++ b/cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
@@ -309,25 +309,26 @@ encodeChainCode = toCBOR . psbToByteArray . unChainCode
 
 type data Validity = Validated | Unchecked
 
--- | Key material with the secret key in @sodium_malloc@'d locked memory.
-data KeyMaterial (v :: Validity) = KeyMaterial
-  { kmSecretKey :: !SecretKey
-  , kmPublicKey :: !PublicKey
-  , kmChainCode :: !ChainCode
+-- | Extended key material with the secret key in @sodium_malloc@'d locked memory, `PublicKey` `and
+-- ChainCode`.
+data ExtKeyMaterial (v :: Validity) = ExtKeyMaterial
+  { ekmSecretKey :: !SecretKey
+  , ekmPublicKey :: !PublicKey
+  , ekmChainCode :: !ChainCode
   }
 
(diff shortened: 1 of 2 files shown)


cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
```diff
@@ -610,7 +611,7 @@ validateSerializedKey eKey =
 
 encryptedKeyFormat :: EncryptedKey -> XPrvFormat
 encryptedKeyFormat (EncryptedKey bs)
-  | BS.length bs == keyMaterialSize = LegacyV1
+  | BS.length bs == extKeyMaterialSize = LegacyV1
```
perturbing, line 630: Given that we compare here the size in the context of _encrypted_ data, I assume that 
```haskell
type KEY_MATERIAL_SIZE = SECRET_KEY_SIZE + PUBLIC_KEY_SIZE + CHAIN_CODE_SIZE
...
extKeyMaterialSize = fromInteger (natVal (Proxy @KEY_MATERIAL_SIZE))
```
is for encrypted material. The naming now suggest that it is for the key material (non encrypted), e.g
```haskell
-- | Extended key material with the secret key in @sodium_malloc@'d locked memory, `PublicKey` `and
-- ChainCode`.
data ExtKeyMaterial (v :: Validity) = ExtKeyMaterial
  { ekmSecretKey :: !SecretKey
  , ekmPublicKey :: !PublicKey
  , ekmChainCode :: !ChainCode
  }
```
Just want to check this, as encrypted data is generally not the same length (to hide the length of the content).

@disassembler

perturbing (approved):
LGTM with one small comment

cardano-crypto-wallet/src/Cardano/Crypto/WalletHD/Encrypted.hs
disassembler: for V2 you'd be correct, ciphertext is bigger. But LegacyV1 used a stream cipher (XOR with a PBKDF2-derived keystream), so the encrypted sk is the same size as the plaintext sk. That's actually why `ENC_SECRET_KEY_SIZE = SECRET_KEY_SIZE` at line 221 - it's documenting that invariant explicitly. The pk and cc aren't encrypted at all in v1, they're just stored plaintext alongside it. So 128 bytes covers both the plaintext layout and the v1 "encrypted" layout.

PR #690, opened 2026-07-16
Remove `LeiosCert` encoder/decoder, export `BitField` constructor

Soupstraw:
# Description

Currently the only way to construct `BitField` is through the decoder. However the decoder only decodes definite-length encoded bytestrings, but for CDDL conformance we'll also need to support indefinite-length encodings. To make it possible to define such a decoder using versioned decoders in `cardano-ledger`, I've exported the `BitField` constructor. 

To prevent users from accidentally using the invalid decoders, I've also removed encoding and decoding functions for `LeiosCert`, `LeiosVoterId` and `BitField`.

Related discussion: https://github.com/IntersectMBO/cardano-ledger/pull/5874#discussion_r3577048812

# Checklist

- [ ] Self-reviewed the diff


lehins (approved):

cardano-crypto-leios/test/Test/Cardano/Crypto/Leios.hs
```diff
@@ -65,11 +52,6 @@ import qualified Test.QuickCheck as QC
 spec :: Spec
 spec = do
   describe "LeiosCert" $ do
-    prop "round-trips through CBOR" prop_roundtrip_LeiosCert
-    prop "decodes indefinite-length encoding" prop_decode_indefinite_LeiosCert
-    it "matches golden encoding" $
-      goldenEncoding "test/golden/LeiosCert" encodeLeiosCert exampleCert
```
ch1bo, line 71: @Soupstraw As you moved the encoders/decoders out of this module, where did the golden tests for `LeiosCert` (and ideally also `LeiosVoterId`) go? I could not find it in [the PR motivating this one](https://github.com/IntersectMBO/cardano-ledger/pull/5874)

Soupstraw: I must have forgotten them, good catch. I'll open an issue on `cardano-ledger` to add them back in.

PR #691, opened 2026-07-21
Update dependencies

erikd:
# Description

Update dependencies. The version of `cardano-crypto-class` was bumped to `2.6.0.0` because this change switched from depending on the `memoery` package to `ram` (supposed to be a drop-in replacement).

When bumping the upper bound on `crypton` I checked that the code built without errors and that `crypton-1.1.4` was actually in the cabal plan.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/cardano-crypto-class/cardano-crypto-class.cabal
+++ b/cardano-crypto-class/cardano-crypto-class.cabal
@@ -1,6 +1,6 @@
 cabal-version: 3.0
 name: cardano-crypto-class
-version: 2.5.1.0
+version: 2.6.0.0
 synopsis:
   Type classes abstracting over cryptography primitives for Cardano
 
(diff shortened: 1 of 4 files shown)


cardano-crypto-class/cardano-crypto-class.cabal
```diff
@@ -128,16 +128,16 @@ library
     cardano-base >=0.1.6,
     cardano-binary >=1.9.1,
     cardano-strict-containers,
-    crypton ^>=1.0,
+    crypton >=1.0 && <1.2,
```
lehins: `crypton-1.0.x` uses `memory`, it only switched to `ram` starting with `1.1.0`
```suggestion
    crypton ^>=1.1,
```

lehins (approved):

cardano-crypto-class/cardano-crypto-class.cabal
erikd: Fixed and commit/changelog entries updated.

PR #693, opened 2026-08-07
Leios keyless seats: verify PoP and require present keys in certs

### LINKED ISSUE #1010: Register BLS keys to enable stake-based committee selection
## Why

Complete the rollout of using proper cryptography on the testnet / in the prototype. This will require all SPOs to register their keys (which they can already do right now, after #776). At the same time, this is the first must-have feature of a production-grade leios-enabled cardano node. It will enable the bootstrapping phase.

## What

- The `cardano-node` can take `--shelley-bls-key` and uses it when set
- No keys are derived from public cold keys anymore
- The Leios components use the registered BLS public keys to derive the committee
- Stake-based committee selection is used (hard-coded param for P99)

## How

- [x] Just do it on `leios-prototype`
- [ ] Submit first equivalent PRs to `cardano-{api,cli}` master -> get BLS key cli into 11.2
  - [x] https://github.com/IntersectMBO/cardano-ledger/issues/5974
  - [ ] plumbing in consensus
  - [ ] plumbing in API
  - [ ] plumbing in CLI
- [x] Write about it in the `leios-design` document and `CIP-164`
  - [x] https://github.com/input-output-hk/ouroboros-leios/pull/1044
  - [x] https://github.com/cardano-scaling/CIPs/pull/33

ch1bo:
A committee seat may now be keyless: `seatVKey` is a `StrictMaybe LeiosVerificationKey`. This lets a committee always mirror the full stake distribution, with a seat per pool whether or not it has registered a Leios key.

This also includes a shift to a smart contstructor that checks the proof of possession and marks the data constructor as unsafe. See the BLS crypto primitives for why this is necessary.

Also update haddocks to be more succinct.


cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -113,105 +116,89 @@ leiosSignatureSize = fixedSize (Proxy @(SigDSIGN LeiosDSIGN))
...
-  -- want to add an orphan. A WHNF-only check on the wrapper is sufficient here:
-  -- the strict 'Vector' forces every cell to WHNF, and a WHNF 'LeiosVoter'
-  -- forces both of its strict fields, so "LeiosCommittee in WHNF" structurally
-  -- implies no thunks anywhere inside.
+  -- WHNF check suffices: a strict 'Vector' of strict-field 'LeiosSeat's has no thunks in WHNF.
   deriving (NoThunks) via OnlyCheckWhnfNamed "LeiosCommittee" LeiosCommittee
 
+-- | Build a 'LeiosCommittee' from an ordered vector of seats. A seat with no
+-- key — or one whose proof of possession fails to verify — is admitted keyless,
+-- so one bad proof can't take down the committee. Seat order is the voter
```
nfrisby: > so one bad proof can't take down the committee

It took me a while to realize that "take down the committee" is the _rogue public key attack_ that the proof-of-possession prevents... right?

If so, it's confusing for the "take down the committee" to end the sentence despite the mention of Proof of Possession to be isolated within the emdash parenthetical

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -113,105 +116,89 @@ leiosSignatureSize = fixedSize (Proxy @(SigDSIGN LeiosDSIGN))
...
+-- indexing, so derive it deterministically.
+--
+-- NOTE: Seat weights are assumed to be in range [0,1] with sum ≤ 1.
```
nfrisby, line 151: More of a `PREREQ` or `ASSUMPTION` than a `NOTE`

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -113,105 +116,89 @@ leiosSignatureSize = fixedSize (Proxy @(SigDSIGN LeiosDSIGN))
...
+--
+-- XXX: the positional @(key, pop)@ input is awkward; a @Map k (Weight, StrictMaybe (key, pop))@
```
nfrisby, line 153: I don't know what this XXX is getting at. I'm guessing that's because I don't know what the `k` type variable is supposed to be

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -113,105 +116,89 @@ leiosSignatureSize = fixedSize (Proxy @(SigDSIGN LeiosDSIGN))
...
+  Vector (StrictMaybe (LeiosVerificationKey, LeiosProofOfPossession), Weight) ->
+  LeiosCommittee
+mkLeiosCommittee seats =
+  UnsafeLeiosCommittee $
+    seats <&> \(mKeyPoP, w) ->
+      LeiosSeat
+        { seatWeight = w
+        , seatVKey = do
+            (vk, pop) <- mKeyPoP
+            case verifyPossessionProofDSIGN leiosSignContext vk pop of
```
nfrisby, line 165: I'm surprised that the `mkleiosCommittee` function is checking the proof of possession. Shouldn't we be checking that proof when the key registration tx/header itself arrives, rather than much much later when we need to the key to be usable?

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -113,105 +116,89 @@ leiosSignatureSize = fixedSize (Proxy @(SigDSIGN LeiosDSIGN))
...
--- up front.
-getLeiosVoterId :: HasCallStack => LeiosVerificationKey -> LeiosCommittee -> Maybe LeiosVoterId
-getLeiosVoterId vk committee =
-  toVoterId <$> V.findIndex ((== vk) . voterVKey) committee.leiosCommitteeVoters
+    idx = fromIntegral @Word16 @Int voterId.leiosSeatIndex
+
+-- | The 'LeiosSeatId' for a verification key (smallest matching index if the
+-- committee has duplicates), or 'Nothing' if absent. Errors on a committee with
+-- more than @2^16@ seats — already malformed for the 16-bit bitfield.
+getLeiosSeatId :: HasCallStack => LeiosVerificationKey -> LeiosCommittee -> Maybe LeiosSeatId
```
nfrisby, line 186: Linear scan over 2^16 array vector is moderately expensive... should we maintain an index to avoid this? How often is it called? Is that call latency critical?

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
```diff
@@ -287,30 +260,17 @@ data VerificationError
...
---   2. Sums those voters' weights from the committee; short-circuits with
---      'InsufficientWeight' if the sum is below the threshold.
---
---   3. Aggregates the contributing verification keys and verifies the
---      certificate's 'leiosCertSignature' against the aggregate key over
---      @msg@.
+-- | Verify a 'LeiosCert' against a committee, a weight threshold, and the
+-- signed message, returning the signers' total weight. Rejects a malformed
+-- bitfield, any bit on a keyless seat, a summed weight below the threshold , or
+-- a bad aggregate signature with a 'VerificationError'. Keys are trusted as
```
nfrisby, line 271: > Keys are trusted as PoP-checked by 'mkLeiosCommittee'.

Repeating my concern that the PoP check should maybe happen earlier? What's the benefit of deferring it until the epoch boundary?

nfrisby: Maybe it's far fetched, but I can imagine a reason to do the check promptly instead of just-in-time:

Suppose I accidentally submit a bad PoP. It'd be nice if I could see that error immediately (tx/header invalidity message) and therefore have time to fix my mistake rather only find out later, after the committee's keys have been fixed

cardano-crypto-leios/test/Test/Cardano/Crypto/Leios.hs
```diff
@@ -270,3 +290,105 @@ prop_aggregateLeiosCert_rejects_empty = forAll genN $ \n ->
...
+  mkLeiosCommittee . V.fromList $
+    [(maybe SNothing (SJust . keyPop) mSk, weight) | mSk <- seats]
+  where
+    weight = 1 / fromIntegral @Int @Weight (length seats)
+    keyPop sk = (deriveVerKeyDSIGN sk, createPossessionProofDSIGN leiosSignContext sk)
+
+-- | Every pool in the stake distribution gets a seat with its stake weight,
+-- whether or not it registered a key. Seats for pools with a registered key
+-- carry it ('Just'); the rest are keyless ('Nothing'). Crucially the seat
+-- count and the weights are independent of which pools have keys, so voter
```
nfrisby: > so voter indices are stable as keys come and go.

Either this isn't as important as it sounds, or I'm confused about things.

Especially the phrase "as keys come and go" makes it sound like a committee member's key can change -during_ the epoch. That's false, right?

nfrisby: "voter indices are stable" ... "stable" over what? Aren't they stable by construction, since the committee is calculated only on epoch boundaries?

So, overall: reword this comment

cardano-crypto-leios/test/Test/Cardano/Crypto/Leios.hs
```diff
@@ -270,3 +290,105 @@ prop_aggregateLeiosCert_rejects_empty = forAll genN $ \n ->
...
+              [ counterexample ("weight " <> show i) $
+                  seatWeight v === 1 / fromIntegral @Int @Weight n
+              | (i :: Int, v) <- zip [0 ..] voters
+              ]
+
+-- | A seat whose registered key comes with a proof of possession that does
+-- not verify is admitted to the committee but stripped of its key, exactly as
+-- if the pool had never registered one. Committee construction never fails on
+-- a bad proof: one misbehaving pool must not take down the whole committee
+-- (and the ledger POOL rule rejects such registrations upstream anyway).
```
nfrisby: > (and the ledger POOL rule rejects such registrations upstream anyway)

Ah ha, so the check in `mkLeiosCommittee` is merely defensive.

I suggest that be called out as such in the function itself, rather than only in this test's comment

nfrisby: Moreover, we could have the upstream check be part of a smart constructor, such that only `ValidatedBlsKey`s are stored for the pool. In which case this downstream defensive check is obviously dead code.... unless you're worried about a ledger state reconstructed from a snapshot file on node init being maliciously modified?

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Yes, I would suggest we do both. The way the types wire through the ledger will result in the PoP being available to us anyways, but the ledger rule checking a `PoolReg` cert would be a prime candidate for checking the PoP. Opened this now https://github.com/IntersectMBO/cardano-ledger/issues/5993

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Once per epoch eventually https://github.com/input-output-hk/ouroboros-leios/issues/1020

I did not change this here and in the prototype this function is called for every vote validation :grimacing:

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: https://github.com/IntersectMBO/cardano-ledger/issues/5993

cardano-crypto-leios/src/Cardano/Crypto/Leios.hs
ch1bo: Uniqueness is not enforced right now. I could have been more clear here.

cardano-crypto-leios/test/Test/Cardano/Crypto/Leios.hs
ch1bo: So a more type safe API in the `cardano-base` to force the ledger to check it earlier (in https://github.com/IntersectMBO/cardano-ledger/issues/5993) and not need the defensiveness here. I like it

cardano-crypto-leios/test/Test/Cardano/Crypto/Leios.hs
ch1bo: I think you have been rewiewing intermediary comments. How can I improve my PRs for you to review, squash them before "ready to review"?

ch1bo:
Removed the dependency change commit from this branch and PR, to keep it instead on a `leios-prototype` branch (until the whole integration is resolved). So this PR stays focused on the functional change.

